PR #34, opened 2015-01-21
Adding 'pgbouncer' as fallback_application_name

ferfebles:
After reading libpq documentation, I first tried to add the startup
parameter called ‘fallback_application_name’. But it failed when
connecting to a postgresql server.

Then, looking at ‘libpq’ source I found that there is no real connection
parameter called ‘fallback_application_name’ when connecting to a
server. It’s just a fallback string for the ’application_name’
parameter.
—————————
file: src/interfaces/libpq/fe-protocol3.c-

/\* Use appname if present, otherwise use fallback */
val = conn->appname ? conn->appname : conn->fbappname;
if (val && val[0])
    ADD_STARTUP_OPTION("application_name", val);
———————————


eulerto:
I'm not sure if pgBouncer wants to be **automagically** announced as such. In some cases, you don't want to tell server that connections are coming from pgBouncer (current behaviour). This patch is not backward-compatible because it'll force adding application_name for those pools that I don't want to expose the application information.

markokr:
It needs to be configurable.  And also use by cf_application_name_add_host.

petere:
I'm closing this PR because as discussed this feature will not be acceptable as currently designed.

PR #44, opened 2015-04-14
Don't prematurely release the server in the middle of a parse/bind/execute on the unnamed statement.

adunstan:
We found this feature useful especially for drivers like Go. It has been tested in the field and found to work seamlessly on our client's site.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,8 @@ endif
 #
 
 pgbouncer_LDADD := $(CARES_LIBS) $(LIBS)
-LIBS :=
+LIBS := -lpthread
+#LIBS :=
 
 EXTRA_pgbouncer_SOURCES = win32/win32support.c win32/win32support.h
 EXTRA_PROGRAMS = pgbevent
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -329,6 +329,8 @@ struct PgSocket {
 	bool own_user:1;	/* console client: client with same uid on unix socket */
 	bool wait_for_response:1;/* console client: waits for completion of PAUSE/SUSPEND cmd */
 
+	bool wait_for_bind:1;  /* state of extended query mode on unnamed statement */
+
 	usec_t connect_time;	/* when connection was made */
 	usec_t request_time;	/* last activity time */
 	usec_t query_start;	/* query start moment */
--- a/src/client.c
+++ b/src/client.c
@@ -492,10 +492,22 @@ static bool handle_client_startup(PgSocket *client, PktHdr *pkt)
 	return true;
 }
 
+static bool is_unnamed_statement_parse(PktHdr *pkt)
+{
+
+	if (pkt->type != 'P')
+		return false; /* not a Parse packet */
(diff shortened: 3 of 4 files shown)


markokr:
Should be fixed in master, please test.

adunstan:
No, unfortunately this has not fixed the issue that my patch addressed.

Here is a test case written in Go that exercises the issue

Expected output:

[andrew@emma pgbouncer-dev]$ go run testme.go 
ok, connected
relname is pg_statistic
relname is pg_type
relname is pg_authid

Actual output:

[andrew@emma pgbouncer-dev]$ go run testme.go 
ok, connected
2015/08/14 12:29:46 pq: unnamed prepared statement does not exist
exit status 1

Program:

package main

import (
	"database/sql"
        "log"
	_ "github.com/lib/pq"
        "fmt"
)

func main() {
	db, err := sql.Open("postgres", "dbname=postgres port=6432 sslmode=disable")
	if err != nil {
		log.Fatal(err)
	}
	fmt.Println("ok, connected")
	rows, err := db.Query("SELECT relname FROM pg_class where relkind = $1 limit 3", "r")
    if err != nil {
            log.Fatal(err)
    }
    defer rows.Close()
    for rows.Next() {
            var name string
            if err := rows.Scan(&name); err != nil {
                    log.Fatal(err)
            }
            fmt.Printf("relname is %s\n", name)
    }
	if err := rows.Err(); err != nil {
            log.Fatal(err)
    }
}

markokr:
What does the driver actually do?  To me it seems like driver bug - it uses Sync while it should use Flush.

markokr:
Ok, I reread the commit: "parse/describe/sync followed by bind/execute ... sync".

If there is really good explanation why driver needs Sync there and not Flush I may reconsider, but until then I see this as example of buggy driver.

adunstan:
What's buggy about it? It doesn't seem to be counter to the protocol as far as I can see, and I know of at least 2 drivers (Go and Rust) that work this way, even though libpq does not. It works perfectly when there is no pgbouncer proxy - Postgres itself handles it just fine.

jberkus:
marko: Go is one of the more popular languages to use with PostgreSQL, so the non-functionality of pgbouncer with Go is not a trivial issue.  Do you have a specific reason why they shouldn't be using Sync which would be persuasive, instead of just "I don't like it"?  Given that it works fine with direct connections to PostgreSQL.

markokr:
Hint: the answer to Sync is ReadyForQuery which also contains transaction state.

When pgbouncer sees ReadyForQuery:no-tx-open, then why does it need to look at the plan name to know protocol state?  In fact this means there is an ad-hoc meta-protocol in top of regular V3 protocol.

> so the non-functionality of pgbouncer with Go is not a trivial issue.

It will work perfectly fine with pgbouncer in session-pooling mode. If you want transaction pooling, your app and driver need to cooperate with pgbouncer.

PgBouncer is not a place for hacks to work around problems with non-cooperating apps&drivers.

PR #48, opened 2015-04-23
It solves possible infinite login to dropped database via pgbouncer.

okbob:
We have to handle this issue, due limmited number of available clients.

Test case:

createdb iii; psql -p 6543 -U pavel -c "select 10" iii; \
echo "select pg_terminate_backend((select pid from pg_stat_activity where datname = 'iii'));drop database iii;" |  \
psql  postgres; \
psql -p 6543 -U pavel -c "select 10" iii;

Solution:

I implemented new config option server_max_connect_timeout. This parameter is checked
on every connect fail and if it is exceed, then client is disconnected forced. Default
value is two minutes.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -218,6 +218,8 @@ struct PgPool {
 
 	usec_t last_lifetime_disconnect;/* last time when server_lifetime was applied */
 
+	/* remember first attempt for connect */
+	usec_t first_connect_time;
 	/* if last connect failed, there should be delay before next */
 	usec_t last_connect_time;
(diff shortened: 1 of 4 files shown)


markokr:
Actually, should not (server_connect_timeout + query_wait_timeout) solve it?

Is there case you don't want query_wait_timeout but want server_max_connect_timeout?

Plus check_fast_fail() shoud kill new clients without waiting for query_wait_timeout.

markokr:
I think default query_wait_timeout should fix this.

PR #49, opened 2015-04-30
Coverity

eulerto:
While testing Coverity, I fix some outstanding defects. Due to excellent code quality, there are only a few more defects (but some of them are false-positives) that I will take a look in another opportunity.

If you want to be included in the Coverity project [1], please let me know.

[1] https://scan.coverity.com/projects/4831

### DIFF
--- a/src/dnslookup.c
+++ b/src/dnslookup.c
@@ -554,6 +554,7 @@ static bool impl_init(struct DNSContext *ctx)
 	int fd;
 	struct dns_ctx *dctx;
 	struct UdnsMeta *udns;
+	int err;
 
 	dns_init(NULL, 0);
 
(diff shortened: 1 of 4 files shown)


lij55:
"res == -1" means something is wrong and errno show the detail of the error.
It just ignore the EAGAIN error code here so I agree that fix should be 
-		if (res == EAGAIN) {
+		if (errno == EAGAIN) {

eulerto:
@hlinnaka, good catch! Patch adjusted.

PJMODOS:
About the 5ceace9 , why do you use the err variable in one place but place event_add inside ifs directly in two other places? IMHO the event_add can be inside if statement even at the first place.

PR #62, opened 2015-08-06
Adapt system tests to work with modern BSD and MacOS

eradman:
- Replace `seq a b` with {a..b}, and auto-upgrade to bash if shell does
  not support range expansion (DASH on Ubuntu)
- Replace ipfw with pf; (MacOS 10.7+, *BSD)
- Replace `echo -n` with `printf` (MacOS)
- Change to test directory before starting
- Provide aproximate compatibility between MacOS and Linux options for
  sed and nc
- Auto-populate userlist.txt

Tested on MacOS 10.10, Ubuntu 14, OpenBSD 5.7

### DIFF
--- a/test/test.sh
+++ b/test/test.sh
@@ -2,8 +2,9 @@
 
 # Notes:
 # - uses iptables and -F with some tests, probably not very friendly to your firewall
-# - uses nc (netcat) with some tests, skips if not in path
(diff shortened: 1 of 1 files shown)


markokr:
Looks good, thanks!

PR #73, opened 2015-09-27
Consistent search_path between client and server

rabbitt:
This enables consistent search_path caching on the client, which is applied to the server each time a client is assigned a server socket.

This is useful for transaction and statement modes as it will ensure that, in e.g., a multi-tenant situation using multiple schemas, the state of the search_path a client socket sets up and expects for a given set of requests, will be maintained across multiple backend server assignments.  In particular, every time a client sends a 'SET SEARCH_PATH...', the client->vars are updated with the current search_path value. This value is then set on a server each time that client is reassigned a server socket.

Note: search_path is not cached on the client if the pool_mode is set to POOL_SESSION as it would be superfluous to do so, given that there is a 1:1 mapping between client & server.

Additional Note: This also adds output to 'SHOW FDS' to allow viewing of currently cached search_path values on client fds.

P.s. This backports pretty cleanly to the 1.6 stable line - I can push a patch for that too if you accept this one.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -476,6 +476,8 @@ extern char *cf_server_tls_cert_file;
 extern char *cf_server_tls_key_file;
 extern char *cf_server_tls_ciphers;
 
+extern int cf_consistent_search_path;
+
 extern const struct CfLookup pool_mode_map[];
 
 extern usec_t g_suspend_start;
--- a/include/loader.h
+++ b/include/loader.h
@@ -1,12 +1,12 @@
 /*
  * PgBouncer - Lightweight connection pooler for PostgreSQL.
- * 
+ *
  * Copyright (c) 2007-2009  Marko Kreen, Skype Technologies OÜ
- * 
+ *
  * Permission to use, copy, modify, and/or distribute this software for any
  * purpose with or without fee is hereby granted, provided that the above
  * copyright notice and this permission notice appear in all copies.
- * 
+ *
  * THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
  * WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
  * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
@@ -18,7 +18,6 @@
 
 /* connstring parsing */
 bool parse_database(void *base, const char *name, const char *connstr);
-
 bool parse_user(void *base, const char *name, const char *params);
 
 /* user file parsing */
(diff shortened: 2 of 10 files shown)


markokr:
pgbouncer is a postgres protocol pooler, it should not parse SQL.  It cannot work reliably and it's a performance hit to all queries.

If you really want tx-mode and differnet search-path for connections in same pool, you should be using SET LOCAL in all transactions.

rabbitt:
I'm actually on the train with my kids now but I'll respond later today.

rabbitt:
I understand that you are aiming to be a pure data pipe between client and server, but I think we can have our cake and eat it too - if done correctly or course.

As for "SET LOCAL", unfortunately for us, and I'm sure others, we don't wrap everything in explicit transactions, so SET LOCAL would have a limited effect for us. Thus, unless I'm missing something (entirely possible!), we end with up the problem where we need a consistent search_path state but cannot have both that and reduced connections to the backend (i.e., we have to use pool_mode == SESSION). 

I think these changes can help in this particular case, and given that it comes with a configuration option defaulting to disabled, I think we can do it at minimal cost to other users. If you are open to giving me a little more time, I'm currently working on benchmarking and tweaking my changes to ensure that they are as efficient as possible. Additionally, I'll work on putting together some graphs to show how PgBouncer is affected under the all three pool modes when the feature is both disabled and enabled.

Lastly, you mentioned that you didn't feel it could work reliably - do you have particular cases in mind where this approach might fail?

PJMODOS:
It's not full SQL parser, you can easily inject stuff to the query that will break the special purpose query parser. And I agree with Marko that it's not worth the overhead.

rabbitt:
@PJMODOS I'm not injecting anything into the query - I'm capturing the search_path value and adding that value to client->vars. That value is then applied during the next server acquisition (as part of the already builtin varcache_apply functionality) to recreate the previous search_path state. I'm not touching incoming queries apart from simply looking at them as much as necessary to determine if they are a "set search_path" query, and if so, capture the value. 

So far, my basic benchmark tests show that non "set search_path" queries operate at roughly the same speed as before  (+/- 1000ns). search_path setting queries obviously take longer than previously, hitting somewhere between 2k-10k ns more. 

Granted, these tests aren't yet extensive enough to give a true measure of the affect of the patch but, it's enough for me to certainly feel like it's worthwhile to continue pursuing this avenue.

markokr:
The unreliability is caused by PgBouncer not seeing whole SQL. Also PgBouncer does not see what happens in db functions or anon code blocks. And even if PgBouncer sees SQL, emulating behavior of session features is usually non-trivial.  And even if it's doable, it slows things down by additional requests, additional context switches.

And search_path is especially nutty case - it's historical accident that is still without sane semantics.  It does not invalidate cached plans.  Ok, we could do it ourselves.  This would give significant performance hit, but who cares about that.  So it's only half-nutty.  Now the cached plans will be replanned with old search_path.  This gives us full-nutty.  The proper solution would be to use search_path as part of key in plan caching but looks like this is not coming.

Conclusion: if you want kittens to be happy, don't do mid-session search_path switch.

----

Unrelated to all that, there is need to keep complexity of PgBouncer down. Currently the red-line is that PgBouncer will not parse SQL. Because if we start to do it for one feature, why not do this other feature too? What is the next complexity barrier we should not cross?

World does not need second PgPool.

Closing.

rabbitt:
@markokr Thank you for taking the time to provide insight into why this might be an unreliable solution. You could have simply closed the PR without providing an insights, but you didn't and I appreciate that.

PR #74, opened 2015-09-27
adds rpm spec and systemd service file

rabbitt:


markokr:
I could add it but I'm not using redhat.  How much maintenance this will be?  Does it work across different RHEL/CentOS/Fedora versions?

rabbitt:
Should be pretty low maintenance, and considering my company will be using it, I am certainly more than happy to help.

PJMODOS:
Might be worth to coordinate this with Devrim as he builds the official rpms.

PR #78, opened 2015-10-09
Add PAM support

istaheev:
Synopsis
------------
We use PgBouncer in our organization since it provides all the capabilities we need. The only thing we need is to support PAM authorization to be able to use advanced authorization methods such as LDAP. Since PgBouncer doesn't support PAM right now we decided to implement it ourselves and then contribute it to the main project.

Implementation details
-------------------------------------
Initially it was an attempt to implement full PAM support, both on the global level and in HBA. Unfortunately it was discovered that specifying PAM in HBA requires much more complex work so we implemented it only on the global level. Full support can be done later.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ pgbouncer_SOURCES = \
 	src/loader.c \
 	src/main.c \
 	src/objects.c \
+	src/pam.c \
 	src/pktbuf.c \
 	src/pooler.c \
 	src/proto.c \
--- a/README.rst
+++ b/README.rst
@@ -65,6 +65,12 @@ and their probing order:
 `./configure` also has flags `--enable-evdns` and `--disable-evdns` which
 turn off automatic probing and force use of either `evdns` or `getaddrinfo_a()`.
 
+PAM authorization
+-----------------
+
+To enable PAM authorization `./configure` has a flag `--with-pam` (default value is no). When compiled with
+PAM support new global authorization type `pam` appears which can be used to validate users through PAM.
+
 Building from GIT
 -----------------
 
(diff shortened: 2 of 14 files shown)


markokr:
Please remove most of #ifdefs - they are needed only for code that actually do pam calls.

One thing to note is that pam API is "sync" api - everything will be stopped until your LDAP call finishes. Is this OK?  If it is then I guess it OK to keep it as is, as it does not affect anybody else who does not use pam.

To get rid of such "stop-the-world" behaviour" you would need to move pam code to separate thead or process.

istaheev:
Thank you for the response.
I will remove most of the #ifdefs, I agree it would be better.
PAM calls are "sync" and they will block all the other communications. It should not be a problem if "fast" PAM modules are used (e.g. modules capable of doing authentication quickly). If general throughput is not so important PAM can be used in its current form, but in the general case it is better to make it asynchronous as you proposed, I will look into this.

markokr:
Another style issue - please convert all comments to C-style /* .. */ , C++ comments (//) are not used in current code and are reserved for test/debug commenting.

istaheev:
Asynchronous PAM authorization is added, please take a look if this is accepted. I will also do some minor cleanups soon.

istaheev:
@markokr: do you have any comments for the change?

Tuxie:
I for one would love to see this in an official release soon!

markokr:
I am postponing dealing with it until 1.7 is out.

But otherwise I am positive to have it merged.

PJMODOS:
I pushed this as single squished commit.

Issue #83, opened 2015-10-26
Test documentation


doismellburning:
As someone looking to make changes to pgbouncer, I'd love some additional documentation on the tests, their dependencies, et cetera, so I can test my changes locally etc.

doismellburning:
So far I've established, for the `tests` dir:

## `test.sh`

* Needs `initdb` on the `PATH`
* Looks for `pgdata/postgresql.conf`

## `conntest.sh`

Used by `run-conntest.sh`

## `hba_test`

Controlled via the `Makefile`

## `run-conntest.sh`

* Needs `pgbouncer` in the CWD
* Needs `asynctest` in the CWD (presumably made from `asynctest.c`)

## `stress.py`

* Needs `psycopg2`

## `test.sh`

* Needs `pgdata/postgresql.conf`

Issue #82, opened 2015-10-26
No rule to make target `doc/pgbouncer.1'


doismellburning:
Trying to build from git (6259672072c658dafdb60917d5e2e238ee210d71) based on the instructions at https://github.com/pgbouncer/pgbouncer/tree/6259672072c658dafdb60917d5e2e238ee210d71#building-from-git I hit the following:

```
$ make
make: *** No rule to make target `doc/pgbouncer.1', needed by `all'.  Stop.
```

A cheeky tweak to `Makefile` provided a short term fix:

```
 $ git diff
diff --git a/Makefile b/Makefile
index b787343..3495d1e 100644
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ dist_doc_DATA = README.rst NEWS.rst etc/pgbouncer.ini etc/userlist.txt
 DISTCLEANFILES = config.mak config.status lib/usual/config.h config.log
 
 DIST_SUBDIRS = doc
-dist_man_MANS = doc/pgbouncer.1 doc/pgbouncer.5
+dist_man_MANS =
 
 # files in tgz
 EXTRA_DIST = AUTHORS COPYRIGHT Makefile \
```

but I'm not sure how you'd fix it "properly"

Paxa:
I think just need to update documentation, need to add in build steps:
```
cd doc
make
cd ..
```

markokr:
pgbouncer.1 is included in tarball.

when building from git you need rst2man tool from python-docutils

doismellburning:
@Paxa Thanks!

Presumably there's a `Makefile` target that doesn't try to build the docs (so `rst2man` isn't a required tool); alas I'm unable to work out what it might be!

doismellburning:
@markokr That file may be included in the tarball, but as per my link in the original description, I'm talking about building from git

doismellburning:
Neat, cheers!

PR #84, opened 2015-10-26
CircleCI config

doismellburning:
I couldn't find any CI setup for pgbouncer, and figured it might help to avoid issues like #82 (and also illustrate how to run testing, cf. #83)

Would happily help maintain + expand this as you find useful; I figure it could be quite useful for development etc.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,14 @@
+dependencies:
+    pre:
+        - sudo apt-get update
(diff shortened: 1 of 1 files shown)


doismellburning:
(Note that this fails because of #82)

petere:
I've added a Travis CI config now, because that's what I was familiar with. I think that addresses this issue.

PR #112, opened 2016-02-21
Make build reproducible by dropping DBGVER handling

ChristophBerg:
Debian wants debug symbols for all builds (the binaries are stripped by
debhelper's dh_strip), but the "(compiled by...)" information added for
debug builds defeats the goal to make builds reproducible on the binary
level. Fix by dropping the DBGVER information which doesn't serve any
purpose if the build output is byte-identical on rebuild.

https://wiki.debian.org/ReproducibleBuilds


markokr:
Indeed, it seems to be generally pointless info.  Merged.

Thanks for the patch.

PR #120, opened 2016-03-07
Generate auth_file with md5 passwords instead of cleartext

rlucia:
Message above is quite self explanatory :-)

### DIFF
--- a/etc/mkauth.py
+++ b/etc/mkauth.py
@@ -16,7 +16,7 @@ except IOError:
 # create new file data
 db = psycopg2.connect(sys.argv[2])
 curs = db.cursor()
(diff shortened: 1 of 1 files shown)


markokr:
Storing always md5 is good idea.  But the 'passwd' in pg_shadow may already be md5-hashed.  Could you make query to detect that and return passwd as-is?

rlucia:
Good point! I fixed the query to cope with that.

petere:
I think this request is obsolete, because plain text passwords are no longer possible as of PostgreSQL 10, and in the presence of SCRAM, just MD5-hashing everything not starting with `md5` wouldn't work either. I think it's best to take the passwords as they are stored.

PR #123, opened 2016-03-18
Split log disconnections for client and server

ephillipe:


markokr:
Could you create single commit of those for simpler overview?

petere:
Closing because no response.

A future implementation should probably address both `log_connections` and `log_disconnections` consistently, and perhaps maintain the old settings names for convenience and backward compatibility.

PR #124, opened 2016-03-21
Stable 1.7 cdb

azamorano:


PJMODOS:
I don't think we want to merge the changelog part given that it does not say anything about what actually changed just who did it. Can you make pull request that contains only the actually relevant changes?

azamorano:
Sorry pushed to the wrong repo :)

PR #146, opened 2016-09-13
Fix a typo in the max_client_conn docs

mchung:
Hi,

The config.rst includes an equation describing max_client_conns that I believe contains a typo. This PR replaces the '+' with '='.

Thanks for maintaining this project.

cc: @markokr


markokr:
That's not a typo, pgbouncer needs a fd for each client connection (max_client_conn) and for each server connection (max_pool_size * total_databases * total_users)

Issue #141, opened 2016-08-11
Client goes out of sync when pgbouncer returns ReadyForQuery for CancelRequest 


cpatru:
What I observed is that sometimes clients get out of sync and start reading the result of the previous query. This doesn't happen deterministically but it's roughly this sequence:

```
    1. A query will timeout.
    2. A cancel request is sent due to timeout.
    3. The query eventually completes.
    4. A new query is send and the response is a plain `ReadyForQuery` (result is not there).
    5. The next query receives the result of the previous query.
```

I wrote [this java console app](https://github.com/cpatru/pgbouncer-check) that can reproduce the issue. However, I couldn't get it to fail with local pgbouncer and postgres. It's only reproducing when pgbouncer and postgres are on separate boxes with arround 0.3ms latency. 

This is what the client, using pgjdbc, is reading off the buffer (left). You can see the `ReadyForQuery` being received in the middle of a query being executed, as a result of the `CancelRequest` (sent on a separate connection). On the right, is what pgbouncer is observing.

![pgbouncer_readyforquery](https://cloud.githubusercontent.com/assets/6660918/17589004/2b39feb2-5fca-11e6-8407-a3670700615e.jpg)

This only happening when hitting pgbouncer. It didn't repro when connecting to postgres directly.

`CancelRequests` are sent through a separate connection on the client. Pgbouncer handles them by trying to find the original connection that the `CancelRequest` is referring to and then it might decide to send a `ReadyForQuery` message on that connection. It only does this if it considers that client connection [not linked](https://github.com/pgbouncer/pgbouncer/blob/fdf5b8ea95fb0489ec6569df4b23c7c4b218ee79/src/objects.c#L1295).

I find the behaviour pgbouncer exhibits for `CancelRequest` not inline with the postgres protocol [documentation](https://www.postgresql.org/docs/9.1/static/protocol-flow.html) and may be breaking the correctness of the protocol.

I've commented [this line](https://github.com/pgbouncer/pgbouncer/blob/fdf5b8ea95fb0489ec6569df4b23c7c4b218ee79/src/objects.c#L1295) in pgbouncer and the problem went away. 

So it is a race condition that's making pgbouncer think that the client connection is [not linked](https://github.com/pgbouncer/pgbouncer/blob/fdf5b8ea95fb0489ec6569df4b23c7c4b218ee79/src/objects.c#L1282) as shown in this diagram.

[...]

PR #147, opened 2016-09-14
Don't send ReadyForQuery on main_connection for CancelQuery requests

cpatru:
Don't send ReadyForQuery on main_connection as it may cause clients to get out of sync. Drop the connection for idle clients.

https://github.com/pgbouncer/pgbouncer/issues/141

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -1529,18 +1529,11 @@ void admin_pause_done(void)
 /* admin on console has pressed ^C */
 void admin_handle_cancel(PgSocket *admin)
 {
-	bool res;
-
 	/* weird, but no reason to fail */
 	if (!admin->wait_for_response)
 		slog_warning(admin, "admin cancel request for non-waiting client?");
 
 	if (cf_pause_mode != P_NONE)
 		full_resume();
-
-	/* notify readiness */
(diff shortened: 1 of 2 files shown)


PJMODOS:
Hmm I wonder if disconnecting the client is the right thing to do here. AFAIK when this would happen to normal postgres backend which is idle, nothing would be done (no protocol message, no close, no nothing). So ISTM that the correct action is to do nothing here as well?

OTOH I am not quite sure what would be the side effects inside pgbouncer so maybe killing connection is the safe action here for now.

BTW the admin connection will do same ReadyForQuery still, so that ought to be handled as well probably.

cpatru:
Yes, not being sure what the side effect was my main concern. What I was basically doing is trying to get closer to the original behaviour, before the send ReadyForQuery was introduced in this commit: https://github.com/pgbouncer/pgbouncer/commit/8613ca9320d1deacf30ad5b24d93a712c2ef41ee

It looks like the client was being disconnected before this was introduced.

I was aware of the admin connection, but I thought that shouldn't be as critical, but you're right, it's best to stay consistent.

cpatru:
@PJMODOS I've fixed your comments.

PR #151, opened 2016-10-17
track maxwait_us to enable seeing sub 1s maxwait stats

tcrayford:
seeing as `maxwait` is measured in seconds, and cast from an integer, you cannot see any `maxwait` below 999ms - they all show up as `0`.

The existing `maxwait` stat is left around for backwards compatability, but in general is far less useful to the average user.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -792,7 +792,7 @@ static bool admin_show_pools(PgSocket *admin, const char *arg)
 				    "sv_active", "sv_idle",
 				    "sv_used", "sv_tested",
 				    "sv_login", "maxwait",
-				    "pool_mode");
+                    "maxwait_us", "pool_mode");
 	statlist_for_each(item, &pool_list) {
 		pool = container_of(item, PgPool, head);
 		waiter = first_socket(&pool->waiting_client_list);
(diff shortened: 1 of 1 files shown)


markokr:
Is such compat really necessary?   I would change it to float instead.

tcrayford:
@markokr it doesn't need to be! I couldn't find any guidelines on the stance of pgbouncer on backwards compatability on these metrics.

markokr:
I don't know any guidelines either...

I don't want unnecessarility break stuff, but I don't want to unnecessarily maintain compat goo either.

In this case it seems simpler to just change the type and let whatever breaks just sync it.  The breakage will be easily detectable.  I cannot imagine huge amounts of legacy code depending on it.

In this case the question is that how exact float it should show there.  Fully lossless may be too noisy, maybe with msec precision?

tcrayford:
@markokr another option is to not show a float, just up the precision to being microseconds, which is what the tracking is done in (I think)

markokr:
Yeah, but that is internal detail, I would prefer to avoid exposing it to user.  The console originally was mainly meant to be usable for manual inspection and I think we should keep it that way, and maxwait is visually very simple <1 good, >1 bad.   It seems wrong to change units.

tcrayford:
@markokr for my application, over 10ms of `max_wait` is bad. I wouldn't be surprised if others had similar kinds of needs out of pgbouncer

markokr:
merged #212 instead

PR #168, opened 2016-12-12
Add 'pool_waitings' counter to 'SHOW STATS'.

Nastradamus:
'cl_waiting' is main metric for monitoring pool satturation and postgres performance.
But it is not enough because it shows only current state and can't be aggregated.
It would be nice to have counter of waiting clients in pgbouncer pool.

I added 'pool_waitings' counter to 'SHOW STATS' which counts only logged in clients waiting for busy pool.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -191,6 +191,7 @@ struct PgStats {
 	uint64_t request_count;
 	uint64_t server_bytes;
 	uint64_t client_bytes;
+	uint64_t pool_waitings;
 	usec_t query_time;	/* total req time in us */
 };
 
--- a/src/objects.c
+++ b/src/objects.c
@@ -185,6 +185,9 @@ void change_client_state(PgSocket *client, SocketState newstate)
 		statlist_append(&login_client_list, &client->head);
 		break;
 	case CL_WAITING:
+                /* count only logged in clients waiting for busy pool */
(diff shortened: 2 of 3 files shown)


Nastradamus:
There is an error. It must count 'max_db_connections' too.

curlup:
@Nastradamus any plans on this? Does it still contain an error wrt `max_db_connections`?

Nastradamus:
@curlup unfortunately now I have no motivation and time to deeper research that problem.

curlup:
@petere wdyt? shouldn't all the waitings be counted, no matter the reason? 

I can rebase this, but my understanding would be that it should be counted in the same spot as `wait_time` is, i.e. here https://github.com/pgbouncer/pgbouncer/blob/a0e1f696b94f91d79a1a45c72633c2c4f842fc8d/src/objects.c#L569

Also, from my point of view count of waitings should include `CL_WAITING_LOGIN` because client is waiting for a free SV conn anyway. But this goes against current impl of `wait_time` accounting, which is, again IMO, is wrong in that sense.

JelteF:
Overall this seems like a useful statistic. I think `avg_wait_time` would cover most cases too nowadays though. Closing this as it is too outdated to be mergeable. Feel free to open a new PR with an updated version of this.

Issue #177, opened 2017-01-24
Cannot see if a database is paused or suspended with show database


marco44:
Hi,

I got bitten this morning by a database that was paused. "show databases" didn't display that the database was paused.

I have a patch adding paused and disabled columns to the show databases command. Are you interested in a pull request ?

Regards

PJMODOS:
I merged your patch, thanks.

PR #178, opened 2017-01-24
Add paused and disabled field on the show databases command

marco44:
For issue #177


PJMODOS:
Thanks, seems useful and otherwise harmless, merged.

PR #182, opened 2017-02-01
update auth_query to ignore expired roles

### LINKED ISSUE #181: auth_query does not check if role is expired
roles in postgresql does have expiration date/time, and it would be nice, if pgbouncer would follow that and auth_query in main.c file would be changed to `SELECT usename, passwd FROM pg_shadow WHERE usename=$1 AND (valuntil IS NULL OR valuntil > now())` so roles that are expired would not "ping" database anymore.

lonepsycho:
roles in postgersql does have expiration date (default null), but if they are expired they are not allowed to connect. without modifications, expired roles can "ping" database, i.e. try to connect, while with provided change, they will be denied to login at pgbouncer level.

closes #181


eulerto:
@lonepsycho auth_query is a configurable parameter to allow users to add whatever condition you want (I could block superusers or query my application user table). Also, it is just a good default example. I don't see this change as an improvement.

lonepsycho:
agreed. missed that point, that it is a configurable option. now I like pgbouncer even more.

PR #183, opened 2017-02-07
Increased maximum length for postgres password

jeduardo:
As the postgres password is hashed it can easily become bigger than 64 characters on some deployments. 

Some providers (for example Heroku) have begun to adopt improved security policies for the postgres credentials generated automatically by their systems. In the situation where the credential is bigger than 64 characters, pgbouncer will simply read an incomplete password from the configuration file and then try to use it for authentication, resulting on a log message stating that the password authentication has failed, as below:

`WARNING server login failed: FATAL password authentication failed for user "blablatest"`

There is no indication whatsoever that the problem is actually located on the pgbouncer side. This small change increases the size of the password buffer to cope with lengthier passwords.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -855,6 +855,8 @@ for this database.
 Otherwise PgBouncer tries to log into the destination database with client
 username, meaning that there will be one pool per user.
 
+The value for ``password``is limited to 128 characters maximum.
+
 auth_user
 ---------
 
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -855,7 +855,7 @@ for this database.
 Otherwise PgBouncer tries to log into the destination database with client
 username, meaning that there will be one pool per user.
 
-The value for ``password``is limited to 128 characters maximum.
+The length for ``password`` is limited to 128 characters maximum.
 
 auth_user
 ---------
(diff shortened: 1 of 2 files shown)


michrome:
👍 Heroku Postgres passwords can be 65 chars.

(You beat me to the pull request by 2 mins, @jeduardo !)

jeduardo:
@michrome hah! But thanks for adding the additional info straight from the source. :)

uhoh-itsmaciek:
Also perhaps a more graceful failure if this limit is exceeded? Though that can be a separate issue.

jeduardo:
@uhoh-itsmaciek I think that indeed would be a separated issue. 

For the db and user names I can see where the 64 chars size came from, as postgres identifiers are limited to 63 chars, but the passwords can get quite lengthy if someone wants them to be so. 

Also looking into how the MAX_PASSWORD constant is used in the loops inside the rest of the pgbouncer code it would be a bit tricky to say "the config you set is exceeding our limits" without defining and enforcing said limits. 

I think adding a warning in the documentation section (which I just did) mentioning the maximum value processed for the password field is a good compromise for this particular change, specially because since this was missing from the documentation I needed to do a non-trivial amount of testing in order to isolate this problem to the pgbouncer itself.

jeduardo:
Hey @markokr, thanks for merging this one!

PR #196, opened 2017-03-30
Fix incorrect references to auth_mode

mhagander:
The parameter is called auth_type, not auth_mode.


markokr:
Thanks for spotting.

PR #200, opened 2017-04-11
Update README.rst

joshgrima:
Added package requirement for PAM support


titanofold:
libpam-dev is a package specific to Debian and is not the actual library required. It would be better, I think, to say "(Requires PAM.)"

Although, that's implied.

markokr:
If its debian-specific, then not worth adding extra noise to readme.

PR #202, opened 2017-04-16
Global management of 'auth_user'

gbartolini:
Please review and test this patch. I was able to make it working on my laptop, but it requires more testing. Thanks.


markokr:
Looks reasonable.

Issue #166, opened 2016-12-07
Feature request: Per-user configuration of pool size


stevenwinfield:
Hi,

We use pgbouncer in our production systems and on the whole we've been extremely pleased with it, though one feature we feel is missing is the ability to configure pool_size and reserve_pool_size on a per-user (or per-database/per-user) basis.

As many others have commented in similar feature requests, we have a small number of users (some automated) who need to use a lot of connections, and that is normal behaviour, but at the same time we would like to limit the pool size for other users.

Currently our options are to set the maximum number of connections for each user database-side, but then we lose pgbouncer's reserve pool behaviour, or set up more pgbouncer instances pointing at the same databases but with different pool sizes, but then the proper allocation of resources between these instances and managing their co-operation becomes cumbersome.

I imagine being able to add something like the following to the config file:
``[users]``
``username = default_pool_size=N reserve_pool_size=M ...``
...which would override the per-database settings, or allow new sections to be added to the config file:
``[databasename/users]``
``username = pool_size=N reserve_pool_size=M ...``
...to specify the pool sizes for <username> in database <databasename>.

Is this possible?

Best regards,
Steven.

topharley:
You can use following syntax:

```
[databases]
yourdb = host=localhost port=5432 dbname=yourdb auth_user=postgres pool_size=200
yourdb_20_pool_size = host=localhost port=5432 dbname=yourdb auth_user=postgres pool_size=20 reserve_pool=20```

As you can see we add FAKE database "yourdb_20_pool_size" that is pointed to your real "yourdb" database using dbname=. And now we can override pool_size/reserve_pool. Of course we'll have to change our connection string for that user, change dbname "yourdb" to "yourdb_20_pool_size", but often int's not a big deal.

notEpsilon:
Any plans for this?

andruche:
Looks like this feature was added in issue #911

PR #204, opened 2017-04-28
Allow setting per-user connection limits to unlimited, improve logging

qris:
See https://github.com/pgbouncer/pgbouncer/issues/166 for rationale.

There is undocumented support for per-user connection limits already in pgBouncer. However it is not quite flexible enough for us. We wanted to be able to put a limit on all users, except for specific users (e.g. DBAs, system processes) which are unlimited. So we hoped to have a low setting for the
global `max_user_connections`, and override it for specific users to *unlimited* in `pgbouncer.ini:`

    max_user_connections=3

    [users]
    system = max_user_connections=0
    dba = max_user_connections=0

This patch makes it possible to set the per-user connection limit to either `0`
(unlimited) or `-1` (fallback to global default), or a specific limit as before.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -180,6 +180,9 @@ extern int cf_sbuf_len;
 #define POOL_STMT	2
 #define POOL_INHERIT	3
 
+#define MAXCONN_UNLIMITED 0
+#define MAXCONN_FALLBACK  -1
+
 #define BACKENDKEY_LEN	8
 
 /* buffer size for startup noise */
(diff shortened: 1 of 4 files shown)


qris:
Steve Winfield pointed out that since I've inserted CL_WAITING_SLOT into the SocketState enum, the values of the other existing entries below it will shift by one, which would probably make it unsafe to do an online upgrade across this change, as the client state passed to the new pgBouncer instance over the unix domain socket could have some SocketState values misinterpreted.

petere:
This should be split into two separate pull requests.

qris:
Split off #404 as requested. Cut down this patch to contain only the "allow setting per-user connection limits to unlimited" part, and rebased to master.

JelteF:
This same feature has now been merged in #1137

PR #205, opened 2017-05-11
Added doc for how to build on OSX

raarts:
Just sparing somebody the time - especially the ACLOCAL trick.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -130,3 +130,20 @@ To unregister it, do::
     
         $ regsvr32 /u pgbevent.dll
 
+Building for OS/X
+-----------------
+
+After cloning the git repository:
+
+        $ brew install autoconf
+        $ brew install automake
+        $ brew install libtool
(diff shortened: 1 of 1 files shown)


markokr:
This cannot be right, with all the version specific paths in commands.  Is the brew broken so it fails to configure its own packages properly.

In any case, brew is only one pkg managements system on mac, seems improper to hardcode for it.

raarts:
I guess you're not a Mac user..

Homebrew is definitely the leading package management system. By far. Calling it 'only one' isn't doing it justice. Also, there are reasons brew is handling it this way. For one, it tries not to conflict with standard OSX packages, but on the other hand these are quite old, and sometimes packages will not build, hence the specific paths. And that's exactly what the ./configure command line options are for.

markokr:
Yeah, but such low-level stuff belongs to brew receipe not in app readme.

PR #227, opened 2017-08-11
Add SASL and SCRAM support

### LINKED ISSUE #208: SCRAM support
Add SCRAM support for incoming and outgoing connections.

(If anyone is already working in this, please note it here.)

petere:
Add support for logging in to PostgreSQL 10+ servers requiring
SASL/SCRAM-SHA-256 authentication.

(Support for pgbouncer requiring SCRAM from its clients is not yet
implemented.)

Import a few files from PostgreSQL core (commit REL_10_BETA3) to support
SCRAM.

resolves #208 

---

Let me know what you think, especially how to manage the imported code. I'm thinking, `wchar.c` can perhaps be avoided by using the `utf8.c` module in libusual. Also, I'm thinking about splitting out some of the new code in `proto.c` into a new file `scram.c`.

I'll work on the client->pgbouncer support next.


michaelpq:
Did you try to link to the source code of Postgres instead of copying directly the files from upstream? It feels a bit a waste to have to copy things all over.. Any fixes happening in upstream need to be monitored and backported accordingly.

ruimarinho:
Also looking for this :)

MichaelDBA:
Is this already integrated into the master yet?  or am I dead in the water using PG scram authentication?

DylanYoung:
@MichaelDBA Dead in the Water.  Should be relatively straightforward to fix this PR up though.  The maintainer just wants to link to libpq instead of copying the source files (quite understandably).

Issue #228, opened 2017-08-23
pgboouncer 1.7.2 crash with simple hba file


MaximBoguk:
Hi,

I found that pgbouncer crash if try use hba file with following content (2 lines only):
=======================
host    all     all                             127.0.0.1/32    md5
host    all     xxxxxxxxxxxxxxxxxxxxxxxx        127.0.0.1/32    md5
=======================
Error depend on length of the 'xxx' username (23 bytes - no core, 25+ bytes - no core,  exactly 24 bytes - always core dumped on start) and doesn't depend on content of xxxx (any 24 latin symbols leads to crash).

```
gdb bt shows:
Core was generated by `/usr/sbin/pgbouncer -d /etc/pgbouncer/pgbouncer_tmp.ini'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f36ff35f428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007f36ff35f428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007f36ff36102a in __GI_abort () at abort.c:89
#2  0x00007f36ff3a17ea in __libc_message (do_abort=do_abort@entry=2, fmt=fmt@entry=0x7f36ff4bae98 "*** Error in `%s': %s: 0x%s ***\n") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007f36ff3ad4aa in malloc_printerr (ar_ptr=0x3, ptr=<optimized out>, str=<optimized out>, action=3) at malloc.c:5006
#4  _int_realloc (av=av@entry=0x7f36ff6eeb20 <main_arena>, oldp=oldp@entry=0x2625960, oldsize=oldsize@entry=32, nb=nb@entry=64) at malloc.c:4252
#5  0x00007f36ff3ae839 in __GI___libc_realloc (oldmem=0x2625970, bytes=52) at malloc.c:3045
#6  0x000000000040c33a in ?? ()
#7  0x000000000040c75b in ?? ()
#8  0x000000000040cc96 in ?? ()
#9  0x0000000000410082 in ?? ()
#10 0x00000000004061a2 in ?? ()
[...]

PR #229, opened 2017-08-24
Fix crash from memory overrun

petere:
Fixes: #228


petere:
We don't really have a general-purpose test suite that I'm aware of.

PR #231, opened 2017-08-28
Allow waiting clients to be prioritized based on their app name

terrence2:
When pgbouncer is running in an environment that is serving both realtime and non-realtime workloads (e.g. web servers and cron jobs), it is possible for a surge in background work to delay more urgent tasks unacceptably. This PR allows one to configure pgbouncer to cheat and send high-priority work to the front of the waiting_clients list so that it will be handled before lower priority tasks.

This new behavior must be configured explicitly in a new `[priorities]` section. Please see the updated documentation in `config.rst` for the full details.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -921,6 +921,25 @@ Set the pool mode to be used for all connections from this user. If not set, the
 database or default pool_mode is used.
 
 
+Section [priorities]
+====================
+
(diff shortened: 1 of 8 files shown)


terrence2:
This did not have as big an impact as we were hoping. Certainly not enough to justify the architectural carnage it inflicts. Closing for now. We can always re-open if we find a way to make it work better.

PR #232, opened 2017-08-29
add config to disable libevent2 randomize case

xuxiangyang:
add config to disable libevent2 randomize case，just like discuss in this issue https://github.com/libevent/libevent/pull/97 ，some nameservers do not follow the expected response behavior, they do not match the exact case of the name in the response.

### DIFF
--- a/etc/pgbouncer.ini
+++ b/etc/pgbouncer.ini
@@ -324,6 +324,9 @@ default_pool_size = 20
 ;; DNS negative result caching time
 ;dns_nxdomain_ttl = 15
 
+;; DNS randomize case
+;cf_dns_randomize_case = 1
+
 ;;;
 ;;; Random stuff
 ;;;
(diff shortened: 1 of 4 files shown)


cbandy:
> just like discuss in this issue libevent/libevent#97 ，some nameservers do not follow the expected response behavior, they do not match the exact case of the name in the response.

This was fixed in https://github.com/libevent/libevent/issues/288. The default of `"randomize-case", "1"` performs a `case-insensitive` comparison of the of the DNS answer:

https://github.com/libevent/libevent/blob/master/evdns.c#L1079-L1085

xuxiangyang:
some dns server just ignore this typical DNS security feature. people who use pgbouncer can not force dns server change to support this feature.... it is real world

petere:
I'm sympathetic to this problem.

I wonder whether we should add some kind of generic way to set libevent DNS options, so we don't have to expose each one of them via a separate pgbouncer option. But maybe that's overkill, I don't know.

petere:
This behavior in libevent can be disabled by putting

    options randomize-case:0

into `/etc/resolv.conf`.  Furthermore, #379 added the option of specifying a separate `resolv.conf` for PgBouncer, so this can now (when released) be set specifically for PgBouncer as well.

Therefore, I don't think we need a separate top-level configuration setting for this anymore (which would be specific to evdns, which would be potentially confusing).  So I'm closing this issue.

PR #243, opened 2017-10-23
add clues for sysadmin concerning HBA issues

gplv2:
Parsing the HBA config files is being done a bit too quiet.  It took me a lot of time sorting out an existing installation to figure out there was a failing line in the pgbouncer hba config file.  Logging a few warnings would help a lot in understanding where the problem is.   Hope this helps others.


PJMODOS:
Merged, thanks. I did changed the wording of for the hba.conf parse problem.

Issue #41, opened 2015-03-24
FATAL @src/objects.c:298 in function put_in_order(): put_in_order: found existing elem


petere:
Under certain circumstances, when making a configuration change and reloading, pgbouncer will crash on the next suitable connection attempt.  (This issue is similar to http://permalink.gmane.org/gmane.comp.db.postgresql.pgbouncer.general/898, but not the same.  Reproducible in 1.5.4 and 1.6rc1.)

Setup:
- `pgbouncer.ini`

``` ini
[databases]
* =
;test = user=postgres

[pgbouncer]
logfile = pgbouncer.log
pidfile = pgbouncer.pid

listen_addr = 127.0.0.1
listen_port = 6432

auth_type = trust
auth_file = userlist.txt
```
- `userlist.txt`

```
"postgres" "foo"
```
- You need a regular PostgreSQL server running on port 5432 with a database called `test`.
1. Start pgbouncer.
2. Connect: `psql -h localhost -p 6432 -U postgres test`
3. Disconnect
4. Uncomment the `test` database entry in `pgbouncer.ini`.
5. Reload: `kill -HUP $(cat pgbouncer.pid)`
6. Connect again.

Now pgbouncer will crash with this message:

```
FATAL @src/objects.c:298 in function put_in_order(): put_in_order: found existing elem
```

It is important that the newly added database entry contains a `user` parameter, otherwise it doesn't crash.

rsn86:
After changing pool_size and max_db_connections and reloading:

PgBouncer chrased with this message:

2015-11-17 15:38:10.038 15862 DEBUG C-0x24260d8: (nodb)/(nouser)@10.1.61.64:59991 got var: user=postgres
2015-11-17 15:38:10.038 15862 DEBUG C-0x24260d8: (nodb)/(nouser)@10.1.61.64:59991 got var: database=db_1g_ext_descanso
2015-11-17 15:38:10.038 15862 DEBUG C-0x24260d8: (nodb)/(nouser)@10.1.61.64:59991 got var: client_encoding=UTF8
2015-11-17 15:38:10.038 15862 DEBUG C-0x24260d8: (nodb)/(nouser)@10.1.61.64:59991 got var: DateStyle=ISO
2015-11-17 15:38:10.038 15862 DEBUG C-0x24260d8: (nodb)/(nouser)@10.1.61.64:59991 got var: TimeZone=Brazil/East
[...]

PR #250, opened 2017-11-29
Fatal error when reloaded w/ user=somebody

lawrencejones:
This change includes a script that can be used verify that PgBouncer 1.7.2 is definitely vulnerable to this bug (production issue... ouch) and the second commit modifies `force_user` to avoid this problem.

Given I'm unfamiliar with PgBouncer as a codebase, I'd love to get someone with more expertise to vet the change.

Thanks!

---

As described in issue #41, PgBouncer will fatally crash when config is
reloaded that specifies a user parameter for a database that already exists.

This is due to the config loader calling force_user for the newly
supplied user parameter, which creates a new user with an empty list of
user pools. When we later call find_pool(db, user) it is determined that
there is no associated pool, so we recreate a pool that already exists
and add it to the global pool list which then FATALs as a consequence of
duplicate elements.

We fix this bug by having force_user only create a new user if one for
this username does not already exist.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -445,17 +445,22 @@ PgUser *add_pam_user(const char *name, const char *passwd)
 /* create separate user object for storing server user info */
 PgUser *force_user(PgDatabase *db, const char *name, const char *passwd)
 {
-	PgUser *user = db->forced_user;
+	PgUser *user = find_user(name);
+
 	if (!user) {
 		user = slab_alloc(user_cache);
 		if (!user)
 			return NULL;
+
 		list_init(&user->head);
 		list_init(&user->pool_list);
+
 		user->pool_mode = POOL_INHERIT;
+
+		safe_strcpy(user->name, name, sizeof(user->name));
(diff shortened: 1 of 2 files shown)


lawrencejones:
@petere is there anything I can do to help this get merged?

lawrencejones:
Hey @petere, this feels like quite a critical bug fix, given it's being reported by several individuals and causes PgBouncer to die after performing an explicitly safe action (configuration reload).

We're really concerned that this bug is still lying around in our infra. Ideally we'd get this merged but I'm not sure how to best help people review this. Is there anything I can do to make this more reviewable?

Thanks.

lawrencejones:
Hey @petere! Do you know if you're ever going to be able to get this merged? It's quite impactful to people when this goes wrong and I'd imagine the effort involved reviewing this would be quite low.

If you don't want to merge this change then that's totally fine, but it would be good to close this PR before it hits it's second birthday!

lawrencejones:
Hey @petere, going to close this PR in a spring-clean. I think this is still a problem in the latest version though, so worth reviving this PR if it gets re-reported.

PR #254, opened 2017-12-07
SHOW VERSION as a row.

eulerto:
SHOW VERSION is historically printed as a NOTICE. The other SHOW commands are
printed as a row(set). Currently, it is not possible to expose version using a
view+dblink strategy (as it is done with the other SHOW commands) because of
this limitation.


davidfetter:
Can we get one of these pushed, please?

gplv2:
Bump

PR #255, opened 2017-12-08
Fix reserve_pool parameter

eulerto:
The correct per pool parameter is reserve_pool. The global parameter is named
reserve_pool_size.

While at it, document reserve_pool parameter.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -878,6 +878,11 @@ pool_size
 Set maximum size of pools for this database.  If not set,
 the default_pool_size is used.
 
+reserve_pool
+------------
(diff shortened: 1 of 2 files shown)


petere:
This is weird. Why isn't the per-pool parameter also named `reserve_pool_size`? The global and per-pool parameters should be named the same.

eulerto:
It is like that since day 1 (cf commit e94aeecaeebd277606f4e71ebfff55f68cdabca0). Maybe this wasn't intentional. Should we consider replacing `reserve_pool` by `reserve_pool_size` and silently accept `reserve_pool`?

PR #256, opened 2017-12-08
Document application_name parameter

eulerto:
It also fixes a typo.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -394,7 +394,7 @@ stats_users
 -----------
 
 Comma-separated list of database users that are allowed to connect and
-run read-only queries on console. Thats means all SHOW commands except
+run read-only queries on console. That means all SHOW commands except
 SHOW FDS.
 
 Default: empty.
@@ -896,6 +896,13 @@ max_db_connections
 
 Configure a database-wide maximum (i.e. all pools within the database will
 not have more than this many server connections).
+
+application_name
+----------------
+
+Set application name on connection start.  This logic applies only on start of
+connection, if application_name is later changed with SET, pgbouncer does not
+change it again.
 
 client_encoding
 ---------------


doc/config.rst
```diff
@@ -394,7 +394,7 @@ stats_users
 -----------
 
 Comma-separated list of database users that are allowed to connect and
-run read-only queries on console. Thats means all SHOW commands except
+run read-only queries on console. That means all SHOW commands except
```
petere: I've committed that separately.

doc/config.rst
```diff
@@ -897,6 +897,13 @@ max_db_connections
 Configure a database-wide maximum (i.e. all pools within the database will
 not have more than this many server connections).
 
+application_name
+----------------
+
+Set application name on connection start.  This logic applies only on start of
+connection, if application_name is later changed with SET, pgbouncer does not
```
petere: What do you mean by "changed with SET"? Do you mean if it's changed later by reloading the config file?

eulerto: Ooops. I mean "changed by SET".

petere: Still don't get it. `SET` is a PostgreSQL command, not pgbouncer. What is the scenario you are describing here?

eulerto: @petere I use the same sentence of `application_name_add_host` parameter in config.rst. If you don't get it, let's try to rephrase both such as:

> pgbouncer sets application_name when it opens a connection. If a client changes application_name via SET command then pgbouncer will not reset it.

curlup: @petere, I believe it's about PgBouncer's `SET command` 

from docs:
```
SET key = arg;
Changes a configuration setting (see also SHOW CONFIG). For example:

SET log_connections = 1;
```
here 
https://github.com/pgbouncer/pgbouncer/blob/master/doc/usage.rst#set-key--arg
or here
https://pgbouncer.github.io/usage.html#other-commands

PR #260, opened 2017-12-14
detect and fix errors in documentation sources

petere:


PJMODOS:
Why adding the AC check for rst2man, there already seems to be one present?
```
 +AC_CHECK_PROGS(RST2MAN, rst2man, rst2man) 
```

petere:
This changes the existing check to set `RST2MAN` to `rst2man` by default. This was the intention of the existing code, but the line `RST2MAN ?= rst2man` in `doc/Makefile` didn't actually work. Currently, if you don't have `rst2man` installed, you get junk man pages built.

PJMODOS:
Okay, makes sense

PR #279, opened 2018-02-09
Log as warning in case of fail due to full connection pool on launch_new_connection

weastur:
For example if we have pool_size greater than max_db_connection, than we will not be notified if launch_new_connection cannot create new one(of cause if we haven't verbosity level set to 2). So we need set log level in warning to see these important messages.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1129,7 +1129,7 @@ void launch_new_connection(PgPool *pool)
 				}
 			}
 		}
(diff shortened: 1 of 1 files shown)


amelius0712:
Any updates?

JelteF:
I'm closing this due to no responses from the author. But I agree with @petere that it's a useful change, but it indeed needs some more thought.

PR #286, opened 2018-04-06
Add server_max_requests configuration parameter

callerbear:
PostgreSQL worker processes cache information about every relationship (table) for each schema that a connection has referenced.  In a large multi-tenant database, perhaps with hundreds of schemas each of which contains hundreds of tables, this "relcache" can grow to unmanageable sizes and cause an unacceptable amount of virtual memory page swapping. The recommended solution is to periodically close the connections, thus freeing the memory allocated to the relcache.

The server lifetime configuration parameter isn't quite satisfactory for this purpose, since the paging problem is related to the number of different schemas that have been processed through each connection, not the length of time the connection has been open.

This pull request adds a new configuration parameter "server_max_requests".  After a connection has serviced the configured number of requests it will be closed, allowing a replacement connection to be opened and added to the pool.  The default value for this parameter is 0, which means no request limit will be enforced -- so the default behavior of the application is unchanged.

In our environment, we find that setting max_server_requests to values in the range of 25 to 100 avoids the problem of page swapping caused by huge worker relcaches without impacting performance in any significant way.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -468,6 +468,14 @@ then closed. [seconds]
 
 Default: 3600.0
 
+server_max_requests
+-------------------
+
+The pooler will try to close server connections after they have serviced this many requests.
+Setting it to 0 means there is no limit.
+
+Default: 0
+
 server_idle_timeout
 -------------------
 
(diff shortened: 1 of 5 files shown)


petere:
Closing because discussion had died down and the use case seems too narrow.

PR #287, opened 2018-04-18
Add option to disable printing stats to log

weastur:
Currently many of users of pgbouncer don't use logs to monitor it. For example we are using influxdb+grafana+custom-plugin to collect stats and display it to dashboard. So we absolutely don't need printing stats to log. So I added ability to disable this behavior(by default printing is on). Doc has been updated to. Additionally stats_period was moved from "Log settings" to "General setting" to reflect that this property means stats refreshing period - not only period of writing info to log.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -318,6 +318,13 @@ job_name
 
 Alias for `service_name`_.
 
+stats_period
+------------
+
+Period for writing aggregated stats into log.
+
+Default: 60
+
 
 Log settings
 ============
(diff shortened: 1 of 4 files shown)


petere:
In the way you have implemented it, if stats logging is disabled, we would still set the timer and compute all the stats but then just not print them. That seems wasteful. The effect of disabling this functionality should be that the timer is not set to begin with.

You could also just set `stats_period` to a very large value, so that the log entry rarely happens.

psapezhko:
Hello again. As I mentioned above, we are using stats in form of grafana dashboards. So stats are something that we need. But absolulty not in our logs.

curlup:
I see a reason in @psapezhko request - average stats are still needed to be counted but no need to log them, as external monitoring tool will collect them regularly.

PR #292, opened 2018-05-04
Make the client connection behave more like the Postgres server for SSL

adunstan:
Change the default cipher set to be the same as Postgres' default cipher
set, and have pgbouncer prefer the server's cipher set (i.e. its cipher
set) to the client's, like modern Postgres does by default.

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -606,7 +606,9 @@ Default: ``all``
 client_tls_ciphers
 ------------------
 
-Default: ``fast``
+Default is the same as the Postgres default
+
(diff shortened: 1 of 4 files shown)


petere:
The pgbouncer cipher sets are configuration in libusual, which has it's own predefined sets ("secure", "fast", "normal", etc.). Those sets already contain the corrects that the PostgreSQL default contains. So I don't think we need to change anything about that.

Preferring the server cipher list seems reasonable, but it seems most other products that use SSL have an option for this, so maybe we should too.

petere:
I think this is obsolete.

PR #298, opened 2018-05-29
Ignore cancel request during setting server variables

pf-qiu:
pgbouncer called 'sbuf_pause' to block the request from client during setting variables in 'find_server'. However, cancel request to Postgresql backend is sent through a new connection from psql client. In this case, the 'SET' SQL will be cancelled and an error message 'invalid server parameter' is returned to client, which is confusing from the psql client side.


lij55:
Hi,
Is there any progress about this PR? We encounter this issue while using connection pool of JDBC driver. When user cancel the query explicitly, it often give this 'invalid server parameter' error.
Thanks!

JelteF:
This seems like a real bug, and the fix was very small and looked sensible. So I resolved the merge conflicts.

JelteF (approved):

PR #301, opened 2018-06-07
Fix a double-free and a NULL pointer deref

danielgustafsson:
This addresses two bugs which are perhaps not terribly likely to happen:

* If the `fopen()` call fails `parse_namefile()` frees the filename buffer `fn`, and the caller does the same. Avoid freeing in `parse_namefile()` and leave the cleanup to the caller.
* If we fail to `calloc()` the `rule` buffer, going to `failed:` will dereference `rule` via the `rule_free()` call. Avoid by returning `false` directly and bypass the `failed:` block.

### DIFF
--- a/src/hba.c
+++ b/src/hba.c
@@ -359,7 +359,6 @@ static bool parse_namefile(struct HBAName *hname, const char *fn, bool is_db)
 
 	f = fopen(fn, "r");
 	if (!f) {
-		free(fn);
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


PJMODOS:
Looks good, merged

danielgustafsson:
> On 13 Jun 2018, at 23:16, Petr Jelinek <notifications@github.com> wrote:
> 
> Looks good, merged

Thanks!

PR #313, opened 2018-08-05
auth_dbname option for database definitions

dbonne:
Now you can specify an database's entry name for authentication.
If auth_dbname is set, auth_query is sended to that database.
This is util were we want use only one database for all users
authentications, i.e

   p0 = host=127.0.0.1 dbname=pgbench auth_dbname=authdb auth_user=pswcheck
   authdb = host=127.0.0.1 dbname=pgbouncer

If there isn't an entry named as the value specified in auth_dbname,
the database specified in dbname will be used instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,9 @@
 /test/test.log
 /test/test.pid
 
+# Visual Studio Code
+.vscode/
+
 *.html
 *.xml
 *.exe
--- a/test/test.sh
+++ b/test/test.sh
@@ -49,9 +49,9 @@ if test -n "$CAN_SUDO"; then
 		sudo pfctl -a pgbouncer -F all -q 2>&1 | grep -q "pfctl:" && {
 			cat <<-EOF
 			Please enable PF and add the following rule to /etc/pf.conf
-			
+
 			  anchor "pgbouncer/*"
-			
+
 			EOF
 			exit 1
 		}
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -16,6 +16,9 @@
 /test/test.log
 /test/test.pid
 
+# Visual Studio Code
```
petere: remove

test/test.sh
```diff
@@ -49,9 +49,9 @@ if test -n "$CAN_SUDO"; then
 		sudo pfctl -a pgbouncer -F all -q 2>&1 | grep -q "pfctl:" && {
 			cat <<-EOF
 			Please enable PF and add the following rule to /etc/pf.conf
-
```
petere: remove or make separate commit

test/test.sh
```diff
@@ -75,8 +75,8 @@ stopit test.pid
 stopit pgdata/postmaster.pid
 
 mkdir -p $LOGDIR
-rm -f $BOUNCER_LOG $PG_LOG
-rm -rf $PGDATA
+echo > $BOUNCER_LOG
```
petere: explain please

dbonne: My mistake including this in the commit. I will fix this.

test/test.sh
dbonne: I should remove my branch and recreate it for create a separate commit? I'm not a git expert.

.gitignore
dbonne: I will fix this.

PR #315, opened 2018-08-06
auth_dbname option for database definitions

dbonne:
Now you can specify an database's entry name for authentication. If `auth_dbname` is set, `auth_query` is sent to that database. This is useful where we want use only one database for all users authentications, e.g.,

```ini
p0 = host=127.0.0.1 dbname=pgbench auth_dbname=authdb auth_user=pswcheck
authdb = host=127.0.0.1 dbname=pgbouncer
```

If there isn't an entry named as the value specified in `auth_dbname`, the database specified in `dbname` will be used instead.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,74 @@
 PgBouncer changelog
 ===================
 
+PgBouncer 1.17.x
+----------------
+
+**2022-03-23  -  PgBouncer 1.17.0  -  "A line has been drawn"**
+
+- Features
+  * A database definition can specify a comma-separated host list.
+    The hosts will be connected to in a round-robin manner.
+  * When connecting to a non-existing database, the error ("no such
+    database") is now reported after authentication.  This prevents
+    unauthenticated clients from probing what databases exist.  (This
+    is similar to the change in version 1.15.0 to report missing users
+    after authentication.)
+  * Don't send server disconnect errors to the client before login.
+    This could reveal not-quite-public information, such as
+    configuration details, to a client that is not logged in yet.
+  * Increase maximum password length again.  Apparently, the last
+    increase wasn't enough for long enough.
+  * Remove automatic `auth_file` reload.  The `auth_file` is now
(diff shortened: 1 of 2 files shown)


dbonne:
Something weird is happening in CI. I don't know if there is an error in the code, or in the build process.I need some help here.

mpalmer:
I'd love to see this land.  I've just tested it against the current master, and it applies cleanly and works great.  *Much* better than having to define the auth lookup function against every database.

dbonne:
> I'd love to see this land. I've just tested it against the current master, and it applies cleanly and works great. _Much_ better than having to define the auth lookup function against every database.

Thanks for testing and using this patch. I hope this contribution can be integrated into the master branch. Thanks for the feedback.

Regards

martinmarques:
@petere anything we can do (run some tests over it) to get it merged in master?

mpalmer:
FYI, I've been running this patch on some 10+ pgbouncer multi-database instances in production since my last post, and it hasn't caused a single problem.

vjtm:
Hi folks,

This PR would be very wellcome. It is going to be merged soon?

Regards

protetore:
Same here, this would be very useful. This is what's holding us back in adopting pgbouncer in some scenarios.

grillorafael:
Same here @protetore

natethor:
This would be very useful for us in locking down the pgbouncer user to connect and validate authentication against only a specific database on our large multi-database servers.

SamSaffron:
@petere is there anything we can do to get traction on this PR, I am very happy to throw engineering hours at helping this meet your standards?

petere:
One question that comes to mind is whether the `auth_dbname` should be in terms of the PgBouncer database definitions or a database on the remote PostgreSQL server you're connecting to. I can see uses for both.

This patch is notably lacking documentation, which would surely be necessary to explain at least this particular aspect.

MichaelDBA:
Ahhh a function of course!  I must have missed that in the convo

PR #319, opened 2018-08-14
Slightly more secure function

davidfetter:
also, made it SQL, since it doesn't need PL/pgsql

### DIFF
--- a/doc/config.rst
+++ b/doc/config.rst
@@ -1053,16 +1053,17 @@ Database defaults::
   ; access to destination database will go with single user
   forcedb = host=127.0.0.1 port=300 user=baz password=foo client_encoding=UNICODE datestyle=ISO
 
-Example of secure function for auth_query::
+Example of a secure function for auth_query::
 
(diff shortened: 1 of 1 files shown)


petere:
How is this more secure?

davidfetter:
On Mon, Sep 10, 2018 at 01:35:43AM -0700, Peter Eisentraut wrote:
> How is this more secure?

It's guaranteed not to leak superuser information, at least to the
extent that the underlying system does.

Best,
David.
-- 
David Fetter <david(at)fetter(dot)org> http://fetter.org/
Phone: +1 415 235 3778

Remember to vote!
Consider donating to Postgres: http://www.postgresql.org/about/donate

petere:
I don't think this change is sensible, at least as a general example. If one thinks that giving out the password to a superuser is a problem, then one should also disallow superuser access via .pgpass, for example. Also, there may be other users that are not superusers but are also quite powerful. This is not the place to sort that out.

PR #321, opened 2018-08-18
Change `SHOW version` from a NOTICE to a query

davidfetter:
There's not really a good reason why this SHOW should give its output in
the form of a NOTICE where everything else returns a result set.


petere:
duplicate of #254

petere:
closing because duplicate

PR #324, opened 2018-08-31
Create Docker image

chriscowley:
Added a Dockerfile, entrypoint script and some basic docs to use with Docker

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM alpine:latest AS build_stage
+
+MAINTAINER chris@chriscowley.me.uk
+
+WORKDIR /src
+RUN apk --no-cache add git python py-pip build-base automake libtool m4 \
+                     autoconf libevent-dev openssl-dev c-ares-dev
+RUN pip install docutils
+COPY . /src/
+RUN ln -sv ../usr/bin/rst2man.py /bin/rst2man
+
+WORKDIR /src
+RUN mkdir /pgbouncer
+RUN git submodule init
+RUN git submodule update
+RUN ./autogen.sh && ./configure --prefix=/pgbouncer --with-libevent=/usr/lib
+RUN make
+RUN make install
+RUN ls -R /pgbouncer
+
+FROM alpine:latest
+RUN apk --no-cache add libevent openssl c-ares
+WORKDIR /
+COPY --from=build_stage /pgbouncer /pgbouncer
+ADD entrypoint.sh ./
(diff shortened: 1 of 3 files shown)


petere:
What is the purpose of this?

petere:
So why should this Docker image be based on Alpine Linux instead of something else? How can we settle that?

chriscowley:
@johanneswuerbach is exactly right. This would allow an official image on Dockerhub instead of the 1 I maintain for my own personal/professional needs. As I am not a pgbouncer expert, I am probably not the best person to be maintaining it long-term

@petere Alpine is a pretty lightweight and secure base image, so has become my go to. It also has `/bin/sh` and a package manager which is useful for troubleshooting. I regularly end up running tests by installing various clients and/or curl to sanity check things.

petere:
I don't know much about how this is supposed to work, but if the kidz these days want a Docker image, maybe we can consider it. ;-) But creating a copy of each configuration setting in your `entrypoint.sh` is clearly not maintainable, and ugly, and it's not clear what it's supposed to be for. That needs a better idea.

chriscowley:
@petere I will not try and say you are wrong because you're not. In an ideal world, pgbouncer would be able to take its configuration from environment variables - which is what the entrypoint.sh is trying to replicate. Really that is the best way to pass configuration into containers. The alternative is mounting a file in your container (do-able, but always feels a bit "dirty").

Laykou:
Would it be possible to define multiple databases (blocks) inside `[databases]`?

Maybe as multiple variables like `DB_HOST `, `DB_HOST_2`, .. `DB_HOST_N`? (+ similarly other variables). During initialization bash could iterate from 1 while the variable exists. Would that work, please?

My usecase: we have master and slave (via Patroni + HAProxy) and would like to have PGBouncer over both.

+ would you add also `CLIENT_ENCODING` variable? https://github.com/edoburu/docker-pgbouncer/pull/4/files#diff-b958f585a04af5ee2087610ea7180f34R59

petere:
This is not going to be acceptable in this form.

thenewguy:
@petere Any chance you can revisit this?  What would be acceptable?  An official image on docker hub would really be helpful even if only intended as reference for local projects to build upon for configuration needs.  I haven't looked at the code, but if you don't want environment variables, you can always bind mount a configuration file.  So you can provide a minimal default that ships with the package and have the assumption that projects will override the configuration file

PR #326, opened 2018-09-13
Enable "pam" authentication support via HBA

### LINKED ISSUE #501: Feature Request: multiple authentication methods
It would be really neat if pgbouncer supported multiple authentication methods through the databases section.

For instance I could have 2 db aliases where the absence of auth_user would imply using the default, auth_type=pam and the other implies validation using the auth_query in the database for the app users.
```
connect_apps = port=6432 dbname=connectdb auth_user=pgbouncer
connect_users = port=6432 dbname=connectdb 
auth_type = pam
```
This would accomplish a common scenario where we have individual users in an SDLC environment accessing the databases and app users where connection pooling applies the most.

Also, by distinguishing regular users from app users, we can now take advantage of sizing the pools more efficiently, which is already in pgbouncer.

If only pgadmin4 would work nicely with pgbouncer.  It hates db aliases! Requested a feature request there to support pgbouncer aliases (only a global change i think), but they would not budge.

moench-tegeder:
Cursory code review has shown no reason why "pam" should not be
supported in pg_hba.
Add "pam" to the list of supported authentication methods in
the HBA file - only if generally available, of course - and
edit docs to match the new reality.

Fixes #501

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -470,7 +470,7 @@ hba
 pam
 :   PAM is used to authenticate users, `auth_file` is ignored. This method is not
     compatible with databases using the `auth_user` option. The service name reported to
-    PAM is "pgbouncer". `pam` is not supported in the HBA configuration file.
+    PAM is "pgbouncer".
 
 ### auth_hba_file
 
@@ -1556,7 +1556,7 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
 * User name field: Supports `all`, `@file`, multiple names.  Not supported: `+groupname`.
 * Address field: Supports `all`, IPv4, IPv6.  Not supported: `samehost`, `samenet`, DNS names, domain prefixes.
 * Auth-method field: Only methods supported by PgBouncer's `auth_type`
-  are supported, plus `peer` and `reject`, but except `any` and `pam`, which only work globally.
+  are supported, plus `peer` and `reject`, but except `any`, which only works globally.
 * User name map (`map=`) parameter is supported when `auth_type` is `cert` or `peer`.
 
 ## Ident map file format
--- a/src/client.c
+++ b/src/client.c
@@ -535,7 +535,7 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 
 		if (!check_user_connection_count(client))
 			return false;
-	} else if (cf_auth_type == AUTH_TYPE_PAM) {
+	} else if (client->client_auth_type == AUTH_TYPE_PAM) {
 		if (client->db->auth_user_credentials) {
 			slog_error(client, "PAM can't be used together with database authentication");
 			disconnect_client(client, true, "bouncer config error");
--- a/src/hba.c
+++ b/src/hba.c
@@ -755,6 +755,10 @@ static bool parse_line(struct HBA *hba, struct Ident *ident, struct TokParser *t
 		rule->rule_method = AUTH_TYPE_PEER;
 	} else if (eat_kw(tp, "cert")) {
 		rule->rule_method = AUTH_TYPE_CERT;
+#ifdef HAVE_PAM
+	} else if (eat_kw(tp, "pam")) {
+		rule->rule_method = AUTH_TYPE_PAM;
+#endif
 	} else if (eat_kw(tp, "scram-sha-256")) {
 		rule->rule_method = AUTH_TYPE_SCRAM_SHA_256;
 	} else {


doc/config.md
```diff
@@ -89,9 +89,9 @@ Default: not set
 How to authenticate users.
 
 pam
-:   PAM is used to authenticate users, `auth_file` is ignored. This method is not
+    PAM is used to authenticate users, `auth_file` is ignored. This method is not
```
petere: This change is incorrect.

src/client.c
```diff
@@ -272,16 +274,22 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 		return false;
 	}
 
+	if (cf_auth_type == AUTH_HBA)
```
petere: This information is already available in `client->client_auth_type`. No need to look it up again.

petere (dismissed):
You need to update all the places where `cf_auth_type == AUTH_PAM` is checked. See `src/objects.c` and `src/admin.c` for missing places. The problem is that without these changes the takeover does not work correctly when you have a `pg_hba.conf` with mixed `pam` and on-`pam` entries. `use_server_socket()` in `src/objects.c` needs to decide when processing a server socket entry from `SHOW FDS` whether to look up the user in the normal user list or add it to the PAM user list. Right now it knows this from the global `auth_type` setting, because it's the same for all sockets. But if we allow a mix of PAM and non-PAM authentication, then we have to solve this in another way. I think a good way to do this would be to change `SHOW FDS` to ship a flag "this is a PAM user". (Updating the `SHOW FDS` output needs to be done with care to allow cross-version upgrades. I'm not sure about how to do that in detail.)

wibrt:
In a more general way: 
the way to authenticate directly to postgres does not fully match the way one can authenticate against pgbouncer

Just thinking out freely, this may be wrong, but..:
what if there was a pam module "pam-postgresql"
which would allow pam to manage it's authentication against a postgresql instance;
(thus having pam delegate its authencation to postgres)
having this single pam module would facilitate pgbouncer all the authentication methods of postgres, also future ones.

A possible limitation:
what if pgbouncer connects to multiple different postgres instances?

petere:
> Just thinking out freely, this may be wrong, but..:
> what if there was a pam module "pam-postgresql"
> which would allow pam to manage it's authentication against a postgresql instance;

This exists, it's called pam-pgsql.

But I don't see what this has to do with this thread. If you do PAM authentication and use pam-pgsql for the PAM  module, then you are mostly just replicating what `auth_query` does.

wibrt:
as i understand it: pam-psql in this form has little added value; just as you mention.

what i mean: a pam module to check whether a user can authenticate against a postgresql cluster X, thus effectively authenticated against the pg_hba.conf of this postgresql cluster X;

the problem at hand to be solved is that the authentication methods for pgbouncer does not match 100% the authentication methods of postgresql, or a combination of them.
Example (in short): lets say you have mixture of pam, scram/md5, ldap/ad user authentication in your pg_hba.conf of postgres, there no straightforward way of using pgbouncer in this case (unless limited to certain auth methods; or running multiple pgbouncers on different ports for different auth methods)

So instead of adding auth methods to pgbouncer, maybe just one is enough,
a pam-module that delegates it

petere:
I've studied this a bit today. One major issue that I have identified is how authenticating the same user sometimes via PAM and sometimes not works. Internally, PgBouncer tracks PAM and non-PAM users separately (see `add_pam_user()`). If we allow both PAM and non-PAM authentication at the same time, then we could have the same user name registered twice. This would first of all crash in some places, but when that's fixed, is that what we want? For example, you'd have potentially two pools with the same database and user name, except that one is for a PAM user and one is not. This would make a lot of things very confusing.

I think one way to move forward here would be to analyze whether we can get rid of `add_pam_user()` and keep the PAM users in the normal user tree, perhaps with a flag and a separate password field, or something like that.

JelteF:
Given that #1052 changed the user management it should now be fine to have PAM users and non-PAM users in one system. The `SHOW FDS` thing we decided we don't care anymore about breaking because now we have a different method for online restarts.

So I decided to update this patch and address the other feedback from @petere. I intend to merge this soon given its simplicity.

JelteF (approved):

clarkbreyman:
The change to client.c:538 seems to be causing actual PAM auth segfaults by skipping the path that allocates the pool. I'm looking to see if there is an omission of setting client->auth_type to PAM when cf_auth_type is PAM.

clarkbreyman:
client->client_auth_type is not set until `finish_set_pool()` which is after the test in `set_pool()`. At the time of the test cf_auth_type is PAM but the client is still ANY.

lalitvc:
I was testing user pam auth methods with pgbouncer **`auth_type = hba`**  not working.
I performed various tests for scram-sha-256 and pam auth methods with pgbouncer auth_type = hba

**Result:** Login for users with auth methods scram-sha-256,md5 from pgbouncer is working , but PAM auth is not working.

I don't see any working example of this **`auth_type = hba`**   pam auth support.  I would appreciate it if anyone could review my example and let me know what went wrong.  

```
[postgres@node1 ~]$ pgbouncer --version
PgBouncer 1.24.0
libevent 2.1.8-stable
adns: evdns2
tls: OpenSSL 1.1.1k  FIPS 25 Mar 2021
systemd: yes

[postgres@node1 ~]$ ldd /usr/bin/pgbouncer
    linux-vdso.so.1 (0x00007ffc95bf6000)
    libevent-2.1.so.6 => /lib64/libevent-2.1.so.6 (0x00007f877c6cb000)
    libssl.so.1.1 => /lib64/libssl.so.1.1 (0x00007f877c436000)
    libcrypto.so.1.1 => /lib64/libcrypto.so.1.1 (0x00007f877bf4b000)
    libsystemd.so.0 => /lib64/libsystemd.so.0 (0x00007f877bbfe000)
    libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f877b9de000)
    **libpam.so.0** => /lib64/libpam.so.0 (0x00007f877b7ce000)
```

## Example:
#All users, including PAM users, are able to login to postgresql directly.
```
[postgres@node1 ~]$ psql -h 127.0.0.1 -U testusr1 -d db1 -p 5432
Password for user testusr1: 
psql (16.6 - Percona Distribution)
Type "help" for help.

db1=> 


[postgres@node1 ~]$ psql -h 127.0.0.1 -U pgbounceauth -d db1 -p 5432
Password for user pgbounceauth: 
psql (16.6 - Percona Distribution)
Type "help" for help.

db1=> 

[postgres@node1 ~]$ psql -h 127.0.0.1 -U devuser1 -d db1 -p 5432
Password for user devuser1: 
psql (16.6 - Percona Distribution)
Type "help" for help.

db1=> \du
                               List of roles
  Role name   |                         Attributes                         
--------------+------------------------------------------------------------
 dbauser1     | Superuser
 dbauser2     | Superuser
 devuser1     | 
 devuser2     | 
 pgbounceauth | 
 postgres     | Superuser, Create role, Create DB, Replication, Bypass RLS
 testusr      | 
 testusr1     | 
 testusrm
```


# pgbouncer.ini

```
[databases]
* = host=127.0.0.1 port=5432 auth_user=pgbounceauth

[pgbouncer]


auth_type = hba

auth_user = pgbounceauth
auth_query = SELECT pgbounce_user, pgbounce_password FROM bounce.lookup($1)
```
 

#userlist.txt

```
$ cat /etc/pgbouncer/userlist.txt 
"pgbounceauth" "md5f9fc3711c004565c6f120ad0ef59738d"
```
 

#/etc/pgbouncer/pg_hba.conf

```
host    all             testusr1        127.0.0.1/32            scram-sha-256
host    all             pgbounceauth    127.0.0.1/32            md5
host    all             devuser1,devuser2             127.0.0.1/32            pam
```


#DB pg_hba.conf

```
host    all             testusr1        127.0.0.1/32            scram-sha-256
host    all             pgbounceauth    127.0.0.1/32            md5
host    all             devuser1,devuser2             127.0.0.1/32            pam pamservice=postgresql
```

### pgbouncer PAM user login failing with below error,

```
[postgres@node1 ~]$ psql -h 127.0.0.1 -U devuser1 -d db1 -p 6432
Password for user devuser1: 
psql: error: connection to server at "127.0.0.1", port 6432 failed: FATAL:  PAM authentication failed
[postgres@node1 ~]$ 


[postgres@node1 ~]$ pgbouncer /etc/pgbouncer/pgbouncer.ini 
2025-03-17 09:52:10.530 UTC [2817] LOG kernel file descriptor limit: 1048576 (hard: 1048576); max_client_conn: 100, max expected fd use: 112
[warn] Unable to parse nameserver address fe80::216:3eff:fef0:8e48%eth0
2025-03-17 09:52:10.531 UTC [2817] LOG listening on 0.0.0.0:6432
2025-03-17 09:52:10.531 UTC [2817] LOG listening on [::]:6432
2025-03-17 09:52:10.531 UTC [2817] LOG listening on unix:/tmp/.s.PGSQL.6432
2025-03-17 09:52:10.531 UTC [2817] LOG process up: PgBouncer 1.24.0, libevent 2.1.8-stable (epoll), adns: evdns2, tls: OpenSSL 1.1.1k  FIPS 25 Mar 2021
2025-03-17 09:52:42.793 UTC [2817] LOG C-0x557ad46fdcb0: (nodb)/(nouser)@127.0.0.1:57786 registered new auto-database: db1
2025-03-17 09:52:42.793 UTC [2817] LOG S-0x557ad4706800: db1/pgbounceauth@127.0.0.1:5432 new connection to server (from 127.0.0.1:47046)
2025-03-17 09:52:42.798 UTC [2817] LOG C-0x557ad46fdcb0: db1/devuser1@127.0.0.1:57786 login attempt: db=db1 user=devuser1 tls=no replication=no
2025-03-17 09:52:50.449 UTC [2817] LOG C-0x557ad46fdcb0: db1/devuser1@127.0.0.1:38546 login attempt: db=db1 user=devuser1 tls=no replication=no
2025-03-17 09:52:50.450 UTC [2817] WARNING pam_authenticate() failed: Authentication failure
2025-03-17 09:52:50.535 UTC [2817] LOG C-0x557ad46fdcb0: db1/devuser1@127.0.0.1:38546 closing because: PAM authentication failed (age=0s)
2025-03-17 09:52:50.535 UTC [2817] WARNING C-0x557ad46fdcb0: db1/devuser1@127.0.0.1:38546 pooler error: PAM authentication failed
```


 

### For user with auth methods scram-sha-256 login from pgbouncer working.

```
[postgres@node1 ~]$ psql -h 127.0.0.1 -U testusr1 -d db1 -p 6432
Password for user testusr1: 
psql (16.6 - Percona Distribution)
Type "help" for help.

db1=> 

2025-03-17 09:55:36.523 UTC [2817] LOG C-0x557ad46fdcb0: db1/testusr1@127.0.0.1:39278 login attempt: db=db1 user=testusr1 tls=no replication=no
2025-03-17 09:55:39.517 UTC [2817] LOG C-0x557ad46fdcb0: db1/testusr1@127.0.0.1:39288 login attempt: db=db1 user=testusr1 tls=no replication=no
2025-03-17 09:55:39.519 UTC [2817] LOG S-0x557ad4706ac0: db1/testusr1@127.0.0.1:5432 new connection to server (from 127.0.0.1:49708)
```

 
Tested with below entry as well PAM auth now working, 

`host    all             devuser1,devuser2             127.0.0.1/32            pam pamservice=pgbouncer`

[postgres@node1 ~]$ pgbouncer /etc/pgbouncer/pgbouncer.ini 
2025-03-17 12:54:17.069 UTC [32457] **WARNING hba line 124: Ident map pamservice=pgbouncer is malformed. It is not in map=value format.**

Issue #59, opened 2015-07-10
auth_user seems to be ignored if the user is defined in pgbouncer.ini


benchub:
If we have configured a user in the Users section of pgbouncer.ini, but left their password out of the auth_file, I would expect that pgbouncer would try to use the auth_query command to lookup the password when they try to connect. Instead, pgbouncer seems to give up and just give an auth error when it can't find the user in auth_file.

Take that same user out of the Users section and it all works fine... except that I can't specify the pooling mode for that user anymore.

keithf4:
Been running into the same issue as you describe in the first paragraph since I assumed these new settings would mean it would just look up the username/password if not defined in the auth_file.

Honestly, i'm just thoroughly confused as to how it's even supposed to work in the first place. These new config options were just added with little explanation of how they're supposed work. Even just an example would be great.

stevenwinfield:
This is still present in 1.9.0

PierreDucroquet:
I think this bug is related to a configuration file parsing bug I found today.
The auth_user option from the [pgbouncer] section in pgbouncer.ini is ignored if it is defined *after* the [databases] section.
If you look at the parse_database function in `src/loader.c`, it uses `cf_auth_user ` as a default auth_user, but this function seems to be called as soon as the [databases] section is found in the configuration file.

pinaraf:
This may be related to #391.

qris:
The subject should be `auth_query` is ignored, not `auth_user`. Because of this bug, it's required to add each such user to the `auth_user` file, so it's definitely not ignored.

greigwise:
I'm running into this in pgbouncer 1.15.0.  Is there a fix anywhere in the near future for this?

cdstg:
Same on 1.16.1   auth_query runs if  add t auth_user in the  [pgbouncer]  section or in the databses section
But if i add user in databases or in the users section it then expects a password in the userlist

What I m trying to do is limit the pool  for a user but use the auth_query for the password lookup
It works if i remove user but then it only allows me to set the limits on the pool not the user

jtorral:
Ran into this issue today. The simple fix which took some trial and error may simply be an undocumented issue the what I encountered was that the username under the [users] section must be listed in the auth_file. Doing so, fixed the problem for me.

JelteF:
This should now be fixed by #1052

PR #333, opened 2018-10-05
Issue #59

stevenwinfield:
Potential fix for #59 
Tested and working as I expect it to, but I'm not sure how it will interact with the "fake" pgbouncer user.
Perhaps some password-not-set sentinel is needed, rather than the empty string.


JelteF:
The original issue has been fixed by #1052

PR #339, opened 2018-11-01
Set client->wait_start for CL_WAITING_LOGIN

mmickan:
Without this change, wait_start is not set when entering the
CL_WAITING_LOGIN state.  This results in a failed assertion in
activate_client() or, when assertions are not enabled, massively
inflated total wait times.


petere:
This has been fixed by c53678edc3255e5c22654a5fed69a8fb2945d496 as part of another issue. Didn't see that one at the time.

PR #342, opened 2018-11-12
Raise errors from log_debug to either warning or error.

dimitri:
Some errors would be logged only in the debug log level, which means
production setups won't see them. Those error messages are very useful when
debugging corner cases, some of them might look like pgbouncer outages
without the full story.

This patch doesn't add new logs, only raises the chatter up to a level where
it's possible to relate strange behaviors with pgbouncer internal activity.

### DIFF
--- a/src/dnslookup.c
+++ b/src/dnslookup.c
@@ -891,7 +891,7 @@ static void xares_host_cb(void *arg, int status, int timeouts, struct hostent *h
 		res = convert_hostent(h);
 		got_result_gai(0, res, req);
 	} else {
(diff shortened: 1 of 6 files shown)


petere:
There is probably merit to this, but I'd need to do more analysis about when each message is expected to happen. Also, when promoting debug messages to user-facing, they should be rephrased to a more user-facing style.

JelteF:
hi @dimitri. Doing some "spring" cleaning. Closing this as this has bitrotted too much.

PR #345, opened 2018-12-05
Add support for SO_REUSEPORT

leev:
SO_REUSEPORT allows multiple applications to listen on the same
address/port.  This allows multiple pgbouncer instances to be started
on the one server and have the operating system load balance requests
between the different instances. pgbouncer is then able to occupy more
than one CPU in a multi CPU environment.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -184,6 +184,19 @@ fi # !cares
 
 ## end of DNS
 
+AC_MSG_CHECKING([whether so_reuseport is available])
+AC_LINK_IFELSE([AC_LANG_SOURCE([
+  #include <sys/socket.h>
+  int main(void) {
+    int val = 1;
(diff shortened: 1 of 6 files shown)


petere:
If you do this and start multiple instances, how would you be able to address individual instances for maintenance and monitoring? Any connection attempt would just go to a random instance.

petere:
Who is to say that I don't want to also do load-balancing on Unix-domain sockets? And who is to say that I want to use administration only over Unix-domain sockets?

Creating an implicit split that Unix-domain sockets are for admin and TCP/IP sockets are for production traffic seems wrong.

leev:
Your merged version seems like a pretty minor change to what I provided here.  Why re-implement someone else's work who put the time and effort in to implement and test it and then take it with no recognition?

petere:
Sorry I forgot to credit you in the commit message.

Issue #139, opened 2016-07-11
 pgbouncer cannot connect to server


aek:
Hi @all
I have an issue with pgbouncer and I think that maybe someone else have the knowledge of how to solve it.
I'm creating databases using pgbouncer by connecting to the postgres database and running create database sql queries automatically of course, not manual, but when I try to connect to the new created database pgbouncer says that cannot connect to the server
I'm closing cursors and everything, the database is there but pgbouncer seems to not know about that. How could I solve it without directly connecting to postgres?

My pgbouncer setup is with fallback connect string
- = host=localhost port=5431
  and transaction pooling

markokr:
Works here.  pgbouncer does not do stuff like "select * from pg_databases", so local config problem?

look at pgbouncer and postges logs to see problem.  and is the port 5431 correct?

aek:
Hi @markokr mark
I still have the issue and my config is the same of pgbouncer default and the port 5431 is setted on postgres config so pgbouncer could use the 5432

markokr:
Ok.  I tried the fallback-db config, create and connect.  And it worked.  So there is something different in your system that makes it fail.  What is it?

sfilargi:
```
I am seeing the same behavior. My code like this:

def db_connect():
    return psycopg2.connect(
        host=pg_host,
        port=pg_port,
        dbname=pg_db, 
        user=pg_user,
        password=pg_pass)


def create_db():
    db = psycopg2.connect(
        host=pg_host,
        port=pg_port,
        dbname='postgres', 
        user=pg_user,
        password=pg_pass)
    db.set_isolation_level(0)
    cur = db.cursor()
    cur.execute('''
        CREATE DATABASE %s
        ''' % (pg_db,))

    db.commit()

    db = db_connect()
```

The second connect fails with: `unix(30884):6432 closing because: pgbouncer cannot connect to server (age=0)`.

I cannot even connect with 'psql --port=6432 newdb" immediately. After a couple of times, it works.

sfilargi:
I am seeing the same behavior. Easily reproducible:

```
$ psql --port 6432 testdb; createdb --port 6432 testdb; psql --port 6432 testdb
psql: ERROR:  database "testdb" does not exist
psql: ERROR:  pgbouncer cannot connect to server
```

aek:
Completing the @sfilargi test I add some wait and reply commands to test when it will be available the db to be used through pgbouncer that it took around 18s to be accesible
```cmd
[...]

PR #346, opened 2018-12-09
Fix to be able to connect to a newly created database

aek:
This fix is to be able to connect to a newly created database without raising the error: "pgbouncer cannot connect to server"

I wanna get some inputs about how we could handle this better covering all the scenarios that could be still needing the error raising

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -598,12 +598,12 @@ bool check_fast_fail(PgSocket *client)
 	cnt = pool_server_count(pool) - statlist_count(&pool->new_server_list);
 	if (cnt)
 		return true;
-	disconnect_client(client, true, "pgbouncer cannot connect to server");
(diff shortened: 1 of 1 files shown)


andrp92:
Can anyone check and merge this PR?

petere:
Can you elaborate on what the problem is? Steps to reproduce? What do you mean by "newly created database"?

aek:
The issue is described at #139 
With "newly created database" I wanna mean that a new database that it have been created very close to the moment of trying to access it using pgbouncer, like in a range of 18s from being created to be accessible using pgbouncer

petere:
Well, just disabling the check isn't going to be an acceptable solution. We need to get to the bottom of the issue.

petere:
Whatever the bug, this clearly isn't the solution.

PR #348, opened 2018-12-14
add application_name_disable setting

aklosslp:
The calls pgbouncer makes to `SET application_name = ` consume about 5% of our database execution time. We'd like an option to disable them.


JelteF:
The varcache stuff got improved a lot in recent versions. So maybe this is not an issue anymore these days. In any case this is too outdated to be mergeable. So closing this.

Issue #350, opened 2018-12-17
pgbouncer statistics can be returned utilizing full uint64 range, but column type is int64


ferringb:
In tracing https://github.com/larseen/pgbouncer_exporter/issues/10 , it was found that pgbouncer internally maintains stats as uint64, but returns that raw value as a postgres int8 field for what it sends over the wire.  int8 has a hard limit of 2^63; thus when a statistic exceeds that (before having a uint64 rollover), this results in a value being serialized that cannot be handled by client libraries.

This behaviour has been confirmed for 1.8.1.

I'm not sure of the best fix; my first thought is to just modify ```pktbuf_write_DataRow``` to force the modulus conversion but that feels wrong/dirty.  The alternative is modifying every usage to ensure that they're forcing modulus to 2^63- which also feels fragile.

petere:
Why can client libraries not handle this? Wouldn't they just see a negative number?

ferringb:
@petere pardon, I put more details into the PR discussion that gives some details.  Reiterating the background; pgbouncer sends stats as OID=20 (int64) which has a signed int64 (long long) range.

To your point "why not have the client library handle this?", you're suggesting that client libraries special case parsing of oid=20 for pgbouncer.  That's suboptimal when one thinks about it that way.  For anything parsing this and strict, they can wind up with negative numbers for stats.  For end users, that's not going to be optimal or obvious.

The thing is, int8 is <2^63 and int4 is <2^31; I already provided a simple patch that just caps the datarow content for this. Those metrics are already going to roll over at some point so any consuming code can already handle that- or should be able to, at least.  I strongly argue the PR is the best way to fix this both for end users, and for development cost.

petere:
Maybe just declare the type to be `numeric`?

ferringb:
@petere what's that ultimately accomplish?  Pgbouncer internal handling of the stats are still going to roll over at 2^64 rather than utilizing the full range of the numeric type.

Making that change will however require client libraries to support that OID- which at least golang lib/pq doesn't.  Python's psycopg2 supports it, and I'd assume ruby bindings are similar; that said, for any code using lib-pg (C/C++) what you're proposing seems like a breaking change.

[...]

PR #360, opened 2019-01-22
Limit pgbouncer statistics to what the wire protocol supports.

ferringb:
Pgbouncer internal statistics are uint8 (64 bit), but the pg protocol is int8;
pgbouncer doesn't roll over the value it writes, instead assuming that the
wire protocol is uint when in fact it's int: see https://www.postgresql.org/docs/9.3/datatype-numeric.html .

To address this, the simplest fix is to force a modulus on what we serialize for uint8
values to limit them to int8 space.

It may be worth refactoring the internals of stats.c/admin.c rather than the
fix I'm proposing here- what is implemented here should cover all stats, but
it feels fragile in the sense that it's a duct tape fix tucked into an internal-
any consuming code that directly uses the underlying pktbuf functions can
bypass and reintroduce a variant of this bug.

I don't see a clean way to address this beyond vigilance combined w/ this
patch, thus proposing this fix.

Backing ticket for this is #350

### DIFF
--- a/src/pktbuf.c
+++ b/src/pktbuf.c
@@ -391,8 +391,8 @@ void pktbuf_write_RowDescription(PktBuf *buf, const char *tupdesc, ...)
  * send DataRow.
  *
  * tupdesc keys:
- * 'i' - int4
- * 'q' - int8
+ * 'i' - int4, limited to 2^31
+ * 'q' - int8, limited to 2^63
  * 's' - string
  * 'T' - usec_t to date
  */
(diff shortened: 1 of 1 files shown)


eulerto:
@ferringb The PostgreSQL protocol isn't limited to int8 (see [1]). PgBouncer is limiting that data to int4 or int8. It seems what you want is change the tupdesc key from _i_ to _q_.

[1] https://www.postgresql.org/docs/11/protocol-message-formats.html

ferringb:
@eulerto pardon, but it's not 2^32 vs 2^64- the problem is that pgbouncer is exceeding the OID range it sets for stat columents.  Protocol breakdown for one of the columns is thus from the RowDescription message:
```table ID: 0
column ID: 0
OID: 20
data type size: 8
type modifier: 0
format code: 0
```
OID20 is int8- per https://www.postgresql.org/docs/9.3/datatype-numeric.html that is limited to 63 bits; in tracing the postgres code I also find multiple capping functions (including in postgres's backend/utils/adt/int8.c that all use int64- signed, not unsigned).   I'm pretty sure my analysis is correct here on that angle, but I'd appreciate a second set of eyes should I have missed something.

see https://github.com/larseen/pgbouncer_exporter/issues/10 for the context this was caught via- pgbouncer_exporter enforces that int8 OID's are type int64 (long long) and was catching that it was being sent a value outside that range- 17606157912109483021, which is between 2^63 and 2^64.

I multiple clients tolerate int8 exceeding 2^63; that said, golang lib/pg doesn't have that toleration, an int8  is a signed 64bit integer.

Capping the pgbouncer statistics to 2^63 resolves this either way.

stanhu:
We ran into the same issue: PgBouncer exports a 64-bit unsigned integer, but PostgreSQL only supports 64-bit signed integers, so the lib/pq client library flagged the errors here: https://github.com/lib/pq/blob/90697d60dd844d5ef6ff15135d0203f65d2f53b8/encode.go#L114

I suggest we either:

1. Merge this pull request to cap the values as this pull request does
1. Change the exported PgBouncer types of the 64-bit integers to use NUMERIC

petere:
I think the best solution would be to use `numeric`.

PR #362, opened 2019-02-03
Extend max password length from 128 to 2048

williamdenton:
When using AWS postgres RDS or Aurora it is possible to authenticate using a cryptographically signed token using IAM credentials. The password generated for connection is between typically about 800 characters long.

The current buffer of 128 is inadequate and results in 
>`WARNING bug in src/objects.c:461 - string truncated`

https://github.com/pgbouncer/pgbouncer/blob/dc226613ca023246684efc24b755e00b61ca6d84/src/objects.c#L449-L464

Passwords can be generated for AWS postgres using the [aws cli](https://docs.aws.amazon.com/cli/latest/reference/rds/generate-db-auth-token.html)

```
aws rds generate-db-auth-token  \
  --hostname clustername.cluster-identifierus-west-2.rds.amazonaws.com \
  --port 5432 \
  --username db_user \
  --region us-west-2
```

Further reading on [AWS IAM authentication for Postgres](https://aws.amazon.com/about-aws/whats-new/2018/11/amazon-aurora-postgresql-supports-iam-authentication/)

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -110,7 +110,7 @@ extern int cf_sbuf_len;
 /* to avoid allocations will use static buffers */
 #define MAX_DBNAME	64
 #define MAX_USERNAME	64
-#define MAX_PASSWORD	128
+#define MAX_PASSWORD	2048
 
 /* no-auth modes */
 #define AUTH_ANY	-1 /* same as trust but without username check */
--- a/include/pktbuf.h
+++ b/include/pktbuf.h
@@ -144,6 +144,6 @@ void pktbuf_write_ExtQuery(PktBuf *buf, const char *query, int nargs, ...);
 	SEND_wrap(16, pktbuf_write_CancelRequest, res, sk, key)
 
 #define SEND_PasswordMessage(res, sk, psw) \
-	SEND_wrap(512, pktbuf_write_PasswordMessage, res, sk, psw)
+	SEND_wrap(MAX_PASSWORD, pktbuf_write_PasswordMessage, res, sk, psw)
 
 void pktbuf_cleanup(void);


williamdenton:
the error message i was seeing would have been improved with  https://github.com/pgbouncer/pgbouncer/commit/645a4c5a6159e69cfa4bf225fe911d5116502e70 but this only improves the error message and doesn't resolve my problem

PJMODOS:
It's kind of similar problem like the neighbouring one with the maximum username length, we use the static length for in memory representation (fixed length for each user) so increasing is not for free. I am not sure how I feel about making it bigger, especially this much. Both of these seem to be problem for cloud uses so we better do something about it, not sure if just plainly increasing the size is the way to go though. Will need to measure the impact on memory usage.

williamdenton:
a compile time switch perhaps?

milesrichardson:
What is the reasoning for static allocations of username/password? Is it for performance, or safety?

hipyhop:
@williamdenton How are you using RDS IAM auth with pgbouncer? Are you passing this through from the client somehow, or embedding this in the userlist.txt?
I'm trying to make this work as a proxy between users and RDS, but can't understand how to pass the password through from the client.

williamdenton:
@hipyhop 
>How are you using RDS IAM auth with pgbouncer?

im running pgbouncer as a sidecar container. My entrypoint script runs `aws rds generate-db-auth-token` in a loop every 5 minutes or so and updates the `pgbouncer.ini` you can then do `kill -SIGHUP pgbouncerpid` to make pgbouncer reload the new configuration with new passwords gracefully.

Pgbouncer is configured to allow any user/password when connecting, you just have to specify the database (it's running as a side car so security here doesn't matter as pgbouncer does not expose ports outside of the task).


Major caveat: RDS dns load balancing is completely incompatible with pgbouncer load balancing.
rds dns returns a new ip address each time you query rather than returning the set of reader only nodes.
pgbouncer requires a list of ip addresses be returned from dns, any addresses that are not returned in a subsequent query are considered dead and it will close those connections ASAP. You end up fixated on one rds readonly node at a time rather than balancing your traffic.

geektuhin123:
Even if we manage to generate RDS auth-token how to solve the password length issue?

williamdenton:
>how to solve password length issue?

You have to fork and apply this patch. That’s what I’ve done.

petere:
As discussed in #374, we should change the static arrays to malloced arrays instead of making them bigger and bigger. It's unlikely that 2048 will be enough for long.

Also, PostgreSQL client tools don't support such long passwords consistently. That is being fixed in PostgreSQL 14.

williamdenton:
stale - closing.

PR #363, opened 2019-02-05
Add a RECONNECT_CLIENTS [db] admin command

film42:
This command is similar to `RECONNECT [db]`, however, instead of only closing the `pgbouncer<->server` connection, this closes the `client<->pgbouncer` connection once the current statement or transaction completes. It does not handle clients in session mode, since the `RECONNECT [db]` command handles that case when `server_fast_close = 1`.

We'd really like the behavior so we can rebalance connections for a particular database without needing to perform a rolling restart of our application servers. You can already get similar behavior by gracefully stopping pgbouncer, however, we'd like to issue this command and have all pgbouncers available to receive connections (instead of the stopped server ending up with no connections). Basic testing shows that active_record should be able to handle connection restarts like this just fine, but YMMV.

NOTE: I left a fairly long revision history, but I'm more than happy to rebase. I first want to get a general :+1: / :-1: . I'm happy to make any changes. Just leave a comment and consider it done. Thanks! :)


film42:
Looks like the docs switched from rst to markdown. I merged those changes and updated the docs that come with this PR.

petere:
I don't understand the use case. Why do you want to make pgbouncer actively close client connections? What are the clients then supposed to do? Won't they just try to reconnect?

film42:
As we perform rolling restarts we often end up with the last pgbouncer with no connections and the rest with an even balance of all the connections. Similarly, when we add a new pgbouncer server to the cluster we end up in a similar situation.  When this happens, we need to rely on application restarts to balance connections. This would allow us to gracefully disconnect the client and rebalance our client connections without relying on application restarts.

This type of command isn’t used very often, but when under fire, say, too few pgbouncer severs, each with high cpu, it’s nice to be able to admin without restarting hundreds of application servers.

davidfetter:
Can we get a rebase, @film42 ?

film42:
Absolutely. I'll try to get that done in the next few days.

film42:
@davidfetter Just merged master branch and resolved conflicts. Had to make some changes to the test setup but it should be passing again.

davidfetter:
@film42 thanks! Apparently, something broke in CI, but it's not clear to me whether it was the patch or the CI breaking.

film42:
It looks like the new test is failing for 2 configurations on travis. Looks like a user: pgbouncer connection is open through the tests. I'll see if I can't make my new test a bit more robust.

src/janitor.c
```diff
@@ -366,6 +366,9 @@ static void pool_client_maint(PgPool *pool)
 	PgSocket *client;
 	usec_t age;
 
+	/* find and disconnect clients if the need to be closed */
+	for_each_client(pool, disconnect_client_if_close_needed);
```
film42, line 370: Although this should only need to run in the release_server method, I think it might make sense to run this as part of the janitor. I initially only had logic here but realized I need to be checking this as soon as a transaction is finished. Thoughts?

JelteF:
Closing this as it is too outdated to be mergeable. Feel free to open a new PR with an updated version of this.

PR #378, opened 2019-03-28
Make PAUSE return when canceled by interrupt, signal, or other resume command.

groner:
This PR makes interrupting a `PAUSE` or `WAIT_CLOSE` request work a little better and it improves the tracking of those requests to reduce the likelihood of interference between concurrent commands.

### DIFF
--- a/include/admin.h
+++ b/include/admin.h
@@ -26,3 +26,4 @@ bool admin_flush(PgSocket *admin, PktBuf *buf, const char *desc) /* _MUSTCHECK *
 bool admin_ready(PgSocket *admin, const char *desc)  _MUSTCHECK;
 void admin_handle_cancel(PgSocket *client);
 void admin_cleanup(void);
+void admin_cancel_all_pauses(PgDatabase *db);
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -88,6 +88,12 @@ enum SSLMode {
 	SSLMODE_VERIFY_FULL
 };
 
+enum WaitCommand {
+	WAIT_CMD_NONE = 0,
+	WAIT_CMD_PAUSE = 1,
+	WAIT_CMD_WAIT_CLOSE = 2
+};
+
 #define is_server_socket(sk) ((sk)->state >= SV_FREE)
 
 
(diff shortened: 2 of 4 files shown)


petere:
I don't understand what this does. Your explanation could use a lot more information and details.

Please squash your commits into sensible pieces.

groner:
Hi Peter,

Thanks for looking at this.  I will try to clarify what this patch does and why.

I want to be able to issue a `PAUSE` and cancel it if connections don't drain in a certain time window.  Currently this is possible using signals, but not using the admin client.  Using the admin client is preferable because it gives feedback letting me know when the database is fully paused.  Database scoping is likely useful for some deployments, though I have no immediate need for it.

This branch also makes similar changes to `WAIT_CLOSE`.  This is only done to be consistent I have no immediate use for it.

Cancellation of `PAUSE` commands is partially implemented already.  The `PAUSE` is canceled, but no feedback is sent to the client which will wait forever or until the user kills it.

These changes facilitate notifying the admin client when their `PAUSE` command is canceled or invalidated by signal or `RESUME` from another client.

groner:
Here is some explanation of how the commits are staged.  If it doesn't seem useful, I'm happy to change it.

> c92ba14 Normalize whitespace.

This makes some existing code consistent with the rest of the code base.  I try to separate formatting changes because it makes the commits that follow less noisy.

> bc54225 Track pending admin command and scope explicitly.

This is all book keeping.  In order to notify the admin client that issued a PAUSE that it has been invalidated we need a reference to it.

> 1ed0c77 Prevent two admins from PAUSEing on the same database.

This is a behavior change.  We don't want to loose track of the existing PAUSE issuer by overwriting it. 

> 80f925c Complete the admin CancelRequest circuit.

Here we notify the admin client that the command is canceled, so it doesn't hang.

> 955d405 Make RESUME and SIGUSR2 cancel pending PAUSE requests.

Here we notify the admin client that the command is invalidated by another client or signal, so it doesn't hang.

davidfetter:
Can we get a rebase, @groner ?

JelteF:
Closing this as it is too outdated to be mergeable. Feel free to open a new PR with an updated version of this.

PR #389, opened 2019-05-05
Make all the help into rowsets

davidfetter:
...so other tools can see them more easily.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -1285,29 +1285,33 @@ static bool copy_arg(const char *src, regmatch_t *glist,
 
 static bool admin_show_help(PgSocket *admin, const char *arg)
 {
-	bool res;
-	SEND_generic(res, admin, 'N',
-		"sssss",
-		"SNOTICE", "C00000", "MConsole usage",
-		"D\n\tSHOW HELP|CONFIG|DATABASES"
-		"|POOLS|CLIENTS|SERVERS|USERS|VERSION\n"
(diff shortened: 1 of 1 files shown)


davidfetter:
#321 is included in this patch.

petere:
Unclear why this is wanted. Why does a tool need to process the help in an automated way?

davidfetter:
For example, [this](https://github.com/davidfetter/pgbouncer_wrapper)

davidfetter:
Ping!

petere:
I'm rejecting this. Help information is not data.

davidfetter:
Perhaps I was unclear as to what I attempted to accomplish.

pgbouncer commands represent, in some sense, a schema wherein there are functions some of which return sets, or one could look at them as an API. I was attempting to make said schema/API available to other software of which my `pgbouncer_wrapper` and Crunchy Data's `pgbouncer_fdw` are two examples. What I've been doing up to this point is groveling through the `pgbouncer` source using a Perl program that could be generously described as being composed of chewing gum and baling wire in order to unearth this schema/API.

Would you consider a patch set that implements a different approach to this issue? Two approaches I've considered are:
- Add one function called `describe_api` that sends back something along the lines of a `JSON`blob describing enough of those functions in order for other software to make reasonably correct inferences about what is available.
- Expose more catalog-like information in a more catalog-like form, the extreme end of which would include `information_schema`. This seems like an enormous amount of work for very little benefit.

PR #390, opened 2019-05-28
Implement PROXY protocol v1

jouir:
Detect PROXY protocol packets header and parse it at client startup. New configuration setting "proxy_protocol_networks" with a comma-separated list of networks to trust to avoid IP spoofing.

**Important note**: this is my very first pull-request in C. I'm sure it's not perfect.

There are several issues with this pull-request:
* IP spoofing: networks are set but not checked. I'm having a hard time trying to tokenize the comma-separated list of string, convert them to networks and try to compare with the received address.
* PROXY protocol v2 is not supported. I can't manage to parse the binary protocol.
* code organization may not be good: it may not be at the good place, functions may be used instead, etc.
* missing tests?

With this patch, I'm able to have a transparent pgbouncer behind an HAProxy with the version 1 of the protocol. Help to improve this pull-request will be appreciated.


petere:
This looks interesting, but I think there should be an agreement across the PostgreSQL ecosystem to adopt this. Just implementing it in PgBouncer and only on one side seems pretty limited.

We should also consider how this would interact with pg_hba.conf. The assumption of the proxy protocol is that an incoming connection would be treated with the IP address of the address that the proxy protocol provided. This might include pg_hba.conf processing.

The setting `proxy_protocol_networks` isn't actually used in your code AFAICT. Again, questions here about how this should interact with pg_hba.conf.

jouir:
> This looks interesting, but I think there should be an agreement across the PostgreSQL ecosystem to adopt this. Just implementing it in PgBouncer and only on one side seems pretty limited.

Here is the [thread](https://www.postgresql.org/message-id/flat/20190519153623.GB13208%40thinkpad) on pgsql-hackers mailing list. From my understanding, the community doesn't understand why we should implement a new sort of protocol inside core to have a transparent proxy when you already have PgBouncer that does it for you. I'm not agree with this idea. PgBouncer itself can't be fully transparent if the backend doesn't support it. Yes, there are workaround like adding the source IP address in the application name but it's not transparent.

My idea is to support PROXY protocol between clients and PgBouncer first so we can consider clients behind an HAProxy like a direct client. Then, we'll can look forward to implement it between PgBouncer and PostgreSQL core. The first step is already hard for my limited C skills, the second is way above my league.

> We should also consider how this would interact with pg_hba.conf. The assumption of the proxy protocol is that an incoming connection would be treated with the IP address of the address that the proxy protocol provided. This might include pg_hba.conf processing.

IMHO, to be fully transparent, a client behind the proxy implementing PROXY protocol should be considered as a direct client, host based authentication included. It could be a security issue, this is why I propose to restrict clients based on a trusted network range (where the proxy lives).

> The setting `proxy_protocol_networks` isn't actually used in your code AFAICT. Again, questions here about how this should interact with pg_hba.conf.

I can't managed to successfully implement this setting, simply because I failed to check if a host IP address belongs to a network range. Some help would be appreciated.

takeda:
@petere PROXY protocol is just a way to pass to the application the 4 components that make a connection (source address, source port, dest address, dest port) this is necessary because once application like pgbouncer is behind a layer 4 load balancer, in the logs, stats page, and even pg_hba.conf all you will see is the load balancer IP. With that mechanism applicatioon behaves as if there is no load balancer in front of it.

PostgreSQL protocol is binary and unlike HTTP you can't use headers like X-Forwarded-Hosts, but frankly PROXY makes a better experience even when used in HTTP.

It's also very simple basically on connection, the load balancer sends a first line that looks like this:
`PROXY TCP4 192.168.0.1 192.168.0.11 56324 5432\r\n`

Then communication continues as usual, but application such as pgbouncer now knows the connection was made from 192.168.0.1:56324 and uses that as IP/port for logging/stats/pg_hba.conf. Yes, this enables ability to spoof connection, but ops person can configure firewall to only allow connections by a load balancer, and looks like even this PR also provides an option to configure which IPs to trust.

This is useful, since proxy protocol is supported by popular load balancers such as ELB or HAProxy. I also was able to create script on F5 BigIP that provided that information even when originally the load balancer did not support it.

petere:
[Here](https://www.postgresql.org/message-id/flat/CABUevExJ0ifpUEiX4uOREy0s2kHBrBrb=pXLEHhpMTR1vVR1XA@mail.gmail.com) is more recent proposal for implementing the PROXY protocol in PostgreSQL. If that gets accepted, then an implementation for PgBouncer can also be considered.

takeda:
@petere there are two parts where the PROXY protocol could be used.

- First one is from layer 4 load balancer to pgbouncer. I believe this ticket was talking about that
- and from pgbouncer to PostgreSQL. This would be great as well, although as you pointed out a functionality needs to be added to PostgreSQL.

PR #393, opened 2019-06-14
Fix a bunch of issues around auth_user

### LINKED ISSUE #391: Invalid/incoherent behaviours with the global auth_user option
Hello

When the global auth_user option was introduced in commit 4ea2306b94647333a9ed8a62b897cd872e3f0005, it seems a lot of cases were not handled properly.

1) The option works only if it is defined *before* the databases
2) If the option is defined twice (for instance if it is defined in the main file and later in an included file), the value used will be the last one preceding the databases definitions, but the value displayed in `show config` will be the second one
3) When calling `SET auth_user = marcel`, the newly-set value is not applied unless you reload and did not define it in your configuration before the databases
4) If you change the option in the configuration and issue a `RELOAD`, the value applied will, again, depend on the location in the configuration file : if it is after the databases, a second `RELOAD` will be required to apply the change.

I suggest completely removing the current application of the config auth_user in parse_database in loader.c, and instead, in each code path that tries to use `db->auth_user`, add a second if branch to use the `cf_auth_user` instead. If these changes would be fine for you, I can implement them and submit a PR.

Regards

pinaraf:
When switching a pgbouncer configuration from using userlist.txt files to using auth_query/auth_user, I encountered several issues, all fixed in this PR.

1. the wait_start field is not always initialized when using auth_query, displaying weird random stats
Example:
```
2019-06-14 15:41:07.500 26350 LOG stats: 0 xacts/s, 0 queries/s, in 0 B/s, out 9 B/s, xact 0 us, query 0 us, wait 183224844063 us
```

2. the global auth_user configuration variable is mis-handled: it has to be defined before the databases in the .ini file, it can not be defined using the pgbouncer shell unless a reload is applied… This is fixed here by simply not using it from parse_database and instead having it be checked when needed.

3. when removing users from userlist.txt, the corresponding objects are not removed from memory. Instead, their passwords are emptied. This prevents auth_query from being called since the if is valid only for non-existing users.

close #391

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -322,7 +322,7 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 				client->db->auth_user = add_user(cf_auth_user, "");
 			}
 		}
-		if (!client->auth_user && client->db->auth_user) {
+		if ((!client->auth_user || (client->auth_user->passwd[0] == '\0')) && client->db->auth_user) {
 			if (takeover) {
 				client->auth_user = add_db_user(client->db, username, password);
 				return finish_set_pool(client, takeover);
@@ -404,10 +404,14 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
 			length = sizeof(user.passwd) - 1;
 		memcpy(user.passwd, password, length);
 
-		client->auth_user = add_db_user(client->db, user.name, user.passwd);
-		if (!client->auth_user) {
-			disconnect_server(server, false, "unable to allocate new user for auth");
-			return false;
+		if (client->auth_user) {
+			memcpy(client->auth_user->passwd, password, length);
+		} else {
+			client->auth_user = add_db_user(client->db, user.name, user.passwd);
+			if (!client->auth_user) {
+				disconnect_server(server, false, "unable to allocate new user for auth");
+				return false;
+			}
 		}
 		break;
 	case 'N':	/* NoticeResponse */
--- a/src/client.c
+++ b/src/client.c
@@ -316,6 +316,12 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 	} else {
 		/* the user clients wants to log in as */
 		client->auth_user = find_user(username);
+		if (!client->auth_user && !client->db->auth_user && cf_auth_user) {
+			client->db->auth_user = find_user(cf_auth_user);
+			if (!client->db->auth_user) {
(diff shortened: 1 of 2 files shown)


petere:
I have committed a fix for the first issue, but with the code a bit different. I'll continue to work through the other issues.

pinaraf:
Hello @petere

Any news regarding merging fixes for the two other issues? To be honest, the total_wait_time was no the one bothering us the most…

Regards

gplv2:
concerning nr 3 of the OP's list:

I like to bump this fix due to walking under a ladder, putting a black cat over my left shoulder ... and oh.. having spent a good 2 days trying to understand why new users in a DB got authenticated using the query and existing ones never went to the database after removing them from the userslist.txt file.

It's what I'm missing to gradually migrate our countless databases to auth_query from a unmanageable userslist.txt file , it would make a transition so much easier and without interruptions.    It really seems like a good fix to me and it would even be so much more awesome if the docs actually mentioned that migrating to auth_query has a few drawbacks.   Eventually this issue was the one that brought me the A-HAAA moment right before I was about to jump from the bridge.

Thanks to Pierre, I'm still here today to complain and bump stuff! ;-)

Seriously though, it would be a nice one to merge.  before it dies from old age that is.  Thanks for all the work, pgbouncer is still awesome.

davidfetter:
Bump!

pinaraf:
I've done a rebase so only the remaining parts are shown here.

davidfetter:
Thanks!

petere:
The second item has been fixed now. (In the future, I would recommend sending separate pull requests for separate issues. Otherwise it makes reviewing really hard if you don't know what change applies to what issue. More so if the fixes are right next to each other.)

petere:
Looking at the third item, I'm not sure about the fix. Having a user with an empty password is a valid state, so we can't just take that as license to override the user with an auth_query user. I think we'd either need to actually delete the user if it disappears from auth_file, or at least flag it as invalid in a different way.

gplv2:
Hi Petere,

I think deleting the user is correct,   I also figured it would be a cool feature to be able to purge all/certain users from memory using a pgbouncer command.  It sure would assist debugging auth problems and help transitioning between both auth systems.  Thanks for the work.

JelteF:
I'm closing this since #1052 should have fixed all issues with auth user, and any that were not fixed probably need a different fix now.

PR #401, opened 2019-08-03
Convert all stats from INT8 -> NUMERIC(20, 0)

stanhu:
Internally these statistics are stored as unsigned 64-bit values,
but the PostgreSQL INT8 types can only express signed 64-bit values.
Convert all the stats to NUMERIC(20, 0) to avoid losing data.

Closes #360

### DIFF
--- a/include/system.h
+++ b/include/system.h
@@ -52,6 +52,10 @@
 #define INT8OID 20
 #define INT4OID 23
 #define TEXTOID 25
+#define NUMERICOID 1700
+
+/* NUMERIC(20, 0) */
+#define NUMERIC_TYPEMOD 0x140005
 
 /*
  * libc compat functions.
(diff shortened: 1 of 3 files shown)


stanhu:
@petere Would you mind reviewing?

stanhu:
@petere Friendly ping. We got bit by recently by a problem with not having metrics for a subset of our PgBouncer hosts. Can we look into rolling this in soon?

petere:
I want to go through all the admin and stats commands and check for similar data type mismatch/overflow problems. I think there are a few more. It will need a bit more time.

petere:
a fix has been committed

stanhu:
@petere Thanks, could you cut a new release for PgBouncer?

petere:
yes :)

PR #416, opened 2019-09-06
feat: extend max username and db_name

ealogar:
The default max value for username allowed for login with pgbouncer is too short (64 characters). this limit does not exits in many postgres versions so we increase to 128 the value.
Additionally the max_db_name is increased accordingly as this value is also affecting.
This problem has arises in azure dbaas as it includes an @<server-name> in login that is easilly overcoming ...


petere:
Please clean up your commit history and remove unrelated changes.

ealogar:
> Please clean up your commit history and remove unrelated changes.

Already done!

palmerabollo:
Is there anything we can do to help merge this feature? We're also facing many issues with Azure usernames due to this limitation in the username's max length.

yshapkin (approved):

PR #422, opened 2019-10-03
Add PAM support to hba.conf

svb007:
I have included this change as a patch in a custom version since the beginning of 2018, and it has run without problems in a production environment of over 80 servers (See issue # 276). I am requesting this change to be included in the next version of pgbouncer


davidfetter:
Bump

petere:
See #326 why it's more complicated.

petere:
I'm closing this because it's essentially a duplicate of #326 and there is more in-depth discussion there.

Issue #376, opened 2019-03-22
Logs spammed with FIXME: query end, but query_start == 0 with npgsql


dezconnect:
Faced with that problem.

Solution from https://github.com/pgbouncer/pgbouncer/issues/277 not works.

I tried to use No Reset On Close=True in application but it is not solved that trouble.

DavidPavlicek:
In our case, this is caused by the command timeout on npgsql connection. Each time npgsql terminates command due to timeout, this appears in our log. So... try to avoid command timeouts on a client side.

dezconnect:
> In our case, this is caused by the command timeout on npgsql connection. Each time npgsql terminates command due to timeout, this appears in our log. So... try to avoid command timeouts on a client side.

Thanks, but it is not works for our case...

dezconnect:
Solved that problem via change log level for this warning and rebuild package for local repo.

kleptog:
This is possibly fixed by #253.

dezconnect:
> This is possibly fixed by #253.

No, I used last version of pgbouncer with that fix, but I faced with warnings. 
How I understood this linked with processing of non-blocking connections.

fedordikarev:
Hi everyone! Are there any updates on this issue?
We also use pgbouncer with npgsql and have this errors. We reproduce them on our QA environment (rate is several errors per second). So we could easily get any debug log or trace if needed. Just tell me which log level to use or what debug information could help you to investigate real root of this issue? (sorry, we don't have developers with enough knowledge of pgbouncer in our team, so we can't debug this issue by ourselves)

dezconnect:
You can find lines of code with these warnings and fix it to debug level. After that, you need to rebuild the package and use it.

kpakozz96pyc:
Also get the same error in pgbouncer-1.9.0-win32

svb007:
So have you tried the patch I put at the top of https://github.com/pgbouncer/pgbouncer/issues/277 ? that does at least get rid of the messages. I agree, https://github.com/pgbouncer/pgbouncer/pull/253 does not seem to fix this issue, but does seem to make npgsql pools behave better when connecting to pgbouncer

dezconnect:
> So have you tried the patch I put at the top of #277 ? that does at least get rid of the messages. I agree, #253 does not seem to fix this issue, but does seem to make npgsql pools behave better when connecting to pgbouncer

Yep, I compiled it with debug log level and it fixed this problem.

[...]

Issue #277, opened 2018-01-23
Logs spammed with FIXME: query end, but query_start == 0 and now also FIXME: transaction end, but xact_start == 0


svb007:
I've seen that this was suggested before, and I attach the patch that fixed this issue for me (by downgrading the log priority to debug)

[fixme_logspam.patch.txt](https://github.com/pgbouncer/pgbouncer/files/1657926/fixme_logspam.patch.txt)

topharley:
In my case I could see errors in PGBouncer's log:

  FIXME: query end, but query_start == 0
  FIXME: transaction end, but xact_start == 0

And from NPGSQL exceptions:
  Npgsql.NpgsqlException (0x80004005): Exception while reading from stream ---> System.IO.EndOfStreamException: Attempted to read past the end of the stream. at Npgsql.ReadBuffer.<Ensure>d__27.MoveNext()...

And finally it was fixed by setting "No Reset On Close=True" in connection string witch is described here https://www.npgsql.org/doc/performance.html

Errors occured in only one case:
1) Using client pooling build in npgsql
2) Using PGBouncer for server pooling
3) PGBouncer used in "transaction" mode
4) Using UPDATE queries in called stored procedure (SELECT are working fine)
5) Multithreading client >=10 threads in app
6) Sending string parameter to SP with length >= 4000 chars (small strings are working fine)

Same behavior on 3.2.7 and 4.0.3

Disabled code:
if (!Settings.NoResetOnClose && DatabaseInfo.SupportsDiscard)
{
	if (PreparedStatementManager.NumPrepared > 0)
	{
		...
	}
	else
	{
		// There are no prepared statements.
		// We simply send DISCARD ALL which is more efficient than sending the above messages separately
		PrependInternalMessage(PregeneratedMessage.DiscardAll); // <-- THIS WILL NO BE CALLED!!!!!!!!!!!!!!!!!!!!!!!!!
	}
} 

I guess this param must be always switched to "No Reset On Close=True" when using PGBouncer in transaction mode because PGBouncer switches connection for each transaction so logic of DISCARD ALL looses sense. Correct me pleas if i am wrong.

kleptog:
This is possibly fixed by #253.

wanziforever:
> 
> 
> This is possibly fixed by #253.

the problem is still exist, not ony under async mode, if a request  contains multiple transaction requests, and the server will receive mutiple response sequencely, and the problem will be reproduced.

[...]

PR #424, opened 2019-10-03
Lower severity of protocol errors produced by npgsql

svb007:
There is currently no way to exclude these log warnings from the log file using the verbose parameter when using npgsql as client.
This error message should really be classified as a debug message, as it is not very useful as a warning, but has a lot more value as a debug message.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -361,7 +361,7 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 				server->pool->stats.query_time += total;
 				slog_debug(client, "query time: %d us", (int)total);
 			} else if ((ready || idle_tx) && !async_response) {
-				slog_warning(client, "FIXME: query end, but query_start == 0");
+				slog_debug(client, "FIXME: query end, but query_start == 0");
 			}
 
 			/* statement ending in "idle" ends a transaction */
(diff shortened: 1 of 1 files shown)


petere:
If you have a way to reproduce these messages, then I'd be interested to fix them rather than hide them.

svb007:
The only way I know to reproduce this, is here : https://github.com/npgsql/npgsql/issues/2522

I can see at least another 2 years of having to patch pgbouncer to switch off these messages before any future npgsql fix will reach my production environment. I suppose the problem is on the client side, but it would still be nice to be able to silence these messages until such time as the client is fixed.

svb007:
Hi Peter, it seems that the npgsql maintainer has posted some data related to this issue, as well as a way to reproduce it here: https://github.com/pgbouncer/pgbouncer/issues/376

hacst:
@petere The linked issues contain a reproducer as well as a package dump (https://github.com/pgbouncer/pgbouncer/issues/277#issuecomment-599139014). Anything else that can be done to help with a real fix? This log spam is a real issue on its own so if isn't going to get fixed because it isn't important it should at least be hidden by default as proposed here.

petere:
This was fixed by d7f5f3d693204857db7c186466bc3f53697fece4.

PR #428, opened 2019-10-11
Patches pgbouncer to add TCP_USER_TIMEOUT option

sdemontfort:
### The issue:
TCP connections that are in the `ESTABLISHED` state to a Postgres server don't get closed properly if the Postgres server dies and doesn't send back any TCP closing packets. This means that PgBouncer ends up holding onto dud connections, and they won't be closed until the Linux kernel level setting `tcp_retries2` kicks in _(at around 15 minutes)_.

### Description of changes:
This change introduces the TCP socket-level setting `TCP_USER_TIMEOUT`, which acts the same as `tcp_retries2` but at the socket level, instead of being OS-wide. It kicks in after the specified keep alive probes, explanation here:

>... We mentioned the TCP_USER_TIMEOUT option before. It sets the maximum amount of time that transmitted data may remain unacknowledged before the kernel forcefully closes the connection. On its own, it doesn't do much in the case of idle connections. The sockets will remain ESTABLISHED even if the connectivity is dropped. However, this socket option does change the semantics of TCP keepalives. The tcp(7) manpage is somewhat confusing:
...
>For the user timeout to have any effect, the icsk_probes_out must not be zero. The check for user timeout is done only after the first probe went out. Let's check it out. Our connection settings:

>TCP_USER_TIMEOUT = 5*1000 - 5 seconds
>SO_KEEPALIVE = 1 - enable keepalives
>TCP_KEEPIDLE = 1 - send first probe quickly - 1 second idle
>TCP_KEEPINTVL = 11 - subsequent probes every 11 seconds
>TCP_KEEPCNT = 3 - send three probes before timing out

#### Note:
`TCP_USER_TIMEOUT` interacts with `keepalive` settings in a particular way: https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/

### Steps to test:
**1.** Set the following config options for PgBouncer:
```
tcp_user_timeout = 12500
tcp_keepalive = 1
tcp_keepidle = 1
tcp_keepintvl = 11
tcp_keepcnt = 3
```
this will ensure that dead TCP connections are killed after 12.5s.
**2.** Start PgBouncer and connect to Postgres, make sure there are some server connections open
**3.** psql to PgBouncer and run `SHOW SERVERS;`, make sure there's a connection and monitor it's `connect_time` value.
**4.** Block packets from getting to Postgres, assuming Postgres is running on port `5432`
```
iptables -A OUTPUT -p tcp --dport 5432 -j REJECT
```
**5.** After 12.5s you should see that the connection is killed via `SHOW SERVERS;`
**6.** To show the issue, repeat the above but remove the `tcp_user_timeout` config, and the server connection should hang for about 15mins.

_P.S_ I'm not a C developer so please excuse any wrong includes– commit a fix if you see obvious issues 😃

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -787,6 +787,12 @@ Default: 45 on Linux, otherwise 0
 
 Default: not set
 
+### tcp_user_timeout
+
+Enables socket-level timeout in milliseconds. This solves the situation where a database server on the end of a connection fails to kill a TCP connectionn properly after a shutdown or crash. TCP connections then get stuck and need clearing out. The Linux kernel setting `tcp_retries2` default value means stuck TCP connections get cleaned up after about 15 minutes. Setting `tcp_user_timeout` to a lower value means the connections get cleaned up more quickly if they're stuck.
+
+This setting works alongside the below keep alive settings. Read more about specifics here: https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/
+
 ### tcp_keepalive
 
 Turns on basic keepalive with OS defaults.
--- a/etc/pgbouncer.ini
+++ b/etc/pgbouncer.ini
@@ -304,6 +304,10 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; In-kernel buffer size (Linux default: 4096)
 ;tcp_socket_buffer = 0
 
+;; Close TCP connections that are falsely stuck in "ESTABLISHED" state
+;; after this much time in milliseconds
+;;tcp_user_timeout = 0
+
 ;; whether tcp keepalive should be turned on (0/1)
 ;tcp_keepalive = 1
 
--- a/src/main.c
+++ b/src/main.c
@@ -78,6 +78,7 @@ int cf_sbuf_len;
 int cf_sbuf_loopcnt;
 int cf_so_reuseport;
 int cf_tcp_socket_buffer;
+int cf_tcp_user_timeout;
 #if defined(TCP_DEFER_ACCEPT) || defined(SO_ACCEPTFILTER)
 int cf_tcp_defer_accept = 1;
 #else
@@ -272,6 +273,7 @@ CF_ABS("tcp_keepalive", CF_INT, cf_tcp_keepalive, 0, "1"),
 CF_ABS("tcp_keepcnt", CF_INT, cf_tcp_keepcnt, 0, "0"),
 CF_ABS("tcp_keepidle", CF_INT, cf_tcp_keepidle, 0, "0"),
 CF_ABS("tcp_keepintvl", CF_INT, cf_tcp_keepintvl, 0, "0"),
+CF_ABS("tcp_user_timeout", CF_INT, cf_tcp_user_timeout, 0, "0"),
 CF_ABS("verbose", CF_INT, cf_verbose, 0, NULL),
 CF_ABS("admin_users", CF_STR, cf_admin_users, 0, ""),
 CF_ABS("stats_users", CF_STR, cf_stats_users, 0, ""),
--- a/src/util.c
+++ b/src/util.c
@@ -142,6 +142,17 @@ bool tune_socket(int sock, bool is_unix)
 		goto fail;
 
 	/*
+	 * TCP user timeout
+	 */
+#ifdef TCP_USER_TIMEOUT
+	val = cf_tcp_user_timeout;
+	errpos = "setsockopt/TCP_USER_TIMEOUT";
+	res = setsockopt(sock, SOL_TCP, TCP_USER_TIMEOUT, &val, sizeof(val));
+	if (res < 0)
+		goto fail;
+#endif
+
+	/*
 	 * set in-kernel socket buffer size
 	 */
 	if (cf_tcp_socket_buffer) {
--- a/test/test.sh
+++ b/test/test.sh
@@ -172,6 +172,17 @@ fw_reject_port() {
 		echo "Unknown OS"; exit 1;;
 	esac
 }
+fw_reject_port_silent() {
+	case `uname` in
+	Linux)
+		sudo iptables -A OUTPUT -p tcp --dport $1 -j REJECT;;
+	Darwin|OpenBSD)
+		echo "block out proto tcp from any to 127.0.0.1 port $1" \
+		    | sudo pfctl -a pgbouncer -f -;;
+	*)
+		echo "Unknown OS"; exit 1;;
+	esac
+}
 
 fw_reset() {
 	case `uname` in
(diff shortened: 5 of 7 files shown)


src/main.c
```diff
@@ -78,6 +78,7 @@ int cf_sbuf_len;
 int cf_sbuf_loopcnt;
 int cf_so_reuseport;
 int cf_tcp_socket_buffer;
+int cf_tcp_user_timeout;
```
sdemontfort: Not sure if this should have a default here?

src/main.c
```diff
@@ -272,6 +273,7 @@ CF_ABS("tcp_keepalive", CF_INT, cf_tcp_keepalive, 0, "1"),
 CF_ABS("tcp_keepcnt", CF_INT, cf_tcp_keepcnt, 0, "0"),
 CF_ABS("tcp_keepidle", CF_INT, cf_tcp_keepidle, 0, "0"),
 CF_ABS("tcp_keepintvl", CF_INT, cf_tcp_keepintvl, 0, "0"),
+CF_ABS("tcp_user_timeout", CF_INT, cf_tcp_user_timeout, 0, "0"),
```
sdemontfort: Not sure if this should have a default here?

src/util.c
```diff
@@ -24,6 +24,12 @@
 
 #include <usual/crypto/md5.h>
 #include <usual/crypto/csrandom.h>
+#include <sys/socket.h>
```
sdemontfort: Not sure if this is needed for usage of `SOL_TCP`?

petere:
The Linux build is failing because

    etc/optscan.sh
    tcp_user_timeout is missing in doc/config.md
    tcp_user_timeout is missing in etc/pgbouncer.ini

The Windows build is failing because there is no `<sys/socket.h>`. That's already included via `usual/socket.h` if present, so you don't have to do it in your patch.

src/main.c
petere: I think the logic you have is OK: By default, it's 0 and you leave the kernel setting alone. Only when you set it to non-zero it will get activated.

sdemontfort:
Amazing, I'll fix that now. Cheers @petere

src/main.c
sdemontfort: Sounds good to me 👍

src/util.c
```diff
@@ -24,6 +24,12 @@
...
+#include <netinet/tcp.h>
+
+#ifndef TCP_USER_TIMEOUT
+	#define TCP_USER_TIMEOUT 18 // how long for loss retry before timeout [ms]
```
sdemontfort: Not sure if this definition is needed here either.

sdemontfort:
I hadn't looked into this before, but it seems Windows does not support `TCP_USER_TIMEOUT`. Should we add an `is_unix` check to the addition of the setting in `util.c` @petere ?

petere:
> I hadn't looked into this before, but it seems Windows does not support `TCP_USER_TIMEOUT`. Should we add an `is_unix` check to the addition of the setting in `util.c` @petere ?

You should do `#ifdef TCP_USER_TIMEOUT` because not all Unixes support it either.

sdemontfort:
@petere I've fixed the failing build and test. Let me know if there's anything you see could be improved.

etc/pgbouncer.ini
```diff
@@ -304,6 +304,10 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; In-kernel buffer size (Linux default: 4096)
 ;tcp_socket_buffer = 0
 
+;; Close TCP connections that are falsely stuck in "ESTABLISHED" state
+;; after this much time in milliseconds
+;;tcp_user_timeout = 12500
```
petere: This comment should reflect the default value.

petere:
> @petere I've fixed the failing build and test by using an `#ifdef` for usage of the setting. Let me know if there's anything you see could be improved.

Adding a test in `test/test.sh` would be fantastic.

sdemontfort:
Good call. I'll add a test and fix the default value in `etc/pgbouncer.ini`.

sdemontfort:
Heads up; I'm working on setting up my environment so that I can write the test, it's just taking some time to get tests working.

sdemontfort:
@petere I've been trying for the past few days to get my environment set up for tests, but am currently hitting this error:
```
...:/pgbouncer# make check
...
sed: can't read pgdata/postgresql.conf: No such file or directory
Creating databases
psql: could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 6666?
could not connect to server: Cannot assign requested address
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 6666?
psql: could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 6666?
could not connect to server: Cannot assign requested address
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 6666?
```

Are there pre-requisites I've missed for running tests?

sdemontfort:
@petere I've added the test: `test_tcp_user_timeout`. Please let me know is that's not sufficient for the changes. Cheers.

sdemontfort:
Great, thanks for the feedback. Will do 👍

sdemontfort:
@petere I've made the test conditional to Linux and squashed the commits. Please let me know what you think.

doc/config.md
```diff
@@ -787,6 +787,12 @@ Default: 45 on Linux, otherwise 0
 
 Default: not set
 
+### tcp_user_timeout
+
+Enables socket-level timeout in milliseconds. This solves the situation where a database server on the end of a connection fails to kill a TCP connectionn properly after a shutdown or crash. TCP connections then get stuck and need clearing out. The Linux kernel setting `tcp_retries2` default value means stuck TCP connections get cleaned up after about 15 minutes. Setting `tcp_user_timeout` to a lower value means the connections get cleaned up more quickly if they're stuck.
```
KevinGreen:
```suggestion
Enables socket-level timeout in milliseconds. This solves the situation where a database server on the end of a connection fails to kill a TCP connection properly after a shutdown or crash. TCP connections then get stuck and need clearing out. The Linux kernel setting `tcp_retries2` default value means stuck TCP connections get cleaned up after about 15 minutes. Setting `tcp_user_timeout` to a lower value means the connections get cleaned up more quickly if they're stuck.
```

philosoralphter:
Excited to see this included!

doc/config.md
sdemontfort: Thanks for fixing my typo Kevin! 😅

sdemontfort: Thanks for fixing my typo Kevin! 😅

sdemontfort:
Me too @philosoralphter! We've been using this change (via a fork) in production for some time now and it's been working great.

src/util.c
```diff
@@ -141,6 +141,17 @@ bool tune_socket(int sock, bool is_unix)
 	if (!ok)
 		goto fail;
 
+	/*
+	 * TCP user timeout
+	 */
+#ifdef TCP_USER_TIMEOUT
+	val = cf_tcp_user_timeout;
+	errpos = "setsockopt/TCP_USER_TIMEOUT";
+	res = setsockopt(sock, SOL_TCP, TCP_USER_TIMEOUT, &val, sizeof(val));
```
petere: This should use `IPPROTO_TCP` instead of `SOL_TCP`.

test/test.sh
```diff
@@ -323,6 +334,34 @@ test_server_login_retry() {
 	return $rc
 }
 
+# tcp_user_timeout
+test_tcp_user_timeout() {
+	case `uname` in
+	Linux)
+		# timeout at ~7s
+		admin "set tcp_user_timeout=2000"
```
petere: If I comment out this line, the test does not fail.

sdemontfort: 😱 of course! On it.

philosoralphter:
@sdemontfort I wanted to check in on your progress here.  My team is not anxious to build our own bouncers ha ha

sdemontfort:
I lost track of this unfortunately. It looks like it made it into master! 🎉 thanks so much @petere for following up the broken test 😃

Issue #90, opened 2015-11-12
pgbouncer crashes if connection is authenticated using “auth_query”, and is then added to the “auth_file”


gclough:
If you’re authenticating via the “auth_query” parameter and then you add the same user into the “auth_file”, then the next time you try to connect pgbouncer crashes.  It appears that it can’t handle the fact that the same user now exists twice, and the `put_in_order()` function triggers the crash:

```
epoll_wait(4, {}, 32, 334)              = 0
stat("/etc/pgbouncer/userlist.txt", {st_mode=S_IFREG|0640, st_size=18, ...}) = 0
epoll_wait(4, {}, 32, 334)              = 0
stat("/etc/pgbouncer/userlist.txt", {st_mode=S_IFREG|0640, st_size=18, ...}) = 0
epoll_wait(4, {}, 32, 334)              = 0
stat("/etc/pgbouncer/userlist.txt", {st_mode=S_IFREG|0640, st_size=18, ...}) = 0
epoll_wait(4, {}, 32, 334)              = 0
stat("/etc/pgbouncer/userlist.txt", {st_mode=S_IFREG|0640, st_size=18, ...}) = 0
epoll_wait(4, {}, 32, 334)              = 0
stat("/etc/pgbouncer/userlist.txt", {st_mode=S_IFREG|0640, st_size=32, ...}) = 0
stat("/etc/pgbouncer/userlist.txt", {st_mode=S_IFREG|0640, st_size=32, ...}) = 0
open("/etc/pgbouncer/userlist.txt", O_RDONLY) = 14
fstat(14, {st_mode=S_IFREG|0640, st_size=32, ...}) = 0
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7ff72bd5f000
read(14, "\"postgres\" \"blah\"\n\"greg\" \"blah\"\n", 4096) = 32
close(14)                               = 0
munmap(0x7ff72bd5f000, 4096)            = 0
write(2, "2015-11-12 18:23:06.169 2182 DEB"..., 842015-11-12 18:23:06.169 2182 DEBUG loading auth_file: "/etc/pgbouncer/userlist.txt"
) = 84
write(3, "2015-11-12 18:23:06.169 2182 DEB"..., 84) = 84
epoll_wait(4, {}, 32, 333)              = 0
[...]

PR #429, opened 2019-10-22
Fix issue #90: pgbouncer crashes if auth_query user is added to auth_file

qris:
When a user is auto-added to pgBouncer using `auth_query` (by `add_db_user`), their pools are also added to the global pool_list.  If the user is then added to `auth_file` (e.g. because of #59), pgBouncer creates a new user object with the same name (in `add_user`).

Then, when it tries to find a pool for this user, and doesn't (because they are newly created, so their own pool list is empty), it calls new_pool to create one, which tries to add it to the global pool list
again (using `put_in_order`), which aborts (kills the process) because the same username (different PgUser object) and DB name are already in the pool.

This patch appears to fix the crash, but is only lightly tested, and I'm not sure whether we are supposed to remove the original user from the db list, and/or update the global pool entry to point to the new PgUser object, or whether it's OK to return a db pool entry from get_pool.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -534,15 +534,28 @@ static PgPool *new_pool(PgDatabase *db, PgUser *user)
 PgPool *get_pool(PgDatabase *db, PgUser *user)
 {
 	struct List *item;
-	PgPool *pool;
+	PgPool *pool_list_item;
 
 	if (!db || !user)
 		return NULL;
 
 	list_for_each(item, &user->pool_list) {
-		pool = container_of(item, PgPool, map_head);
-		if (pool->db == db)
-			return pool;
+		pool_list_item = container_of(item, PgPool, map_head);
+		if (pool_list_item->db == db)
+			return pool_list_item;
+	}
+
+	// We may already have a pool for this database name and user name, with a different user
+	// object, for example because we have users with the same name from auth_file, PAM and/or
+	// auth_query, so search the global list too:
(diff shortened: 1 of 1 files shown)


erthalion:
I'm a bit surprised by the lack of commentaries about this PR. Of course the described issue is probably not the biggest one, but definitely quite annoying.

About the PR itself, maybe it would be better to adjust this even earlier? If I understand correctly, the problem originates from `forced_user` being created without checking, if a user with the same `db`, `name` already exists. Then if we will do something like this:

```diff
 PgUser *force_user(PgDatabase *db, const char *name, const char *passwd)
 {
        PgUser *user = db->forced_user;

+       if (!user) {
+               user = find_user(name);
+       }
+
        if (!user) {
                user = slab_alloc(user_cache);
                if (!user)
```

we will get the `forced_user` from an existing one with correct `pool_list`.

PierreDucroquet:
I think this issue is also fixed in https://github.com/pgbouncer/pgbouncer/pull/393 along with other crashes, but it was never merged.

jhg03a:
Any chance this could be merged?  This is a much smaller fix than all the pieces of #393 and addresses a key issue that's been broken for years now.

JelteF:
I think this should be fixed now. At least issue #90 has been closed for a long time, and #1052 should have fixed any other similar issues.

Issue #374, opened 2019-03-20
Pgbouncer artificially limiting identifier names to 63 characters


lremes:
We are using vault database secret backend with Azure Kubernetes Service and Azure DB for postgresql as described at https://learn.hashicorp.com/vault/identity-access-management/vault-agent-k8s

The setup we have is working, and we are able to generate working credentials with "vault read database/creds/[dbname]" for kubernetes pods.

The issue we are facing is that we would like use pgbouncer sidecar (https://hub.docker.com/_/microsoft-azure-oss-db-tools-pgbouncer-sidecar)
to manage the connections, but the psql role name generated by vault is too long for pgbouncer. pgbouncer limits the username to 63 characters, but Azure DB for PostgreSQL requires that @hostname is included in the username.

So the vault generated psql role name is:
v-kubernet-dbname-xxxxxxxxxxxxxxxxxxxx-xxxxxxxxxx
or 
v-kubernet-[dbname]-[20 chars]-[10 chars]

As Azure postgresql expects @hostname after the username:
v-kubernet-dbname-xxxxxxxxxxxxxxxxxxxx-xxxxxxxxxx@db_hostname

So this leaves 19 characters for the dbname and hostname parts.

The obvious solution is to shorten the database name and hostname, but that really is not very sustainable solution.

If Postgres accepts names of a certain length and pgbouncer places an artificial limit on the names, it's not properly acting as a transparent bouncer.
So the artificial limits should be removed, or at least a flag/config option should be provided to bypass the restriction.

PJMODOS:
The rolname (used for storing user names) in PostgreSQL is limited to 63 characters, if whatever is running on Azure allows more then it's modified version of PostgreSQL, we don't promise to support those. We also don't claim to be fully transparent.

The maximum length of username in pgbouncer affects the memory usage so I am not too keen on increasing it arbitrarily for a fork of PostgreSQL without knowing more about how it works and why it needs such behavior. It's possible to change it in source (`MAX_USERNAME`) and make binaries with that however. This would make it work in your situation (such change should not be problem to integrate into the docker image you mentioned).

lremes:
Azure Database for Posgresql service uses the @db_hostname part of the username to route traffic to servers. Thus, it is a mandatory when using Azure Database for Postgresql, and this limitation will lead to random and unavoidable errors using the Azure Database for Postgresql service, since the length of rolname@db_database may not always be predictable.

petere:
[...]

PR #430, opened 2019-10-25
Increase maximum password length from 160 to 1024

lunaris:
While a maximum length of 160 bytes is acceptable for most applications,
temporary credentials generated by some cloud providers (e.g. AWS when
using IAM roles to access RDS databases) greatly exceed this size,
hitting around 800--900 bytes. This commit extends the maximum password
length and related static buffer sizes to accommodate these use cases.

More context:

When using IAM roles to access AWS's RDS PostgreSQL service, one uses the AWS API to generate a temporary credential which is used to log in to the PostgreSQL server. As best as I can tell, this temporary credential is effectively an HTTP request passed as a cleartext password to PostgreSQL, which forwards it on to an AWS authentication endpoint and uses the response to decide whether or not to authenticate the request. The token is roughly of the form:

```
XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX.YYYYYYYYYYYY.ZZZZZZZZZZZZZZ.rds.amazonaws.com:5432/?Action=connect&DBUser=WWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWWW&X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Expires=900&X-Amz-Date=DDDDDDDDDDDDDDDD&X-Amz-SignedHeaders=host&X-Amz-Security-Token=TTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTTT&X-Amz-Credential=CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCZZZZZZZZZZZZZZCCCrds-dbCCCCCCCCCCCCCCC&X-Amz-Signature=SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
```

Where:

* `XXXX...` is the name of the database, between 0--63 characters;
* `YYYY...` is a 12-character string relating to the account the database is hosted in;
* `ZZZZ...` is the AWS region, where the longest currently in service appear to be in the Asia Pacific regions e.g. `ap-northwest-1`, with a length of 14 characters;
* `WWWW...` is the database username, between 0--63 characters;
* `DDDD...` is a timestamp associated with the token, with a length of 16 characters;
* `TTTT...` and `CCCC...` are generated credentials and AWS access IDs, with fixed lengths;
* `SSSS...` is a signature with fixed length.

Adding these up comes to 803 bytes, but adding some leeway and picking a power of two I've opted to increase to 1024 bytes. Moreover, when allocated a buffer to send passwords across the wire, I've increased that from 512 to 2048 bytes to accommodate. Documentation has been updated too (hopefully in all the relevant places).

Is this a PR that `pgbouncer` could accept? I appreciate that it increases the static footprint of the program at some points but I think this is a common and powerful use case worth supporting. Thanks for the consideration!


willthames:
See also #362 (I'd be happy for any of these to be accepted)

hipyhop:
@lunaris How are you using RDS IAM auth with pgbouncer? Are you passing this through from the client somehow, or embedding this in the userlist.txt?
I'm trying to make this work as a proxy between users and RDS, but can't understand how to pass the password through from the client.

lunaris:
@hipyhop The set-up we're trying involves deploying a collection of containers in a single Kubernetes pod; something like:

* The service that wants access to PostgreSQL
* `pgbouncer`
* The credential manager (more on this shortly)

`pgbouncer` and the credential manager share a volume.

"The credential manager" is effectively just a container running a small Linux image with the AWS CLI installed. It basically runs a Bash script that every 10 minutes (IAM-role-generated tokens last up to 15 minutes) generates fresh credentials and writes a `pgbouncer`-compatible user list to the shared volume:

```bash
echo "\"my_iam_user\"" "\"$(aws rds generate-db-auth-token --hostname <HOST> --port <PORT> --username my_iam_user --region <REGION>)\"" > /shared-volume/userlist.txt
sleep ...
```

The `<HOST>`, `<REGION>` etc. could be pulled from the environment, which in turn could be set in the Kubernetes manifest.

`pgbouncer` mounts the `userlist.txt` as its user list in its configuration:

```
[databases]
db = <HOST>
[pgbouncer]
...
auth_file = /shared-volume/userlist.txt
auth_type = trust
```

(Depending on your RDS settings you might want `server_tls_sslmode = require` in your configuration too)

`pgbouncer` will spot when `userlist.txt` has changed and reload it, so now you have credentials being generated regularly and automatically used for new connections. Now, to tie it all up:

* Give the pod a role that allows it RDS IAM access to `my_iam_user` (make sure you `grant rds_iam to my_iam_user` in PG itself)
* Have `pgbouncer` only listen on loopback (`127.0.0.1`, since containers in a pod see each other on the loopback) on some port, `PG_BOUNCER_PORT` (5432 will do of course).
* Have the service connect to `localhost:PG_BOUNCER_PORT` with `my_iam_user` and no/any password -- `pgbouncer` will sort the rest.

As I say, we're still experimenting but I think this would work. It also frees your application logic from having to manage secrets/PostgreSQL credentials. Hope this helps.

Jan-M:
We do not need this for bouncer yet, but we face similar limits due to "generated" passwords within Postgres' related library defaults.

So giving virtual support on this issue to extend length in excess of 160 chars.

hakamadare:
@markokr @PJMODOS do you foresee being able to merge this patch?  right now we are maintaining our own package repository in order to distribute a patched version, we'd of course prefer to avoid that :)

petere:
As discussed in #374, we should change the static arrays to malloced arrays instead of making them bigger and bigger. It's unlikely that 2048 will be enough for long.

Also, PostgreSQL client tools don't support such long passwords consistently. That is being fixed in PostgreSQL 14.

Issue #426, opened 2019-10-07
The 'min_pool_size' setting is not always respected.


barqshasbite:
This could be considered a bug or "by design", depending on how you look at it.

The behavior of 'min_pool_size' is not entirely clear. Or perhaps the documentation is not clear.

My expectation was that it would maintain the 'min_pool_size' at all times. However this does not appear to be the case.

The documentation currently states:

> Add more server connections to pool if below this number. Improves behavior when usual load comes suddenly back after period of total inactivity.

However, it appears (from testing and looking at the code) that 'min_pool_size' is only used and respected when the number of clients connected to the pool is greater than 0:
```
    /* launch extra connections to satisfy min_pool_size */
    if (cur < cf_min_pool_size &&
        cur < pool->db->pool_size &&
        cf_pause_mode == P_NONE &&
        cf_reboot == 0 &&
        pool_client_count(pool) > 0)
    {
        log_debug("launching new connection to satisfy min_pool_size");
        launch_new_connection(pool);
    }
```

When the 'min_pool_size' setting was introduced, the documentation more accurately (but still not quite right) stated:

> Minimal pool size.  PgBouncer will open this many connections to backend database **after first client connect**.  Optional, not recommended in most normal setups.

It at least indicated the need for a client to connect before the 'min_pool_size' setting was used.

Due to this behavior, it is possible for a pool to drop to zero connections over a prolonged period (due to 'server_lifetime') and then cause a sudden spike in connections when a client connects. This can cause performance problems for both the clients and the database server, which is the opposite of what 'min_pool_size' is supposed to do.

It would be better if the restriction of having at least one client connected be optional (or removed if it is a bug). Adding a new setting that defaults to the current behavior may be the safest approach.

andrp92:
Heeey this issue is very serious, is anybody paying attetion to it?

PJMODOS:
Yes it's by design, I can imagine having option to change the behavior, but nobody is working on adding that at the moment AFAIK.

PR #437, opened 2019-11-20
Add 'min_pool_size_requires_clients' setting.

barqshasbite:
The default behaviour of PGBouncer requires that clients be connected
to a pool before it is initialized. There is no way to toggle this
behaviour to ensure that a pool is initialized immediately by PGBouncer.

This means that 'min_pool_size' will only be respected if clients are
connected. If all clients disconnect, then the pool can drop below the
'min_pool_size' setting to 0 ('server_lifetime' will eventually close
all connections). This can cause performance problems when a client
finally reconnects since all of the pool connections have to be
recreated.

By adding an explicit 'min_pool_size_requires_clients' setting that
defaults to 1 (original PGBouncer behaviour) users can decide on what
behaviour they prefer. If set to 0, PGBouncer will always maintain
the 'min_pool_size'.

Issue #426

### DIFF
--- a/src/janitor.c
+++ b/src/janitor.c
@@ -485,7 +485,8 @@ static void check_pool_size(PgPool *pool)
 	    cur < pool->db->pool_size &&
 	    cf_pause_mode == P_NONE &&
 	    cf_reboot == 0 &&
-	    pool_client_count(pool) > 0)
+		(cf_min_pool_size_requires_clients == 0 || (cf_min_pool_size_requires_clients == 1
+				&& pool_client_count(pool) > 0)))
 	{
 		log_debug("launching new connection to satisfy min_pool_size");
 		launch_new_connection(pool);
@@ -743,5 +744,10 @@ void config_postprocess(void)
 			db->pool_size = cf_default_pool_size;
 		if (db->res_pool_size < 0)
 			db->res_pool_size = cf_res_pool_size;
+		if (db->forced_user != NULL) {
+			// Initialize pool object for each database that has a forced user.
+			// Only creates data structures. Does not create connections.
+			get_pool(db, db->forced_user);
+		}
 	}
 }
--- a/doc/config.md
+++ b/doc/config.md
@@ -202,6 +202,14 @@ of total inactivity.
 
 Default: 0 (disabled)
 
+### min_pool_size_requires_clients
+
+Only maintain `min_pool_size` server connections while there are clients connected to the pool. 
+
+If `min_pool_size` is disabled, this setting has no effect.
+
+Default: 1 (enabled)
+
 ### reserve_pool_size
 
 How many additional connections to allow to a pool. 0 disables.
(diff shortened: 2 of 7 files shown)


markesha:
Hi @petere 
Can I ask you to review this one? Is this something you're considering adding?
Thank you in advance.

petere:
It's an interesting feature. I will look at it.

markesha:
Hi @petere 
Have you had a chance to look at the feature?

markesha:
Also, let me share the reason for the interest in this feature with the hope that it's something you can comment on:
I use pgbouncer with server_tls configuration, and I noticed that pgbouncer will always start even if tls files are not presented, resulting in its faulty state. The state is only discovered when a client connects to pgbouncer, as it's at this stage when pgbouncer initiates a connection to the db. The goal then is to detect and avoid such false positive healthy pgbouncer states, and in the absence of any health checking instruments available - I tried things from [pgbouncer-healthcheck](https://github.com/deliveroo/pgbouncer-healthcheck/)  to configuring `server_check_query` - I thought that having pgbouncer to always maintain an open connection - with `min_pool_size` - would be a nice workaround as pgbouncer would just fail trying to initiate one when its configuration is not correct.

If you happen to know a solution here, it would be greatly appreciated! Thank you in advance.

rchiossi:
@petere Any plans to move forward with this PR?

In my current project we have the need to validate on startup that a connection pool can be created (For early failure detection) and to maintain a minimum pool size even in the abscense of clients.

src/janitor.c
```diff
@@ -485,7 +485,8 @@ static void check_pool_size(PgPool *pool)
 	    cur < pool->db->pool_size &&
 	    cf_pause_mode == P_NONE &&
 	    cf_reboot == 0 &&
-	    pool_client_count(pool) > 0)
+		(cf_min_pool_size_requires_clients == 0 || (cf_min_pool_size_requires_clients == 1
+				&& pool_client_count(pool) > 0)))
```
JelteF, line 489: This seems a simpler way of doing this, since we only support 0 and 1 for `min_pool_size_requires_clients`
```suggestion
		(cf_min_pool_size_requires_clients == 0 || pool_client_count(pool) > 0))
```

src/janitor.c
```diff
@@ -743,5 +744,10 @@ void config_postprocess(void)
 			db->pool_size = cf_default_pool_size;
 		if (db->res_pool_size < 0)
 			db->res_pool_size = cf_res_pool_size;
+		if (db->forced_user != NULL) {
+			// Initialize pool object for each database that has a forced user.
+			// Only creates data structures. Does not create connections.
```
JelteF, line 749: Does this mean that `min_pool_size_requires_connections` only works for forced users?

JelteF (changes requested):
This needs a rebase too

src/janitor.c
fschmager: 👋 Would love to resurrect this PR and get it over the finish line.

I'm unclear about this change. Does this _need_ to happen for `db`s with forced users? Does this need to happen for  all `db`s? Could I get some guidance, please?

JelteF:
Closing this in favor of #947

Issue #434, opened 2019-11-11
SSL broken on 32 bit (Debian i386 and other architectures)


df7cb:
On 32-bit Debian, SSL support seems broken. Exercising the setup from test/ssl/test.sh:
```
test/ssl $ ../../pgbouncer -d test.ini
test/ssl $ PGSSLMODE=require PGPASSWORD=zzz psql -h 127.0.0.1 -p 6667 p0 -U bouncer
psql: Fehler: konnte nicht mit Server verbinden: Server unterstützt kein SSL, aber SSL wurde verlangt
```
SSL support is present, though:
```
test/ssl $ ldd ../../pgbouncer 
	linux-gate.so.1 (0xf7f32000)
	libcares.so.2 => /usr/lib/i386-linux-gnu/libcares.so.2 (0xf7e9f000)
	libevent-2.1.so.6 => /usr/lib/i386-linux-gnu/libevent-2.1.so.6 (0xf7e3d000)
	libssl.so.1.1 => /usr/lib/i386-linux-gnu/libssl.so.1.1 (0xf7da6000)
	libcrypto.so.1.1 => /usr/lib/i386-linux-gnu/libcrypto.so.1.1 (0xf7ae4000)
	libpthread.so.0 => /lib/i386-linux-gnu/libpthread.so.0 (0xf7ac3000)
	libpam.so.0 => /lib/i386-linux-gnu/libpam.so.0 (0xf7ab0000)
	libc.so.6 => /lib/i386-linux-gnu/libc.so.6 (0xf78d5000)
	libdl.so.2 => /lib/i386-linux-gnu/libdl.so.2 (0xf78cf000)
	/lib/ld-linux.so.2 (0xf7f33000)
	libaudit.so.1 => /lib/i386-linux-gnu/libaudit.so.1 (0xf78a4000)
	libcap-ng.so.0 => /lib/i386-linux-gnu/libcap-ng.so.0 (0xf789c000)
```
```
checking for OpenSSL... found
checking for SSL_CTX_use_certificate_chain_mem... no
checking for SSL_CTX_load_verify_mem... no
checking for asn1_time_parse... no
checking for root CA certs... /etc/ssl/certs/ca-certificates.crt
checking whether to build debug binary... yes
checking whether to enable asserts... no
checking whether to fail on warnings... no
configure: creating ./config.status
config.status: creating config.mak
config.status: creating lib/usual/config.h
[...]

PR #438, opened 2019-11-26
test: shorten lifetime of certificates to avoid overflowing 32bit time_t

jcristau:
If the test certs expire later than 2038, the conversion to time_t
overflows on 32bit and we get spurious handshake failures.

Fixes: #434

### DIFF
--- a/test/ssl/newsite.sh
+++ b/test/ssl/newsite.sh
@@ -12,7 +12,7 @@ test -f "$1/ca.key" || {
   exit 1
 }
 
-days=10240
+days=1024
 
 . ./lib.sh
 


petere:
Does `test/ssl/newca.sh` not need to be patched in the same way?

jcristau:
As far as I can tell https://github.com/libusual/libusual/blob/56b3e73181c1682375481b607a35d813cac84382/usual/tls/tls_conninfo.c#L126 doesn't look at the CA cert, only the leaf.  Ideally that code would become Y2038-safe as well, but changing the CA validity isn't needed as such.

PR #450, opened 2020-01-03
Add missing states of server connections

qris:
See `admin_show_servers`.


petere:
Thanks! The list of possible client states was also wrong, so I fixed that too.

PR #455, opened 2020-01-13
applicationname_add_host options for set application name

DanteG41:
Add an address or hostname to the application name. This is useful when using a connection to a bouncer on a localhost or a scheme with several pgbouncer.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -280,7 +280,7 @@ Default: 0
 
 ### application_name_add_host
 
-Add the client hostname or host address and port to the application name setting set on connection start.
+Add pgbouncer server hostname or client host address and port to the application name setting set on connection start.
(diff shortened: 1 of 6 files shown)


petere:
Please explain in more detail, give examples, etc.

DanteG41:
I think the diagram will explain it better.

pgbouncer patch explain

msirovy:
This could be very usefull

JelteF:
applicationname_add_host originally had the purpose of still being able to show the IP of the client in postgres logs when PgBouncer and the client are on different nodes. The usage you're trying to add here, isn't that. Because the client and PgBouncer in your diagram would have the same IP, so looking at IPs in postgres logs will show the correct IP. So it's not completely clear what benefit this has over that.

JelteF:
Closing this, since there was no response for a long time.

PR #481, opened 2020-04-23
Add query_wait_timeout to user settings

levkk:
### Change
Add `query_wait_timeout` to user settings.

### Why
Given two (2) or more read replicas, we want to force the client to failover to a different replica if the replica it's currently trying to use is unresponsive.

This is added because PostgreSQL protocol does not support read timeouts: the client will wait forever until the server answers or TCP keep-alive will break the connection.

Why not use the global `query_wait_timeout` setting? We have two different kind of users, foreground traffic user and background traffic user. The foreground traffic user needs to failover as quickly as is reasonable/possible. The background user can and should wait until replica recovers. The background user tends to run heavier queries as well, so we don't want to move all that expensive traffic to a different read replica: we are prioritizing the foreground traffic.

Why not implement `query_timeout` as well/instead? The assumption is if we can assign a query to a server, the server is healthy and will eventually return a result.

### Tests
Will write tests if proposed change is acceptable.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -298,6 +298,7 @@ struct PgUser {
 	int pool_mode;
 	int max_user_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
+	usec_t query_wait_timeout; /* how much a user is allowed to wait for a query to be assigned to a backend */
 };
 
 /*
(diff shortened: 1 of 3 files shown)


JelteF:
Adding more settings at the user (or database) level is always a good change. But this has bitrotted a bit too much. So closing this PR. Feel free to open a new PR with the same change.

PR #488, opened 2020-05-17
Add unhealthy database client debouncer

levkk:
### Problem
Database becomes unhealthy because of storage latency problems (or other problem). Clients connect, attempt to login but fail because the database can't process the login; Pgbouncer disconnects the client and closes the server connection because it's in an unclean state. Clients reconnect and the situation repeats. This leads to connection thrashing and prolongs the database recovery time by adding load to the database server.

### Solution
Debounce client logins for a period of time to allow the database time to recover from a bad state.

Two configs are added:
- `unhealthy_login_count`: how many failed logins before Pgbouncer considers the database unhealthy and disconnects all new clients; default 5 logins.
- `unhealthy_login_count_timeout`: how many seconds before Pgbouncer resets the above counter and allows new client logins; default 5 seconds.

### Why
We've experienced several failures while running our databases on a certain Cloud relational database provider and we noticed that, when running many thousands of clients and several thousand of requests per second, clients will thrash server connections if their login attempts fail. We believe this thrashing led to CPU spikes and prolonged database recovery time.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -250,6 +250,14 @@ pool.
 
 Default: 0 (unlimited)
 
+### unhealthy_login_count
+
+How many bad logins have to take place before a database is considered unhealthy.
+New logins will be disallowed for `unhealthy_login_count_timeout` seconds. When the
+timeout happens, the counter is reset back to 0.
+
+Default: 5
+
 ### server_round_robin
 
 By default, PgBouncer reuses server connections in LIFO (last-in, first-out) manner,
@@ -727,6 +735,13 @@ How many seconds to wait for buffer flush during SUSPEND or reboot (-R).
 A connection is dropped if the flush does not succeed.
 
 Default: 10
+
+### unhealthy_login_count_timeout
+
+How many seconds to wait before allowing new client logins after a database
+has been decided to be unhealthy.
+
+Default: 5
 
 
 ## Low-level network settings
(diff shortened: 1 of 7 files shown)


Melkij:
Hello

> attempt to login but fail because the database can't process the login

What is the difference with existing `server_login_retry`?

> server_login_retry
> If login failed, because of failure from connect() or authentication that pooler waits this much before retrying to connect.

levkk:
> What is the difference with existing server_login_retry?

You're right, the issue description is erroneously saying that logins failed and that was the problem. I think my root cause analysis was wrong: we did see multiple successful login attempts in the server logs. Therefore, logins did succeed, but clients disconnected all together, causing a new login storm when they reconnected.

I think what happened in our case is the server got slow and could not process queries in a reasonable amount of time. The clients have a local connection pool (we use ActiveRecord & Rails), and that connection pool effectively became saturated with connections that were not reused effectively. The clients then terminated all connections due to an application-set timeout (we use Rack::Timeout middleware), and new requests create a new connection storm.

The goal behind my patch is to prevent this connection storm. If a large amount of clients are disconnected in a state that requires disconnecting Pg server connections, the goal would be to throttle client logins until the presumable root cause, the slow database, recovers; this would prevent a connection storm on the database, helping it recover faster.

JelteF:
Healthy detection is definitely something that can be improved in PgBouncer. I'm closing this PR though as it's a bit unclear what case this is fixing, and also it (unsurprisingly) has a bunch of merge conflicts after all this time.

PR #489, opened 2020-05-27
Add support for IntervalStyle

gmr:
I need to be able to support specifying `IntervalStyle` in the same fashion as `DateStyle`, so here's a PR to do that :)

My testing was limited, but everything worked as expected when it was specified and when it was not.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -269,7 +269,7 @@ Default: 0
 ### ignore_startup_parameters
 
 By default, PgBouncer allows only parameters it can keep track of in startup
-packets: `client_encoding`, `datestyle`, `timezone` and `standard_conforming_strings`.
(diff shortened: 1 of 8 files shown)


davidfetter:
Can we get a rebase, @gmr ?

irminsul:
As a sidenote, Rails 6.1 [added](https://github.com/rails/rails/commit/e5a5cc483573f41fa396779057bd83ce389640d8#diff-2468c670eb10c24bd2823e42708489a336d6f21c6efc7e3c4a574166fa77bb22R798) `SET intervalstyle = iso_8601` to its connection setup.

So folks who are running Rails 6.1 together with pgbouncer should consider adding `connect_query='SET intervalstyle = iso_8601;'` to their pgbouncer config until this gets merged.

If you don't, Rails will silently convert non-ISO8601 values to `nil`.

PR #504, opened 2020-07-14
Fixed SO_REUSEPORT_LB on FreeBSD

dwagin:


petere:
Could you explain in more than zero words what you are trying to fix? ;-)

dwagin:
FreeBSD have both

> SO_REUSEPORT       enables duplicate address and port bindings
> SO_REUSEPORT_LB    enables duplicate address and port bindings with load balancing

 and original code in this situation was compiled with wrong option

PR #510, opened 2020-07-22
Add support for prepared statements to "transaction" mode

bmharper:
Before doing any more work on this PR, I'd like to just
sense-check if there's some glaring issue that I'm missing here.

We've been running this in beta production for a few months,
so.. "works on my machines"... but I'd love to get this merged,
if it can benefit others too.

-- COMMIT MESSAGE --

This commit adds support for an additional config setting
'prepared_statement_lock = 1', which enables some new locking
mechanisms designed to accomodate the use of prepared
statements, while pgbouncer is operating in
'pool_mode = transaction'.

Our use case for pgbouncer is a lot of mostly-idle
services, each having several DB connections permanently
open. The number of real server connections very quickly rises
to many hundreds. This commit allows use to use pgbouncer
in transaction mode, and by setting a low idle timeout
(eg 5 seconds), we can bring our number of real server
connections down substantially (eg 1000 to 50), which
greatly improves our multi-tenant efficiency.

The prepared_statement_lock feature allows us to write code
naturally, without having to jump through any weird hoops,
and basically work around Postgres' multi-process
architecture without having to change any of our code.

Postgres has two types of prepared statements - the 
"anonymous" statement, and named statements. There is one
implicit anonymous statement bound to every database connection,
but clients are free to create as many named statements as they
wish. Anonymous statements are used by many libraries, such
as Go's lib/pq, often without the programmer's knowledge.

Although it is possible to treat the anonymous statement as
a long-lived entity, in practice it seems like most Postgres
libraries use it for short-lived execution. For example,
if you execute the same statement multiple times from Go code,
then it is theoretically possible that the Go code could
reuse the previous anonymous statement for the 2nd and
subsequent executions. However, in practice, this is not the
case for Go's lib/pq, and I suspect it is also not the case
for a great deal of other libpq clients. What seems to happen
in practice, is that the libpq wrapper library will recreate
the anonymous statement every time the programmer needs to
use it, making the code stateless in this regard.

So if this is true - that client code treats the anonymous
statement as a once-off thing, we still need a way to correctly
detect when it's usage starts and ends. As far as I can tell
from my empirical tests here, it is sufficient to use the
server command 'C' (CommandComplete) to detect when an
anonymous statement is done being used.

Named prepared statements are much easier to track than
anonymous statements, because they come through with explicit
packets that we can decode (both the PREPARE and DEALLOCATE).
In this commit, I do the simplest thing possible, and just
lock the server connection when there is at least one named
statement in existence. There are no doubt cases where this
would cause all connections to be locked, but this doesn't
apply to any of the systems that we build. Also, I use
a simple list data structure to keep track of these, so
any clients that use a ton of prepared statements will hit
scaling issues. I'm guessing that this is rare, and these
clients can simply leave the new prepared statement
config setting switched off, and they won't have to pay
that penalty.

I have only added a small suite of tests from Go, because
this is one of our primary use cases. Internally, we also
have our own C++ libpq wrapper, to which I've added a set
of unit tests, but unfortunately I can't incorporate those
here. What I plan to do is to rewrite those as straight
libpq-based tests, and add them to the test/ps directory.

We have been running this branch of pgbouncer for the last
two months on some beta systems, and so far we are not
aware of any issues.

In order to figure out the feasibility of the methods here,
I added functions log_server_event_stream and
log_client_event_stream, which make it easy to see the
flow of events between client and server. This allowed me
to figure out appropriate signals that could be used to
detect the end of an anonymous statement's usage.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,7 @@ pgbouncer_SOURCES = \
 	src/pktbuf.c \
 	src/pooler.c \
 	src/proto.c \
+	src/ps.c \
 	src/sbuf.c \
 	src/scram.c \
 	src/server.c \
@@ -42,6 +43,7 @@ pgbouncer_SOURCES = \
 	include/pktbuf.h \
 	include/pooler.h \
 	include/proto.h \
+	include/ps.h \
 	include/sbuf.h \
 	include/scram.h \
 	include/server.h \
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,5 @@
+module pgbouncer
+
+go 1.13
+
+require github.com/lib/pq v1.3.0
--- a/go.sum
+++ b/go.sum
@@ -1,0 +1,2 @@
+github.com/lib/pq v1.3.0 h1:/qkRGz8zljWiDcFvgpwUpwIAPu3r07TDvs3Rws+o/pU=
+github.com/lib/pq v1.3.0/go.mod h1:5WUZQaWbwv1U+lTReE5YruASi9Al49XbQIvNi/34Woo=
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -123,6 +123,7 @@ extern int cf_sbuf_len;
 #include "janitor.h"
 #include "hba.h"
 #include "pam.h"
+#include "ps.h"
 
 #ifndef WIN32
 #define DEFAULT_UNIX_SOCKET_DIR "/tmp"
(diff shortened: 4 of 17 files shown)


cobolbaby:
@ALL Is there any progress on this request?

bmharper:
@cobolbaby do you need this functionality? I'm happy to clean up the PR so that the red tick boxes go away, and fix up the merge conflict too. But as you can see, I haven't received any feedback from the project maintainers on this.

cobolbaby:
@All @bmharper Yes, this problem is the main reason that hinders me from promoting pgbouncer. If there is no such problem, there will be fewer restrictions when the client is used, and the performance will be better.

cobolbaby:
#695

JelteF:
I'm closing this in favor of #845.

JelteF:
Thank you for the PR, but #845 seems better because it does not require locking server connections to certain clients.

PR #526, opened 2020-09-04
Don't log anything if client disconnects immediately

### LINKED ISSUE #323: Client unexpected eof (age=0)
When using Kubernetes (or other ways to maintain pgbouncer reliably service up and running), there is always the need to know if the pgbouncer server is still running. Kubernetes uses "`livenessProbe`" to check if the service is still responding to TCP requests. However, it only checks if it can open the TCP port and closes it; this results in a bunch of logged errors by pgbouncer, as such: `LOG C-0x55b321680410: (nodb)/(nouser)@10.56.0.1:34772 closing because: client unexpected eof (age=0)`

I suggest making it so when `(age=0)` it (perhaps optionally) ignores an open/close connection so this doesn't happen.

Maybe we could add an option for this, or maybe we could simply ignore open/close connections altogether.

petere:
This avoids log spam when monitoring systems just open a TCP/IP
connection but don't send anything before disconnecting.

closes #323

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -920,7 +920,14 @@ bool client_proto(SBuf *sbuf, SBufEvent evtype, struct MBuf *data)
 	case SBUF_EV_CONNECT_FAILED:
 		/* ^ those should not happen */
 	case SBUF_EV_RECV_FAILED:
-		disconnect_client(client, false, "client unexpected eof");
+		/*
+		 * Don't log error if client disconnects right away,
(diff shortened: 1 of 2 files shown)


ebarault:
@petere Great, could you please tag a new release ?

kaplik:
Would it be possible to release new version with this fix? Thanks

petere:
Yes, soon.

PR #538, opened 2020-10-15
Do not crash during mock authentication without auth_file

gkokolatos:
During mock authentication, pgbouncer should not emmit any connection errors.
Also it should avoid to create a pool. Before this patch, a pool would not be
created, however it would attempt to use one, leading to a crash.

Co-authored-by: Daniel Gustafsson <daniel@yesql.se>

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -238,9 +238,11 @@ static bool finish_set_pool(PgSocket *client, bool takeover)
 		ok = finish_client_login(client);
 		break;
 	case AUTH_TRUST:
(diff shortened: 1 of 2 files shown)


petere:
The test case contained in your patch did not expose any crash for me. It might be that c03275de8dc066e1d4f5b2c110540db79dad674c has already fixed the issue for you. Otherwise, please provide an improved test case.

gkokolatos:
Your commit seems to address the issue.

Thanks,
Georgios

Issue #245, opened 2017-11-08
Delayed cancel hits incorrect query.


skarsol:
We're experiencing an issue where typically non-interactive queries (duration of milliseconds) are getting canceled with the message "canceling statement due to user request." There is no physical way for a user to insert a cancel there. Examining packet traces seems to indicate that pgbouncer is somehow delaying cancel requests and then when it ends up sending it on to postgres, a different query is now running on that backend and it gets cancelled.

In one example we have a front end server (S) running pgbouncer 1.7.2 (PGB1) that talks to a postgres server PSQL that is also running pgbouncer 1.7.2 (PGB2). 

S initiates a query that makes it to PSQL. After running for 8 seconds, S sends a cancel. PGB1 correctly propagates that cancel to the correct PID on PGB2. Almost 1 second later the result set makes its way from PSQL to S. No cancel is sent from PGB2 to PSQL.

8 minutes pass. 

PGB2 sends that pending cancel request to PSQL targeting the correct pid. That pid has processed thousands of queries in the interim. PSQL ACKs the cancel request, then that pid receives a new query, processes the query, returns the result set, receives yet another new query, and then cancels it out. The time frame between the cancelled query being sent to PSQL and the cancel response being generated is .000156 seconds.

![cancel](https://user-images.githubusercontent.com/1915152/32578433-d5d4a71c-c4a2-11e7-9d25-f59d5afbb06b.jpg)

petere:
I haven't analyzed this in much detail, but my gut feeling is that that is just the way it works. The cancel request requires a second connection, and if that connection is being queued up in pgbouncer, then it could conceivably arrive much later when something else is already running. Cancellation works on the session, not a specific query.

Why does your client send so many cancel requests? Is that a part of your architecture?

skarsol:
We have a fairly flexible interface that allows users to create varied queries across their data. If they manage to create a query that we haven't anticipated (or if they're just impatient) that runs slow then they'll hit cancel and make corrections to it (shrink date range, add other constraints, etc).

I understand the mechanics of how Postgres handles cancellations, but there should never be occasion for queries to queue more than a few seconds; our pool sizes are fairly generous. We're seeing pgbouncer sit on the cancel for multiple minutes before sending it. In the instant case above, pgbouncer sat on it for over 8 minutes. If that DB had been queuing for 8 minutes, alarms would have been going off. :)

[...]

PR #542, opened 2020-10-22
launch new connections when cancel requests are queued

thanodnl:

In this patch we introduce the launching of a new connection when cancelations are queued. There already was a loop that would calculate if a new connection should be launched if there was a client waiting for a backend. By not launching connections for cancelations we have observed very chaotic behaviour where a bunch (1200) of cancelations are flushed at once when a new connection was established. These cancelations came to pgbouncer at a moment when no new connections could be launched as other wise they would have been consumed from the queue directly after the connection gets established.

cancelation queueing and connection launch: https://github.com/pgbouncer/pgbouncer/blob/14331ba97f99ed8063363f1a40e6b00634e14428/src/objects.c#L1351-L1354

connection assigned to cancelation:
https://github.com/pgbouncer/pgbouncer/blob/ad6b5af4e27c8e9059b7a1e7dd541ea3b7c72624/src/server.c#L407-L414

Given that cancelations are handled first I propose we launch new connections directly when we have a non-empty cancelation list. There is a back pressure mechanism in `launch_new_connection` that will prevent multiple connections from being started at the same time: https://github.com/pgbouncer/pgbouncer/blob/14331ba97f99ed8063363f1a40e6b00634e14428/src/objects.c#L1139-L1143

### DIFF
--- a/src/janitor.c
+++ b/src/janitor.c
@@ -174,6 +174,12 @@ static void per_loop_activate(PgPool *pool)
 	PgSocket *client;
 	int sv_tested, sv_used;
 
+	/* if there is a cancelation request waiting, open a new socket */
+	if (!statlist_empty(&pool->cancel_req_list)) {
+		launch_new_connection(pool);
+		return;
+	}
+
 	/* see if any server have been freed */
 	sv_tested = statlist_count(&pool->tested_server_list);
 	sv_used = statlist_count(&pool->used_server_list);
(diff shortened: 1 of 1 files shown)


src/janitor.c
```diff
@@ -174,6 +174,12 @@ static void per_loop_activate(PgPool *pool)
 	PgSocket *client;
 	int sv_tested, sv_used;
 
+	/* if there is a cancelation request waiting, open a new socket */
```
petere, line 177: Maybe change this comment to

    if there is a cancel request waiting, open a new connection

Looks good otherwise.

PR #544, opened 2020-10-22
introduce a generation to a server connection

thanodnl:

This patch is most experimental and would require more testing and help from a maintainer.

The goal of this patch is to add the notion of a generation to a `Server` socket. The generation should be incremented every time a command is dispatched to the server. By keeping this generation we could compare the generation (and thus the command sent) at the moment of a cancelation being received by pgbouncer to the generation at the moment of the forward of the cancelation. As described and observed in #245, there can be a delay between pgbouncer receiving the cancelation and the moment a cancelation is processed on a new socket.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -952,6 +952,7 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 	/* tag the server as dirty */
 	client->link->ready = false;
 	client->link->idle_tx = false;
+	client->link->generation++; /* TODO figure out what happens on overflow */
 
 	/* forward the packet */
 	sbuf_prepare_send(sbuf, &client->link->sbuf, pkt->len);
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -407,6 +407,7 @@ struct PgSocket {
 	usec_t wait_start;	/* waiting start moment */
 
 	uint8_t cancel_key[BACKENDKEY_LEN]; /* client: generated, server: remote */
+	uint64_t generation;	/* server: number of commands sent, client: number of commands sent to the server on cancelation request */
 	PgAddr remote_addr;	/* ip:port for remote endpoint */
 	PgAddr local_addr;	/* ip:port for local endpoint */
 
--- a/src/objects.c
+++ b/src/objects.c
@@ -1342,9 +1342,11 @@ found:
 	if (!sbuf_close(&req->sbuf))
 		log_noise("sbuf_close failed, retry later");
 
-	/* remember server key */
+	/* remember server, server key and generation */
 	server = main_client->link;
+	req->link = server;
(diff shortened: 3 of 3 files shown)


src/client.c
```diff
@@ -952,6 +952,7 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 	/* tag the server as dirty */
 	client->link->ready = false;
 	client->link->idle_tx = false;
+	client->link->generation++; /* TODO figure out what happens on overflow */
```
JelteF, line 955: I think it can reasonably be assumed that this never overflows, since it's a 64bit uint. Which means that even when incrementing this every nanosecond this would take 584 years to overflow. Any restart within these 584 years would reset this number to 0. 

Source: https://github.com/jbenet/nanotime#range

petere:
In a way, this would create a behavior in pgbouncer that is stronger than what plain PostgreSQL does. Normally, a cancel request just cancels what is currently running, and there is no guarantee that it cancels what was running when the request was sent.

I'm wondering whether with your other two patches applied, which would result in cancel requests always processed first, this would still be a significant problem.

thanodnl:
What you say here makes a lot of sense. We run with this patch in production currently, but I think it is mainly due to over active cancelations being sent by our app.

For reasons we have a place where we aggressively send a cancelation on a connection when we want to run a different query. This cancelation is sent without checking if anything is running on the connection. Subsequently we send the new statement for execution. With pgbouncer in the mix this cancelation often comes in to the backend when the new query is already dispatched.

This is due to a combination of factors:
 - us not checking the status of the connection and always sending the cancelation
 - network latency
 - ssl handshakes (although I am currently blank if the cancelation needs to wait till an ssl connection is established)

With this generation patch we have observed the cancelation coming in delayed, canceling the subsequent statement, has been completely eliminated.

The over active cancelations are not a problem when connecting directly to postgres as the chances of the new request beating the cancelation on the network are much lower. However if pgbouncer needs to establish a new connection before it can send the cancelation it at least takes 3 trips on the network (tcp handshake), after our application has delivered the cancelation message before it reaches the postgres backend.

Will experiment a bit more with our app only sending a cancelation when it is actually having a connection that is busy.

JelteF:
I took a closer look at this PR and the comments. I disagree that it would create stronger behavior than Postgres has. The problem isn't that it doesn't cancel what was running at the time of the cancel request. The problem is that due to a race condition it can cancel a query from a completely different connection.

However, I don't think that this PR solves this issue in the right way. Looking at the code it seems to be prone to segfaults: The server is added as `req->link`, but isn't set to NULL if the server is freed. I addressed that issue in #717 and by doing so the generation isn't actually needed anymore. Because now I simply infer that the connection was reused/disconnected when `req->link` is NULL.

PR #548, opened 2020-11-19
wait_time is already in microseconds

eulerto:
There is an oversight in the average math (cf 350a0b57). In
calc_average(), wait_time and dur are in microseconds. Hence, the
average does not need to be multiplied by USEC.

This issue was reported by Fabrizio Mello (@fabriziomello).


eulerto:
Nevermind, documentation says
> Time spent by clients waiting for a server, in **microseconds** (average per second).

PR #550, opened 2020-12-04
Support hot reloading of TLS settings

### LINKED ISSUE #380: feature request: change ssl parameters with reload
Hello there,
i wonder if it is possible and how hard to implement it would be to make ssl parameters changeable uppon SIGHUP (similiar as this feature in pg10 https://paquier.xyz/postgresql-2/postgres-10-ssl-reload/)?
This became real issue once pgbouncer is deployed with ssl certificates as there is no possible way to renew certificate without killing all connections.

Thank you.

JelteF:
This allows rotating TLS certificates without downtime. The reason this
change is so small is because OpenSSL already does reference counting of
the SSL context. All that's needed is load the values from the config
again and build a new SSL context that can be used for newly opened
connections. Any existing connections will continue using the old
context, which will get freed automatically when the reference count
hits zero.

Fixes #380

### DIFF
--- a/include/sbuf.h
+++ b/include/sbuf.h
@@ -42,6 +42,15 @@ typedef enum {
 
 struct tls;
 
+/*
+ * client_accept_sslmode is the currently applied sslmode that is used to
+ * accept client connections. This is usually the same as
+ * cf_client_tls_sslmode, except when changing the TLS configuration failed for
+ * some reason (e.g. cert file not found). In this exceptional case,
+ * cf_client_tls_sslmode will be the new sslmode, which is not actually
+ * applied. And client_accept_sslmode is the still applied previous version. So
+ * usually you should use this variable over cf_client_tls_sslmode.
+ */
 extern int client_accept_sslmode;
 
 /* fwd def */
--- a/include/sbuf.h
+++ b/include/sbuf.h
@@ -93,7 +93,7 @@ void sbuf_init(SBuf *sbuf, sbuf_cb_t proto_fn);
 bool sbuf_accept(SBuf *sbuf, int read_sock, bool is_unix)  _MUSTCHECK;
 bool sbuf_connect(SBuf *sbuf, const struct sockaddr *sa, socklen_t sa_len, time_t timeout_sec)  _MUSTCHECK;
 
-void sbuf_tls_setup(void);
+bool sbuf_tls_setup(void);
 bool sbuf_tls_accept(SBuf *sbuf)  _MUSTCHECK;
 bool sbuf_tls_connect(SBuf *sbuf, const char *hostname)  _MUSTCHECK;
 
--- a/include/sbuf.h
+++ b/include/sbuf.h
@@ -42,6 +42,8 @@ typedef enum {
 
 struct tls;
 
+int client_accept_sslmode;
+
 /* fwd def */
 typedef struct SBuf SBuf;
 typedef struct SBufIO SBufIO;
--- a/include/sbuf.h
+++ b/include/sbuf.h
@@ -42,7 +42,7 @@ typedef enum {
 
 struct tls;
 
-int client_accept_sslmode;
+extern int client_accept_sslmode;
 
 /* fwd def */
 typedef struct SBuf SBuf;
--- a/include/sbuf.h
+++ b/include/sbuf.h
@@ -42,17 +42,6 @@ typedef enum {
 
 struct tls;
 
-/*
- * client_accept_sslmode is the currently applied sslmode that is used to
- * accept client connections. This is usually the same as
- * cf_client_tls_sslmode, except when changing the TLS configuration failed for
- * some reason (e.g. cert file not found). In this exceptional case,
- * cf_client_tls_sslmode will be the new sslmode, which is not actually
- * applied. And client_accept_sslmode is the still applied previous version. So
- * usually you should use this variable over cf_client_tls_sslmode.
- */
-extern int client_accept_sslmode;
-
 /* fwd def */
 typedef struct SBuf SBuf;
 typedef struct SBufIO SBufIO;
@@ -103,6 +92,22 @@ struct SBuf {
 void sbuf_init(SBuf *sbuf, sbuf_cb_t proto_fn);
 bool sbuf_accept(SBuf *sbuf, int read_sock, bool is_unix)  _MUSTCHECK;
 bool sbuf_connect(SBuf *sbuf, const struct sockaddr *sa, socklen_t sa_len, time_t timeout_sec)  _MUSTCHECK;
+
+/*
+ * client_accept_sslmode is the currently applied sslmode that is used to
+ * accept client connections. This is usually the same as
+ * cf_client_tls_sslmode, except when changing the TLS configuration failed for
+ * some reason (e.g. cert file not found). In this exceptional case,
+ * cf_client_tls_sslmode will be the new sslmode, which is not actually
+ * applied. And client_accept_sslmode is the still applied previous version. So
+ * usually you should use this variable over cf_client_tls_sslmode.
+ */
+extern int client_accept_sslmode;
+/*
+ * Same as client_accept_sslmode, but for server connections.
+ */
+extern int server_connect_sslmode;
+
 
 bool sbuf_tls_setup(void);
 bool sbuf_tls_accept(SBuf *sbuf)  _MUSTCHECK;
--- a/src/admin.c
+++ b/src/admin.c
@@ -234,8 +234,13 @@ static bool admin_set(PgSocket *admin, const char *key, const char *val)
 	if (admin->admin_user) {
 		ok = set_config_param(key, val);
 		if (ok) {
+			PktBuf *buf = pktbuf_dynamic(256);
+			if (strstr(key, "_tls_") != NULL) {
+				if (!sbuf_tls_setup())
+					pktbuf_write_Notice(buf, "TLS settings could not be applied, still using old configuration");
+			}
 			snprintf(tmp, sizeof(tmp), "SET %s=%s", key, val);
-			return admin_ready(admin, tmp);
+			return admin_flush(admin, buf, tmp);
 		} else {
 			return admin_error(admin, "SET failed");
 		}
--- a/src/admin.c
+++ b/src/admin.c
@@ -976,6 +976,8 @@ static bool admin_cmd_reload(PgSocket *admin, const char *arg)
 
 	log_info("RELOAD command issued");
 	load_config();
+	if (!sbuf_tls_setup())
+		log_error("TLS configuration could not be reloaded, keeping old configuration");
 	return admin_ready(admin, "RELOAD");
 }
 
--- a/src/main.c
+++ b/src/main.c
@@ -277,12 +277,12 @@ CF_ABS("server_login_retry", CF_TIME_USEC, cf_server_login_retry, 0, "15"),
 CF_ABS("server_reset_query", CF_STR, cf_server_reset_query, 0, "DISCARD ALL"),
 CF_ABS("server_reset_query_always", CF_INT, cf_server_reset_query_always, 0, "0"),
 CF_ABS("server_round_robin", CF_INT, cf_server_round_robin, 0, "0"),
-CF_ABS("server_tls_ca_file", CF_STR, cf_server_tls_ca_file, CF_NO_RELOAD, ""),
-CF_ABS("server_tls_cert_file", CF_STR, cf_server_tls_cert_file, CF_NO_RELOAD, ""),
-CF_ABS("server_tls_ciphers", CF_STR, cf_server_tls_ciphers, CF_NO_RELOAD, "fast"),
-CF_ABS("server_tls_key_file", CF_STR, cf_server_tls_key_file, CF_NO_RELOAD, ""),
-CF_ABS("server_tls_protocols", CF_STR, cf_server_tls_protocols, CF_NO_RELOAD, "secure"),
-CF_ABS("server_tls_sslmode", CF_LOOKUP(sslmode_map), cf_server_tls_sslmode, CF_NO_RELOAD, "disable"),
+CF_ABS("server_tls_ca_file", CF_STR, cf_server_tls_ca_file, 0, ""),
+CF_ABS("server_tls_cert_file", CF_STR, cf_server_tls_cert_file, 0, ""),
+CF_ABS("server_tls_ciphers", CF_STR, cf_server_tls_ciphers, 0, "fast"),
+CF_ABS("server_tls_key_file", CF_STR, cf_server_tls_key_file, 0, ""),
+CF_ABS("server_tls_protocols", CF_STR, cf_server_tls_protocols, 0, "secure"),
+CF_ABS("server_tls_sslmode", CF_LOOKUP(sslmode_map), cf_server_tls_sslmode, 0, "disable"),
 #ifdef WIN32
 CF_ABS("service_name", CF_STR, cf_jobname, CF_NO_RELOAD, NULL), /* alias for job_name */
 #endif
@@ -498,7 +498,8 @@ static void handle_sighup(int sock, short flags, void *arg)
 	log_info("got SIGHUP, re-reading config");
 	sd_notify(0, "RELOADING=1");
 	load_config();
-	sbuf_tls_setup();
+	if (!sbuf_tls_setup())
+		log_error("TLS configuration could not be reloaded, keeping old configuration");
 	sd_notify(0, "READY=1");
 }
 #endif
@@ -921,9 +922,8 @@ int main(int argc, char *argv[])
 	init_caches();
 	logging_prefix_cb = log_socket_prefix;
 
-	if (!sbuf_tls_setup()) {
+	if (!sbuf_tls_setup())
 		die("TLS setup failed");
-	}
 
 	/* prefer cmdline over config for username */
 	if (arg_username) {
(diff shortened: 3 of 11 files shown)


JelteF:
The only thing that I realized that's not very user friendly is that if the new SSL settings cannot be applied because of a config error, then the pgbouncer process exits with a non zero exit status and an error message. This is because the TLS config parsing is not part the `load_config` function, but is separate. Making bad configs work would be some more code.

JelteF:
I updated the PR now to include support for reloading of configs that contain bad TLS settings. If the TLS settings cannot be applied successfully, then the old TLS settings continue to be used.

JelteF:
One thing to note is that this PR only allows hot reloading of `client_tls_xxx` settings, not `server_tls_xxx` settings. For the `server_tls_xxx` settings you probably want to close all the connections, because the new settings might impact the security of existing connections. For example, when changing from `server_tls_sslmode=disable` to `require`. Then you wouldn't want to keep existing connections that are unencrypted.

src/main.c
```diff
@@ -498,6 +498,7 @@ static void handle_sighup(int sock, short flags, void *arg)
 	log_info("got SIGHUP, re-reading config");
 	sd_notify(0, "RELOADING=1");
 	load_config();
+	sbuf_tls_setup();
```
petere, line 501: I think it would be good to check the exit status here and in case of failure log a message along the lines of "TLS configuration could not be reloaded, keeping old configuration".

src/main.c
```diff
@@ -912,7 +913,9 @@ int main(int argc, char *argv[])
 	init_caches();
 	logging_prefix_cb = log_socket_prefix;
 
-	sbuf_tls_setup();
+	if (!sbuf_tls_setup()) {
+		die("TLS setup failed");
+	}
```
petere, line 918: Braces can be omitted here.

src/sbuf.c
```diff
@@ -938,56 +951,100 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
 	}
+
+	return true;
 }
 
-void sbuf_tls_setup(void)
+bool sbuf_tls_setup(void)
 {
 	int err;
+	bool ok;
```
petere, line 961: This could be moved to a smaller scope.

src/sbuf.c
```diff
@@ -938,56 +951,100 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+	struct tls_config *new_client_accept_conf = NULL;
+	struct tls_config *new_server_connect_conf = NULL;
```
petere, line 963: You said you are only dealing with `client_tls_*` settings in this patch. Why is this code here concerning itself with the server side of things then?

src/sbuf.c
```diff
@@ -938,56 +951,100 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
 
 	if (cf_client_tls_sslmode != SSLMODE_DISABLED) {
-		client_accept_conf = tls_config_new();
-		if (!client_accept_conf)
-			die("tls_config_new failed 2");
-		setup_tls(client_accept_conf, "client_tls", cf_client_tls_sslmode,
+		new_client_accept_conf = tls_config_new();
+		if (!new_client_accept_conf) {
+			log_error("tls_config_new failed 2");
+			tls_config_free(new_client_accept_conf);
```
petere, line 1010: Let's collect all these free calls at the end of the function behind a "failed" label and just `goto failed` here. There are many examples of that approach in the pgbouncer code.

src/sbuf.c
```diff
@@ -938,56 +951,100 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+		if (err) {
+			log_error("TLS setup failed: %s", tls_error(new_client_accept_base));
+			tls_free(new_client_accept_base);
+			tls_config_free(new_client_accept_conf);
+			tls_config_free(new_server_connect_conf);
+			return false;
+		}
 	}
+
+	client_accept_conf = new_client_accept_conf;
```
petere, line 1044: Don't we need to free the previously assigned objects?

petere (changes requested):
This is a nice feature and looks to be on the right track. Needs a few tweaks.

If you are slightly more ambitious, write a test case, too. (see `test/ssl/`).

petere:
> One thing to note is that this PR only allows hot reloading of `client_tls_xxx` settings, not `server_tls_xxx` settings. For the `server_tls_xxx` settings you probably want to close all the connections, because the new settings might impact the security of existing connections. For example, when changing from `server_tls_sslmode=disable` to `require`. Then you wouldn't want to keep existing connections that are unencrypted.

What you could do is if the `server_tls_*` settings are changed, automatically tag all SSL-using databases or pools as "dirty" so that they are reconnected at the next opportunity. (See `tag_database_dirty()` and related functions.) This is also what happens if a database connection string was changed during a reload.

This could probably be done with a small amount of extra effort. But it's also fair to try to solve the client-side case first.

src/sbuf.c
JelteF: I used the `tag_pool_dirty` function now to also allow changing of `server_tls_*` settings.

src/sbuf.c
JelteF: good catch, done

src/sbuf.c
JelteF: removed the variable completely

JelteF:
I added some tests and found a few issues while writing them. 

I also noticed an unrelated bug when working on this. You can use `SET key=value` on the pgbouncer admin database to change settings that are marked as `CF_NO_RELOAD`. This seems unintended.

test/ssl/lib.sh
```diff
@@ -46,6 +46,6 @@ run_ca() {
     shift
   done
   if test "$1" = '-out'; then
-    cp "${CaName}/certs/$ser.pem" "$2"
+    cp "${CaName}/certs/$ser.pem" "$2" 2>/dev/null
```
JelteF, line 49: Without this I got errors the following error in the test output for subsequent runs of `test/ssl/test.sh`:
```
cp: cannot stat 'TestCA2/certs/02.pem': No such file or directory
```

The first run didn't have this. Even with the error, everything seems to work fine. So that's why I ignored it like this. If you think this should be addressed in a different way, could you let me know how you would like to see it addressed instead?

JelteF:
Thanks for the review, I addressed all of your comments. I pushed all the new changes in new commits to hopefully make the PR easy to review. Please let me know if you want me to squash them all together, or if you don't mind and let Github take care of that when doing the final merge. 

Also, obviously let me know if you want me to change something else. The PR has gotten a bit bigger with these changes, but it has definitely gotten better too. (e.g. you can also use `SET client_tls_xxx=something` and `RELOAD` now in addition to `SIGHUP`)

JelteF:
I ran into an issue in `safe_accept` when using high verbosity, see commit message of this commit: [`68f742f` (#550)](https://github.com/pgbouncer/pgbouncer/pull/550/commits/68f742fb14991d1677d66995f38bf65320d367cb)

JelteF:
@petere could you have another look at this? As far as I can tell I addressed all your comments.

JelteF:
@petere could you take another look at this PR? I think it's good to merge.

akamath46297:
@petere could this be reviewed so that we an take this in.

include/sbuf.h
```diff
@@ -42,6 +42,8 @@ typedef enum {
 
 struct tls;
 
+extern int client_accept_sslmode;
```
petere, line 45: The use of this variable isn't clear to me. Please add a comment somewhere if appropriate.

src/admin.c
```diff
@@ -234,8 +234,13 @@ static bool admin_set(PgSocket *admin, const char *key, const char *val)
 	if (admin->admin_user) {
 		ok = set_config_param(key, val);
 		if (ok) {
+			PktBuf *buf = pktbuf_dynamic(256);
+			if (strstr(key, "_tls_") != NULL) {
```
petere, line 238: This is a bit fishy. ;-) I haven't thought about a better idea though.

src/sbuf.c
```diff
@@ -882,39 +884,52 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
 		err = tls_config_parse_protocols(&protos, protocols);
 		if (err) {
 			log_error("invalid %s_protocols: %s", pfx, protocols);
+			return false;
```
petere, line 887: I have committed these error handling changes in this function separately (4c474cabd24c0bb5d85d3e8088d82a5ad0fa3da6), since they fit with some other work to make configuration error handling more robust.

src/sbuf.c
```diff
@@ -938,56 +953,115 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+			goto failed;
+		}
+		err = tls_configure(new_client_accept_base, new_client_accept_conf);
+		if (err) {
+			log_error("TLS setup failed: %s", tls_error(new_client_accept_base));
+			goto failed;
+		}
 	}
+
+	// To change server TLS settings all connections are marked as dirty. This
```
petere, line 1031: Use `/* ... */` comments please.

src/sbuf.c
```diff
@@ -938,56 +953,115 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+	// has changed.
+	if (server_connect_conf || new_server_connect_conf) {
+		struct List *item;
+		PgPool *pool;
+		statlist_for_each(item, &pool_list) {
+			pool = container_of(item, PgPool, head);
+			// Don't tag the pgbouncer pool as dirty, since this is not an
+			// actual postgres server. Marking it as dirty breaks connecting to
+			// the pgbouncer admin database on future connections.
+			if (strcmp(pool->db->name, "pgbouncer") != 0)
```
petere, line 1047: You can detect the admin database more elegantly using `pool->db->admin`.

src/sbuf.c
```diff
@@ -938,56 +953,115 @@ static void setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+	// of the settings have changed. In all other cases we recycle the
+	// connections to be on the safe side, even though it's possible nothing
+	// has changed.
+	if (server_connect_conf || new_server_connect_conf) {
+		struct List *item;
+		PgPool *pool;
+		statlist_for_each(item, &pool_list) {
+			pool = container_of(item, PgPool, head);
+			// Don't tag the pgbouncer pool as dirty, since this is not an
+			// actual postgres server. Marking it as dirty breaks connecting to
```
petere, line 1045: I think this logic should be moved into `tag_pool_dirty()` then.

test/ssl/lib.sh
petere: You need to add `TestCA2` to the `rm -rf` line in `test/ssl/test.sh` line 5. Otherwise, subsequent runs will keep the previous run's CA setup and keep counting up the serial number.

Also, while you're at it, add the `TestCA2` directory to the top-level `.gitignore`.

test/ssl/test.sh
```diff
@@ -104,7 +106,7 @@ reconf_bouncer() {
 	done
 	test -f test.pid && kill `cat test.pid`
 	sleep 1
-	$BOUNCER_EXE -v -v -v -d tmp/test.ini
+	$BOUNCER_EXE -v -v -d tmp/test.ini
```
petere, line 109: This issue has been addressed in libusual now.

test/ssl/test.sh
```diff
@@ -237,7 +275,7 @@ test_client_ssl() {
 	echo "host all all 127.0.0.1/32 trust" > pgdata/pg_hba.conf
 	echo "host all all ::1/128 trust" >> pgdata/pg_hba.conf
 	reconf_pgsql "ssl=on" "ssl_ca_file='root.crt'"
-	psql_bouncer -q -d "dbname=p0 sslmode=require" -c "select 'client-ssl-connect'" | tee tmp/test.tmp
+	psql_bouncer -q "dbname=p0 sslmode=require" -c "select 'client-ssl-connect'" | tee tmp/test.tmp
```
petere, line 278: This change doesn't make sense.

test/ssl/test.sh
```diff
@@ -251,7 +289,177 @@ test_client_ssl_verify() {
 	echo "host all all 127.0.0.1/32 trust" > pgdata/pg_hba.conf
 	echo "host all all ::1/128 trust" >> pgdata/pg_hba.conf
 	reconf_pgsql "ssl=on" "ssl_ca_file='root.crt'"
-	psql_bouncer -q -d "dbname=p0 sslmode=verify-full sslrootcert=TestCA1/ca.crt" -c "select 'client-ssl-connect'" | tee tmp/test.tmp 2>&1
+	psql_bouncer -q "dbname=p0 sslmode=verify-full sslrootcert=TestCA1/ca.crt" -c "select 'client-ssl-connect'" | tee tmp/test.tmp 2>&1
```
petere, line 292: Here also, and also below.

test/ssl/test.sh
```diff
@@ -251,7 +289,177 @@ test_client_ssl_verify() {
...
+	grep -q "client-ssl-connect"  tmp/test.tmp || return 1
+	admin "set client_tls_key_file='TestCA2/sites/01-localhost.key'"
+	admin "set client_tls_cert_file='TestCA2/sites/01-localhost.crt'"
+	psql_bouncer -q "dbname=p0 sslmode=verify-full sslrootcert=TestCA2/ca.crt" -c "select 'client-ssl-connect'" | tee tmp/test.tmp 2>&1
+	grep -q "client-ssl-connect"  tmp/test.tmp
+	rc=$?
+	return $rc
+}
+
+
```
petere, line 350: extra blank line

petere (changes requested):

src/admin.c
JelteF: I agree, I could also remove this check and call `sbuf_tls_setup()` always.

test/ssl/test.sh
JelteF: I think I changed this when the tests were not working correctly. It didn't help, because there was an actual bug, but then I didn't change it back.

src/sbuf.c
JelteF: Sounds good, rebased these changes on top of the current master branch.

src/sbuf.c
```diff
@@ -904,7 +915,7 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
 		err = tls_config_set_ecdhecurve(conf, ecdhecurve);
 		if (err) {
 			log_error("invalid %s_ecdhecurve: %s", pfx, ecdhecurve);
-			false;
+			return false;
```
JelteF, line 918: When rebasing there was this merge conflict, I left my version since it looks like a typo on the master version. I'm surprised this didn't trigger any compiler warnings.

include/sbuf.h
JelteF: done

test/ssl/test.sh
```diff
@@ -141,7 +143,7 @@ die() {
 }
 
 admin() {
-	psql -X -h /tmp -U pgbouncer pgbouncer -c "$@;" || die "Cannot contact bouncer!"
+	psql -X -h /tmp -U pgbouncer -d pgbouncer -c "$@;" || die "Cannot contact bouncer!"
```
JelteF, line 146: Without this change the tests would now show errors similar to the ones reported here: https://www.postgresql.org/message-id/flat/a1a18525-6b24-411d-860c-404878c73e97%40k4g2000yqb.googlegroups.com#a65e9ecbf62f6a12fd9498daef0d2414

JelteF:
Thanks for the review. I now addressed all your feedback, so if you could take a final look that would be great.

test/ssl/test.sh
JelteF: I guess this started happening because of this commit: https://github.com/pgbouncer/pgbouncer/commit/c7eb9fede68bf2131ef4442ca923f4498a453b04

src/sbuf.c
JelteF: I saw this fix was included on master. Rebased this PR now to include it.

petere: Fixed separately: bfafe67ce084cea479826444fcadcf4771eac0b1

I also found another place with the same mistake. It is indeed mysterious that there was no compiler warning.

test/ssl/test.sh
petere: Fixed separately: d617fbe7148f7b8a4d5decc9b2866af845879250

src/admin.c
```diff
@@ -234,8 +234,13 @@ static bool admin_set(PgSocket *admin, const char *key, const char *val)
...
+				if (!sbuf_tls_setup())
+					pktbuf_write_Notice(buf, "TLS settings could not be applied, still using old configuration");
+			}
 			snprintf(tmp, sizeof(tmp), "SET %s=%s", key, val);
-			return admin_ready(admin, tmp);
+			return admin_flush(admin, buf, tmp);
```
petere, line 243: Any reason for this change?

petere (changes requested):

src/admin.c
JelteF: as far as I remember it was needed to show the notice when the TLS settings could not be applied.

include/sbuf.h
```diff
@@ -42,6 +42,17 @@ typedef enum {
...
+/*
+ * client_accept_sslmode is the currently applied sslmode that is used to
+ * accept client connections. This is usually the same as
+ * cf_client_tls_sslmode, except when changing the TLS configuration failed for
+ * some reason (e.g. cert file not found). In this exceptional case,
+ * cf_client_tls_sslmode will be the new sslmode, which is not actually
+ * applied. And client_accept_sslmode is the still applied previous version. So
+ * usually you should use this variable over cf_client_tls_sslmode.
+ */
+extern int client_accept_sslmode;
```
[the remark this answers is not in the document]
JelteF: Yeah we need the same for the server too, nice catch!

src/sbuf.c
```diff
@@ -54,6 +54,8 @@ enum WaitType {
 	W_ONCE
 };
 
+int client_accept_sslmode;
```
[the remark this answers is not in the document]
JelteF: done

PR #551, opened 2020-12-08
Adding a information in a log about a link of client and server at a moment of a connect.

ofigenov:
Adding a information in a log about a link of client and server at a moment of a connect.
Will info like:
2020-12-08 19:22:38.655 MSK [5508] LOG C-0x2485130: pg11/postgres@192.168.155.95:45254 login attempt: db=pg11 user=postgres tls=no
2020-12-08 19:22:38.655 MSK [5508] LOG S-0x248c040: pg11/postgres@192.168.155.95:5411 new connection to server (from 192.168.155.95:44568)
**2020-12-08 19:22:44.231 MSK [5508] LOG C-0x2485130: pg11/postgres@192.168.155.95:45254 linked client and server S-0x248c040 db=pg11 user=postgres**

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -677,6 +677,8 @@ bool find_server(PgSocket *client)
 		client->link = server;
 		server->link = client;
 		change_server_state(server, SV_ACTIVE);
+		slog_info(client, "linked client and server S-%p db=%s user=%s",
+                                   client->link, client->db->name, client->login_user->name);
 		if (varchange) {
 			server->setting_vars = 1;
 			server->ready = 0;


petere:
What is the purpose of this?

ofigenov:
> What is the purpose of this?
Hello, Peter. I wrote about this in Gitter before. It is necessary for logging facts of link between clients and servers. This is necessary so that later on, using the logs, one can find who came which a client from where and in which server process he worked.

panaxtri:
It would be very useful

src/objects.c
```diff
@@ -677,6 +677,8 @@ bool find_server(PgSocket *client)
 		client->link = server;
 		server->link = client;
 		change_server_state(server, SV_ACTIVE);
+		slog_info(client, "linked client and server S-%p db=%s user=%s",
+                                   client->link, client->db->name, client->login_user->name);
```
JelteF, line 681: I can see this being useful to debug things problems, but it seems very noisy for general logs. Merging this when changed to `slog_debug` would make sense to me.
```suggestion
		slog_debug(client, "linked client and server S-%p db=%s user=%s",
                                   client->link, client->db->name, client->login_user->name);
```

ofigenov: These messages are no more garbage than messages that are now issued about the creation of client and server facilities. Moreover, information about these objects without communication between them is not important. But just with the information, which client object was associated with which server object and give a complete picture. As for the debug, it is needed when there is already any problem and you need to repeat it, and the registration of facts of binding in the log is necessary all the time so that if someone performs some kind of malicious effect, then you can see who it was and you can see who it was and Where did he come from. Maybe it makes sense to come up with some separate option that would be in the config and include this behavior? Something as 'log_connections' and 'log_disconnections'?

ofigenov: These messages are no more garbage than messages that are now issued about the creation of client and server facilities. Moreover, information about these objects without communication between them is not important. But just with the information, which client object was associated with which server object and give a complete picture. As for the debug, it is needed when there is already any problem and you need to repeat it, and the registration of facts of binding in the log is necessary all the time so that if someone performs some kind of malicious effect, then you can see who it was and you can see who it was and Where did he come from. Maybe it makes sense to come up with some separate option that would be in the config and include this behavior? Something as 'log_connections' and 'log_disconnections'?

JelteF: > These messages are no more garbage than messages that are now issued about the creation of client and server facilities. 

In most production use cases connections to clients and servers, happen orders of magnitude less frequently then linking two already opened connections. In transaction pooling mode this new log would be shown **for every** transaction that is started. So while "more garbage" is subjective to your use, it's definitely a lot more chatty.

While I understand the general gist of your response, that having this info only present with debug logging is probably not enough for some production use cases. I think most people would not want to have these logs turned on all the time, because of the log spam it would cause. Having one more option to enable just this logging would work, but it would add even more configuration options to pgbouncer. So I'm not a huge fan of that either.

For the usecase you describe (auditing security incidents), why is it not enough to turn on the  `application_name_add_host` setting.

JelteF: Also, I realised we have a log similar to this nowadays at the noise level: https://github.com/pgbouncer/pgbouncer/blob/a7b3c0a5f4caa9dbe92743d04cf1e28c4c05806c/src/objects.c#L879

JelteF:
Closing this, since I think `application_name_add_host` is a better way of achieving what the user wants. If that doesn't work feel free to re-open.

PR #556, opened 2020-12-21
Enable ccache on Travis CI to speed up builds

PeterDaveHello:
This will speed up the builds on Travis CI 🏃‍♂️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,11 @@
 dist: focal
 language: c
+cache:
+  ccache: true
 env:
   global:
     - PGVERSION=11
+    - PATH="/usr/lib/ccache/:$PATH"
 jobs:
   include:
   - env: configure_args=''


PeterDaveHello:
Hello @petere, do you have a chance to take a look at this PR?

Though I noticed #558 was opened, before the migration was done with enough tests and verification, I believe this will still be useful ;)

petere:
Yeah, it appears that Travis CI isn't quite dead yet. I found this documentation https://docs.travis-ci.com/user/caching/#ccache-cache, which is a bit different from your proposal. Any thoughts?

PeterDaveHello:
@petere it's actually not so different, I defined cache mechanism in list because sometimes we might need multiple, `cache: ccache` or `cache: packages` won't do so. Does this sound fair to you 😄

petere:
This is obsolete now.

PeterDaveHello:
@petere isn't this still useful?

PR #564, opened 2021-01-31
Send FATAL when client is disconnected.

x4m:
Some drivers do not recognize ERRORs when they are followed by a disconnect.
If drivers receive FATAL - it's prepared for unexpected EOF.


petere:
I don't see any callers for this new function. Am I missing something?

damour:
@x4m Thank you for the fix. We experienced the same issue and tried the patched version of pgbouncer but still have an issue with "unexpected EOF" error. Do we need also to apply some additional fixes on the library side like https://github.com/jackc/pgconn/pull/64 ?

x4m:
@damour this PR ought to be enough. Consider using Odyssey, BTW, I've already merged a similar patch there :)

PR #565, opened 2021-02-04
Fixed multiple sequential queries warning logging

### LINKED ISSUE #277: Logs spammed with FIXME: query end, but query_start == 0 and now also FIXME: transaction end, but xact_start == 0

vonzshik:
Fixes #277

The problem here is if multiple queries ("BEGIN TRANSACTION ...", "UPDATE ...") are sent together (that is, we're not waiting for a response from "BEGIN TRANSACTION ..." and immediately send another query to save a roundtrip), the query's `query_time` is set to 0 on the first query response, so there will be a warning for the second one.
The proposed solution is to reset the `query_time` only on the last RFQ. While it's not actually correct, the only other idea I have is to add a queue, which seems like a lot of overhead.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -354,24 +354,24 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 			sbuf_prepare_send(sbuf, &client->sbuf, pkt->len);
 
 			/* every statement (independent or in a transaction) counts as a query */
-			if ((ready || idle_tx) && client->query_start) {
+			if ((ready || idle_tx) && client->query_start && client->expect_rfq_count == 0) {
 				usec_t total;
 				total = get_cached_time() - client->query_start;
 				client->query_start = 0;
 				server->pool->stats.query_time += total;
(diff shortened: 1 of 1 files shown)


nathan-c:
Is there any news on when this might be merged/released?

petere:
Well, this certainly explains a lot. Especially how many people persistently see this and it's hard to reproduce for others, since it requires queries to be sent in this pipelined kind of way. I was now able to reproduce this with PostgreSQL 15's new libpq pipeline mode.

I agree your solution is sensible. I have also taken this opportunity to restructure this code a bit and linked in all the related GitHub issues in #628.

vonzshik:
@petere great, thank you! Then I'm closing this pr in favor of #628.

PR #569, opened 2021-02-11
Fix crash on loging into the admin database

### LINKED ISSUE #568: Crash with auth_user when connecting to the admin database
Hello

It seems 383af567605fadceb351833ac346313500f13687 introduced a severe regression. Making auth_user global that way made it possible to crash pgbouncer by simply doing `psql -p 6432 -U does_not_exists pgbouncer`
When this is done, pgbouncer writes in its log the following line and exits:
```
2021-02-11 11:08:25.338 UTC [22164] FATAL @src/objects.c:312 in function put_in_order(): put_in_order: found existing elem
```

PierreDucroquet:
The admin database is restricted, the global auth_user setting does not apply to it.
Not doing so crashes pgbouncer when an unknown user logs in to the admin db.

fix #568

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -339,7 +339,7 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 			 * no auth_user is set for the db, see if the
 			 * global auth_user is set and use that.
 			 */
-			if (!client->db->auth_user && cf_auth_user) {
+			if (!client->db->auth_user && cf_auth_user && !client->db->admin) {
 				client->db->auth_user = find_user(cf_auth_user);
 				if (!client->db->auth_user)
 					client->db->auth_user = add_user(cf_auth_user, "");


petere:
I don't think this is the right solution for the original problem. As discussed there, the root cause there is really in the way users are managed internally in pgbouncer. Restricting how users can be used just papers over one particular symptom but doesn't fix anything. AFAICT, the crash only happens when you use "pgbouncer" as the auth_user, so there is an available workaround.

petere:
I don't think this is the right solution for the original problem. As discussed there, the root cause there is really in the way users are managed internally in pgbouncer. Restricting how users can be used just papers over one particular symptom but doesn't fix anything. AFAICT, the crash only happens when you use "pgbouncer" as the auth_user, so there is an available workaround.

sdigit:
> I don't think this is the right solution for the original problem. As discussed there, the root cause there is really in the way users are managed internally in pgbouncer. Restricting how users can be used just papers over one particular symptom but doesn't fix anything. AFAICT, the crash only happens when you use "pgbouncer" as the auth_user, so there is an available workaround.

I get this crash when attempting to login to `postgres` when `postgres` is not configured in `pgbouncer.ini`; the username is not `pgbouncer` in my case. I just ran into it when connected to a database I wanted to drop, trying to switch to `postgres`.

JelteF:
The issue that this tries to fix is almost certainly fixed by #1052

PR #573, opened 2021-02-20
Authentication for admin database

erthalion:
Hi,

From what I understand, if `stats_users` specified in configuration, it needs
to be present in `auth_file` to be able to connect to the admin db. This is the
case even if `stats_users` could be authenticated via `auth_user` +
`auth_query` for all other "regular" databases. Looks like it happens because
the admin db is a special "fake" database without an `auth_user`. If pgbouncer
would be able to authenticate `stats_users` connecting to the admin db using
`auth_query` as well, this will be helpful because one doesn't need to maintain
another set of credentials for a monitoring user.

From the first glance there are just few bits missing to implement this. E.g. a
`default_auth_db` to connect to for authentication purposes in case of admin
db. Does this implementation make sense?

This idea also partially intersects with #569, but instead of restricting this
patch for admin databases it handles them differently.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -32,3 +32,4 @@ on_failure:
   - set HOME=.
   - sh -l -c "test -f config.status || cat config.log"
   - sh -l -c "find test/ -name '*.log' | xargs tail -n40 -v"
+  - ls -l / && ls -l /tmp
--- a/doc/config.md
+++ b/doc/config.md
@@ -141,6 +141,14 @@ is used, it needs to be installed into each database.
 
 Default: `SELECT usename, passwd FROM pg_shadow WHERE usename=$1`
 
+### default_auth_db
+
+Default database to connect to for authentication purposes. Used for
+authentication of `stats_users` connecting to admin database.
+
+Default: `postgres`
+
+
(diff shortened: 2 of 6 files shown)


erthalion:
Looks like appveyor tests are failing, because postgres version has changed since the last build (from postgresql-12.4-1 to postgresql-13.1-1). Probably something needs to be adjusted in the postgres config, looking into this, but I guess it's not relevant for this particular patch:

```
2021-02-21 12:37:45.633 UTC [5280] LOG:  listening on IPv6 address "::1", port 6666
2021-02-21 12:37:45.633 UTC [5280] LOG:  listening on IPv4 address "127.0.0.1", port 6666
2021-02-21 12:37:45.648 UTC [5280] FATAL:  could not create lock file "/tmp/.s.PGSQL.6666.lock": No such file or directory
2021-02-21 12:37:45.648 UTC [5280] LOG:  database system is shut down
 stopped waiting
pg_ctl: could not start server
Examine the log output.
```

JelteF:
I think this works nowadays if you use `auth_dbname`. Feel free to open a new PR if that's not the case.

PR #574, opened 2021-02-24
Removed userlist reload from the maintenance

### LINKED ISSUE #377: stat() of auth_file blocking connections
If you strace a pgbouncer PID you'll see that the auth_file is constantly being stat'd. If there is a high disk i/o event on the disk that stores the userlist.txt, pgbouncer itself seems to not accept new connections until the stat() is able to succeed.

We ran into this issue with a heavy logrotation done on the / partition which stores the auth_file, while postgres db was running on a different set of disks. 

Output from strace:
```recvfrom(217, "2\0\0\0\4C\0\0\0\rSELECT 0\0Z\0\0\0\5I", 4096, 0, NULL, NULL) = 25
sendto(172, "2\0\0\0\4C\0\0\0\rSELECT 0\0Z\0\0\0\5I", 25, 0, NULL, 0) = 25
epoll_wait(4, [], 512, 225)             = 0
stat("/etc/pgbouncer/userlist.txt", {st_mode=S_IFREG|0600, st_size=489, ...}) = 0
epoll_wait(4, [{EPOLLIN, {u32=123, u64=123}}], 512, 334) = 1
```

This is a **theory** based on pgbouncer connection stats, disk i/o, and a heavy log rotation that happened on the root partition when our app failed to connect. I will try to reproduce this in a lab, but would like to know if this is possible that a delay in the stat() call of the auth_file could block/hinder connections to pgbouncer. 
If so, I would recommend a flag for caching/disabling the checking of changes to this file.

Thanks

vonzshik:
Fixes #377

### DIFF
--- a/src/janitor.c
+++ b/src/janitor.c
@@ -664,9 +664,6 @@ static void do_full_maint(evutil_socket_t sock, short flags, void *arg)
 		return;
 	}
 
-	if (requires_auth_file(cf_auth_type))
-		loader_users_check();
-
 	adns_zone_cache_maint(adns);
 }
 


erthalion:
Indeed, looks like a good way to fix doing frequent stat on auth_file. I'm only curious if originally it was done like that on purpose and someone is using such reload? Maybe just in case add a configuration parameter to specify whether or not to do loader_users_check here?

vonzshik:
> Maybe just in case add a configuration parameter to specify whether or not to do loader_users_check here?

Sure, if the reviewer will request it.

PR #583, opened 2021-04-07
Revise server host IP address detection and handling

gibeom-gwon:
Use getaddrinfo(3) to detect and convert server host IP address.
Additionally, this commit sets sin6_scope_id, you can use link-local
IPv6 address as a server address.(e.g., host=fe80::...%eth0)


petere:
What is the benefit of this change?

gibeom-gwon:
As wrote in description, we can connect PostgreSQL server with IPv6 link-local address.

JelteF:
@gibeom-gwon are you still interested in working on this. I think this needs some tests to be able to be merged, or if automated tests are hard, it needs at the very least a step-by-step explanation of what usage this change makes possible.

Issue #588, opened 2021-05-06
Log message does not provide enough context


pgodfrin-nov:
Our install, with Kubernetes, combines the postgres log with the pgbouncer log. When pgbouncer issues a message it's not clear that the source is pgbouncer, the second line does provide a filename, so that helps.
```
2021-05-04 15:09:25.484 UTC [1] FATAL cannot load config file
2021-05-04 15:09:25.484 UTC [1] ERROR syntax error in configuration (/etc/pgbouncer/pgbouncer-sidecar.ini:33), stopping loading
```
Would it be helpful to have an option for format the log header, similar to the postgres log_line_prefix parameter?

petere:
I don't think we want to start worrying about disambiguating our own log messages in case the logs are combined with some other software. If you are doing that, you might want to use syslog, so that the source is automatically tagged.

eulerto:
I would say if your Kubernetes installation is combining the logs from 2 different softwares, it should provide a tag for each log message too. Postgres doesn't provide a tag either (application name is not a tag; there are a lot of messages that won't provide it).

Issue #584, opened 2021-04-15
The message for a bad %include directive lacks context


cbandy:
When the configuration file specifies an `%include` of a file that does not exist, `pgbouncer` exits with a message that does not provide enough information to correct the problem.

For example, this configuration `%include`s two files, one that exists and one that does not. The FATAL message does not say which line or which file caused the failure.

```
● pgbouncer --version
PgBouncer 1.15.0
libevent 2.1.8-stable
adns: evdns2
tls: OpenSSL 1.1.1g FIPS  21 Apr 2020
systemd: yes

● cat /tmp/pgbouncer.ini
[pgbouncer]
%include /dev/null
%include /not/a/file

● pgbouncer /tmp/pgbouncer.ini
2021-04-15 18:02:24.571 UTC [15] FATAL cannot load config file
```

Increasing the verbosity provides some context:

```
● pgbouncer -v /tmp/pgbouncer.ini
2021-04-15 18:06:06.158 UTC [16] DEBUG parse_ini_file: [pgbouncer]
2021-04-15 18:06:06.158 UTC [16] DEBUG processing include: /dev/null
2021-04-15 18:06:06.158 UTC [16] DEBUG returned to processing file /tmp/pgbouncer.ini
2021-04-15 18:06:06.158 UTC [16] DEBUG processing include: /not/a/file
2021-04-15 18:06:06.158 UTC [16] FATAL cannot load config file
```

The results are similar when reloading:

```
● cat /tmp/pgbouncer.ini
[pgbouncer]

● pgbouncer /tmp/pgbouncer.ini
2021-04-15 18:38:42.423 UTC [21] LOG kernel file descriptor limit: 1048576 (hard: 1048576); max_client_conn: 100, max expected fd use: 112
2021-04-15 18:38:42.424 UTC [21] LOG listening on unix:/tmp/.s.PGSQL.6432
2021-04-15 18:38:42.424 UTC [21] LOG process up: PgBouncer 1.15.0, libevent 2.1.8-stable (epoll), adns: evdns2, tls: OpenSSL 1.1.1g FIPS  21 Apr 2020

# add "%include /not/a/file" to the file and send SIGHUP

2021-04-15 18:39:05.117 UTC [21] LOG got SIGHUP, re-reading config
2021-04-15 18:39:05.118 UTC [21] WARNING config file loading failed
```

An `ERROR` message similar to the one for bad values would be very helpful:

```
● cat /tmp/pgbouncer.ini
[...]

PR #592, opened 2021-05-11
Clarify messages

eulerto:
Inform the config file name in the messages.

Issues #584 and #588 complained about not enough context for this message. Further work should be done in `libusual` since it is responsible to parse configuration files.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -402,7 +402,7 @@ void load_config(void)
 			loader_users_check();
 		loaded = true;
 	} else if (!loaded) {
-		die("cannot load config file");
+		die("cannot load config file %s", cf_config_file);
 	} else {
 		log_warning("config file loading failed");
 		/* if ini file missing, don't kill anybody */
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -402,7 +402,7 @@ void load_config(void)
 			loader_users_check();
 		loaded = true;
 	} else if (!loaded) {
-		die("cannot load config file");
+		die("cannot load config file %s", cf_config_file);
```
petere, line 405: This is okay. We should probably quote the file name, like in other messages. Also, there is a `log_warning` below that should perhaps be adjusted in the same way?

src/main.c
```diff
@@ -495,7 +495,7 @@ static void handle_sigusr2(int sock, short flags, void *arg)
 
 static void handle_sighup(int sock, short flags, void *arg)
 {
-	log_info("got SIGHUP, re-reading config");
+	log_info("got SIGHUP, re-reading config file %s", cf_config_file);
```
petere, line 498: I don't think this is appropriate. The message refers to reloading the whole configuration, not only a particular file.

petere (changes requested):

Issue #585, opened 2021-04-19
Support a way to fast-fail queries when pgbouncer can't connect to the underlying database


james727:
We're operating a distributed Postgres cluster where all data is replicated to multiple databases. Failover happens automatically when a single node is down. Thus, in situations where a given database is momentarily unaccessible, we'd like queries to that database to fail fast, instead of queueing up in pgbouncer.

It'd be nice to be able to specify an option in the configuration file that prevents this from happening. I understand `query_wait_timeout` is the recommended approach here but during peak load periods we'd like to allow queries to queue up if need be. I'm not sure the best configuration interface here, but from my perspective something like `query_db_availability_timeout` would be useful. This could be a similar concept to `query_wait_timeout`, except it would represent a limit on the amount of time spent waiting for connection acquisition while `sv_active + sv_idle` is less than the maximum number of connections for the underlying database.

petere:
Isn't this supposed to be the behavior of the existing fast-fail check? The comment at the top of `check_fast_fail()` says as much:

https://github.com/pgbouncer/pgbouncer/blob/ab7957160b2ec92a183f699e3112b2d0e10bfbbb/src/objects.c#L569

Is this different from what you want?

dimikot:
@petere BTW in this fast-fail usecase, mentioning "server_login_retry" in "server login has been failing, try again later (server_login_retry)" error message is a little misleading. If I understand the code and the behavior correctly (at lease I am able to reproduce it), the error is delivered to the client way before server_login_retry timeout expires (because it's fast-fail). I spent several hours trying to understand, why increasing server_login_retry change doesn't make any effect on the behavior (since I saw "server_login_retry" in the error message), and then looked at the source code and understood, why.

Maybe it's worth mentioned in this error instead, that it's a fast-fail situation, and remove mentions of server_login_retry?

![CleanShot 2024-01-27 at 18 39 22@2x](https://github.com/pgbouncer/pgbouncer/assets/65643/dac292d8-47b0-4b1f-865c-beca3fcfc2a3)

dimikot:
@petere 

> Isn't this supposed to be the behavior of the existing fast-fail check?
> Is this different from what you want?

I think it IS different, yes. I just recorded a short screencast (below) with steps to reproduce. The current behavior of 1.21.0 is following.

1. I run `psql` which opens connection to pgbouncer, but I don't run any queries yet.
2. Then I ^C the PG docker instance.
[...]

PR #594, opened 2021-05-13
Add query_db_wait_timeout configuration option

james727:
Hello PgBouncer team. I'd like feedback on a new configuration field. This adds a parameter (`query_db_wait_timeout`) to address the problem described in in #585.

Setting `query_db_wait_timeout` to a nonzero value will disconnect a client if the following two things are true:
1. It's been waiting to be assigned a server connection for longer than query_db_wait_timeout
2. The underlying pool is detected to be in a fast-failure state, as determined by existing logic for fast failing new connections.

Note that this does not require the pool to have been in a fast failure state for `query_db_wait_timeout` seconds, just that the client has been waiting and the pool is currently in a broken state when the bookkeeping occurs.

Thanks in advance.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -728,6 +728,20 @@ If this is disabled, clients will be queued indefinitely.
 
 Default: 120
 
+### query_db_wait_timeout
+
+Maximum time queries are allowed to spend waiting for execution while the underlying
+database is inaccessible. If the query is not assigned to a server during this time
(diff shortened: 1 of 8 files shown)


JelteF:
Given its age and the amount of merge conflicts I'm going to close this PR. As described in the related issue I think it makes sense to have a shorter query_wait_timeout during when fast-fail is happening for new connections though. So new PRs are welcome.

Issue #578, opened 2021-03-10
Crash: FATAL @src/takeover.c:373 in function takeover_login_failed(): login failed


wit:
When authorizing in the admin database, the application crashes.
To reproduce, you need to reload the configuration a couple of times.
If the app has recently been launched, it won't crash.
Version pgbouncer-1.15.0-1

```
2021-03-10 19:18:07.736 MSK [63738] WARNING server login failed: FATAL password authentication failed for user "pgbouncer"
2021-03-10 19:18:07.736 MSK [63738] WARNING C-0x2675d78: pgbouncer/pgbouncer@[::1]:57584 pooler error: password authentication failed for user "pgbouncer"
2021-03-10 19:18:07.736 MSK [63738] FATAL @src/takeover.c:373 in function takeover_login_failed(): login failed

```

```
[databases]
pgbouncer = dbname=pgbouncer host=127.0.0.1 port=5432

[pgbouncer]
logfile = /var/log/pgbouncer/pgbouncer.log
pidfile = /var/run/pgbouncer/pgbouncer.pid
listen_addr = *
listen_port = 6432
unix_socket_dir = /var/run/pgbouncer
auth_type = md5
auth_file = /etc/pgbouncer/userlist.txt

admin_users = pgbouncer
stats_users = pgbouncer,
pool_mode = transaction
server_reset_query =
ignore_startup_parameters = extra_float_digits
max_client_conn = 5000
default_pool_size = 1000
reserve_pool_size = 0
log_connections = 0
log_disconnections = 0
log_pooler_errors = 1
verbose=0
server_lifetime = 300
server_idle_timeout = 15
server_connect_timeout = 10
server_login_retry = 10
tcp_keepalive = 1
```

petere:
fixed by 952381bf9651c8214b80fb6e48cf3c2f80835119

PR #598, opened 2021-05-15
Skip special database pgbouncer

eulerto:
If the user set "pgbouncer" as a database in the "databases" section,
skip it. This database entry is created by admin_setup() with certain
options. I also added a note saying so.

This is related to issue #134

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -885,6 +885,9 @@ Such automatically created database entries are cleaned up
 if they stay idle longer than the time specified by the `autodb_idle_timeout`
 parameter.
 
+The special database **pgbouncer** is reserved. An error will be reported if
+you include it here.
+
 ### dbname
 
 Destination database name.
--- a/src/loader.c
+++ b/src/loader.c
@@ -201,6 +201,12 @@ bool parse_database(void *base, const char *name, const char *connstr)
 	cv.value_p = &pool_mode;
 	cv.extra = (const void *)pool_mode_map;
 
+	/* pgbouncer pool is set by admin_setup() */
+	if (strcmp(name, "pgbouncer") == 0) {
+		log_error("database %s is reserved", name);
+		return true;
+	}
+
 	if (strcmp(name, "*") == 0) {
 		set_autodb(connstr);
 		return true;


eulerto:
This PR fixes #578.

petere:
How about we make having a `pgbouncer` entry in `[databases]` an error instead?

eulerto:
I promoted it to an error.

doc/config.md
```diff
@@ -885,6 +885,9 @@ Such automatically created database entries are cleaned up
 if they stay idle longer than the time specified by the `autodb_idle_timeout`
 parameter.
 
+The special database **pgbouncer** does not need to be included here, it is
```
petere, line 888: This wording is not strong enough, given that it is an error. I would just leave this out.

src/loader.c
```diff
@@ -201,6 +201,12 @@ bool parse_database(void *base, const char *name, const char *connstr)
 	cv.value_p = &pool_mode;
 	cv.extra = (const void *)pool_mode_map;
 
+	/* pgbouncer pool is set by admin_setup() */
+	if (strcmp(name, "pgbouncer") == 0) {
+		log_error("skipping database %s because it is set internally", name);
+		return true;
```
petere, line 207: I'm surprised by the error handling in this function. It seems true is returned even in error cases. But maybe we should at least use `goto fail` like other cases?

The error message isn't really strong enough for an error. It should be something like "database name %s is reserved".

petere (changes requested):

src/loader.c
eulerto: `goto` was the first thing I thought but decided not to use it because I would have to add an `if` to handle `free`. The other option is to move the code a few lines below (after `strdup`) but I decided that it doesn't worth it.

I will change the wording as you suggested.

doc/config.md
eulerto: I will remove it.

petere:
I think the error handling in `parse_database()` is kind of bogus. I put together #616 to improve this. That would also help the error handling in this patch.

PR #600, opened 2021-05-17
Include number of query cancellations into SHOW POOLS

eulerto:
This commit includes a new metricto SHOW POOLS command: cl_cancel_req.
It informs the number of clients that request query cancellations.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -365,6 +365,9 @@ cl_active
 cl_waiting
 :   Client connections that have sent queries but have not yet got a server connection.
 
+cl_cancel_req
+:   Client connections that have not forwarded query cancellations to the server yet.
+
 sv_active
 :   Server connections that are linked to a client.
 
--- a/src/admin.c
+++ b/src/admin.c
@@ -810,9 +810,10 @@ static bool admin_show_pools(PgSocket *admin, const char *arg)
 		admin_error(admin, "no mem");
 		return true;
 	}
-	pktbuf_write_RowDescription(buf, "ssiiiiiiiiis",
+	pktbuf_write_RowDescription(buf, "ssiiiiiiiiiis",
 				    "database", "user",
 				    "cl_active", "cl_waiting",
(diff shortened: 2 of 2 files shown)


doc/usage.md
```diff
@@ -365,6 +365,9 @@ cl_active
 cl_waiting
 :   Client connections that have sent queries but have not yet got a server connection.
 
+cl_cancel_req
+:   Client connections that have sent query cancellations.
```
petere, line 369: I think this description is a bit misleading. It seems to imply that it is an all-time count of cancel requests that have passed through the pool. But it's actually just the count of cancel requests that have not been forwarded to the server yet, similar to `cl_waiting`. Let's find a better description.

petere (changes requested):

eulerto:
I improved the description and commit message.

petere:
After the recent changes to prioritize cancel requests, is this still necessary? It will be almost impossible to see anything but zero in this field now.

eulerto:
Fair point. However, since it is now limited to 2 x pool size, I'm afraid that a small pool reaches its limit.

Another idea is to add a WARNING to cover these cases (small pool and high number of query cancellations) such as:

```c
  /* when a cancel request is queued allow connections up to twice the pool size */
  if (!statlist_empty(&pool->cancel_req_list) && max < (2 * pool->db->pool_size)) {
    log_debug("launch_new_connection: bypass pool limitations for cancel request");
    goto force_new;
  }
  else if (!statlist_empty(&pool->cancel_req_list)) {
    log_warning("pool %s: query cancellation requests were queued because the connection limit per pool was reached", pool->db->name);
  }
```

We might improve this message informing the `2 * pool size`.

jcoleman:
In my opinion the fact that various bug fixes have been committed around this list is orthogonal to whether this gets merged -- those issues would have been far easier to diagnose, I believe, with this feature added. So it seems to me that it's worth having to be able to expose errors if more are found in the future.

PR #603, opened 2021-05-19
Use true and false for bool constants consistently

petere:
Some code was using 0 and 1, although most newer code seems to prefer
true and false.

Also, turn the declaration of some variables from integer to bool if
they are actually used as Booleans.


eulerto:
Since you are doing this refactor, you could possibly adjust some variables of the `per_loop_maint` function too.

petere:
> Since you are doing this refactor, you could possibly adjust some variables of the `per_loop_maint` function too.

done

PR #614, opened 2021-06-09
Fix md5 authentication overwriting user name in memory

petere:
When using md5 authentication with a plain-text password, the old code
did the inner md5() call and overwrote the password in the in-memory
structures with that result.  That makes sense if md5 is the only
authentication method you ever use in an instance, because then you
don't need to do that call on every client login.  But with hba, a
different authentication method could be used next time, but this
behavior destroys the plain-text password.  For example, logging in
with md5 first will make a subsequent scram login fail, but not vice
versa.  To fix, don't cache the md5() result but just compute it every
time it is needed.  (We could in theory create a separate field for
caching the md5() output, but since this method is obsolescent, let's
not bother.)

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -37,7 +37,6 @@ static const char *hdr2hex(const struct MBuf *data, char *buf, unsigned buflen)
 
 static bool check_client_passwd(PgSocket *client, const char *passwd)
 {
-	char md5[MD5_PASSWD_LEN + 1];
 	PgUser *user = client->login_user;
 	int auth_type = client->client_auth_type;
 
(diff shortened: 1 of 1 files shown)


eulerto:
Looks sane to me. You should probably fix some cosmetic issues.

* `case AUTH_MD5`: bracket shouldn't be in a new line
* else: should use brackets or add a new line after it for clarity

petere:
> Looks sane to me. You should probably fix some cosmetic issues.
> 
> * `case AUTH_MD5`: bracket shouldn't be in a new line
> * else: should use brackets or add a new line after it for clarity

both fixed

eshkinkot:
BWT this patch also fix problem with plain-text password authentication, without it pgbouncer send MD5 hash on server's plaintext password authentication request

PR #621, opened 2021-06-29
Required dependency libraries on Windows

eulerto:
This PR contains 2 commits. The first one installs all required libraries for PgBouncer on Windows. It means that when you `make install`, you don't need to copy some dependency libraries from a MinGW directory. The other commit allows `make zip` to include all required libraries into a final zip.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,6 +131,7 @@ ifeq ($(cares_support),yes)
 dep_win_libs_abs += $(wildcard $(dep_win_dir)/libcares*.dll)
 endif
 dist_bin_DATA = $(dep_win_libs_abs)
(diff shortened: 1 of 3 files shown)


petere:
The AppVeyor build indicates that this does not work.

eulerto:
The problem is that Appveyor installs PgBouncer in the same directory as the required libraries. It is not a common pattern in my opinion. Anyway, the fix is simple: copy the required libraries if and only if the installation directory is different from the required libraries.

petere:
I wonder whether it would be better to build static binaries. I have tried this here: https://ci.appveyor.com/project/petere/pgbouncer/builds/39995929, but it doesn't seem to work quite yet.

PR #629, opened 2021-07-09
Don't close connection with query timeout when no query is in flight

### LINKED ISSUE #579: query_timeout limits whole transaction time
`query_timeout` closes any connection with a transaction older than query_timeout.

Seems like the same issue as https://github.com/pgbouncer/pgbouncer/issues/22 which was meant to be fixed 2 years ago.

Steps to repro:

```
> BEGIN;
> /* wait query_timeout seconds */
> SELECT 1;
ERROR:  query timeout
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

Tested with 1.15.0 and 1.12.0.

jcoleman:
Fixes #579.

### DIFF
--- a/src/janitor.c
+++ b/src/janitor.c
@@ -541,7 +541,8 @@ static void pool_server_maint(PgPool *pool)
 			age_client = now - server->link->request_time;
 			age_server = now - server->request_time;
 
-			if (cf_query_timeout > 0 && age_client > cf_query_timeout) {
+			if (cf_query_timeout > 0 && age_client > cf_query_timeout &&
+					!server->idle_tx) {
 				disconnect_server(server, true, "query timeout");
 			} else if (cf_idle_transaction_timeout > 0 &&
 				   server->idle_tx &&


JelteF (changes requested):
I agree with Peter here that this is currently not the right fix, also this should have a test.

JelteF:
To clarify, I agree that there is a bug here. But the we should not fix it in the way that this PR currently does. The suggestion from @petere seems like a better approach.

JelteF:
@jcoleman do you still intend to address the feedback?

jcoleman:
@JelteF I'd lost track of this entirely, but I'll put it on my list to try to take a look at again.

imsurajmishra:
@JelteF 
i think the proposed feedback is to calculate the query timeout based on actual `query_start` and not the client `request_time`.
in `pool_server_main` method
from:
age_client = now - server->link->request_time;
To
`age_query = (server->query_start !=0) ? now - server->query_start : 0;`

jcoleman:
@JelteF Perhaps slightly late for an introduction, but @imsurajmishra is on one of my teams, and he's interested in contributing to Postgres-adjacent open source, so I asked if he'd like to take a look at this PR to update it per the feedback.

JelteF:
@imsurajmishra yeah that sounds correct.

PR #630, opened 2021-07-09
Use life_over function instead of reimplementing logic

jcoleman:
I don't have a bug report to make with this, but it seems like a minor, but worthwhile, cleanup.


jcoleman:
Heh, I didn't take the time to test this, and it has compilation errors. I'd be happy to fix those, but I'd like input on whether maintainers agree that it's a useful change first.

petere (changes requested):
I think this would be sensible, but as is it doesn't compile, since `life_over()` is static in a different file. Once that is fixed, additional code can probably be cleaned away in `check_unused_servers()`, such as the `lifetime_kill_gap` variable.

jcoleman:
@petere Fixed.

PR #631, opened 2021-07-09
Fix DNS error treated as success

jcoleman:
We have a failed DNS lookup, and even log as such at the debug level,
but then instruct the DNS result function that our result status was a
success, meaning we won't log_warning that failure as well as incorrect
update the cached results and ttl.


petere:
I'm not sure your analysis is correct. In the error case, `res` will be NULL, which will still trigger the error path in `got_result_gai()`.

jcoleman:
Ah, I see that now. The first arg IMO is still confusing in this case (though maybe it implies there's an even cleaner refactoring that would remove the multiple ways it can be treated as a failure case).

PR #645, opened 2021-10-06
feat: introduce support for the `auth_dbname` option

gbartolini:
Introduce a global or database-specific option (specified in the
connection parameters string) called `auth_dbname` which points
to the name of a database defined in the `[databases]` section
to be reserved for authentication purposes.

If required, the `auth_query` will be run against this database.

By default it is not set. It requires that `auth_user` is set.

Signed-off-by: Gabriele Bartolini <gabriele.bartolini@enterprisedb.com>
Signed-off-by: Leonardo Cecchi <leonardo.cecchi@enterprisedb.com>

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -148,6 +148,18 @@ is used, it needs to be installed into each database.
 
 Default: `SELECT usename, passwd FROM pg_shadow WHERE usename=$1`
 
+### auth_dbname
+
+Name of the database in the `[databases]` section to be used for
+authentication purposes. This option can be either global or overridden
+in the connection parameters string of a given database (including the
+automated fallback database `*`).
+
+When defined, the authentication query will be run on this database
+via the specified `auth_user` option.
+
+Default: not set
+
 ### auth_user
 
 If `auth_user` is set, then any user not specified in `auth_file` will be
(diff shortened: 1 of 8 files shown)


3nprob:
Would be great to get this moving. I think many others have been running with self-built binaries with the existing #315 (which seems to have been blocked due to missing configuratIion)

JelteF:
Needs a rebase before this can be considered. Also valgrind seems to have failed for the old code. The feature itself seems useful, mostly so you don't need to define the `SECURITY DEFINER` function in each postgres database.

JelteF (changes requested):
needs a rebase and valgrind fixes

alvherre-edb:
There's large community demand for this feature, as seen here: https://github.com/pgbouncer/pgbouncer/pull/315  Please merge it soon.

JelteF:
Both me and and @eulerto left reviews on this PR without addressing those, this cannot be merged.

If you want this feature feel free to make these changes (possibly in a new PR if the original author does not have time).

goksgie:
Implemented here: https://github.com/pgbouncer/pgbouncer/pull/764

eulerto:
Refer to #764 .

Issue #647, opened 2021-10-19
Connection thrashing observed when min_pool_size is bigger than a particular user's max_user_connections


waltercacau:
Hi there,

We have this use case of a PgBouncer instance that has a few users talking to a single PostgreSQL db and we want to configure different pool sizes for them.

For the purposes of describing the issue, let's say we have two users application and background. We would like to have the pool for (db, application) to keep 30 connections in its pool at all times and for the pool (db, background) to stay with 2 connections at all time. To achieve that, we originally:
- set min_pool_size = 30
- set default_pool_size = 30
- set application's user max_user_connections to 30
- set default max_user_connections to 2

This seems to achieve the desired result but we see constant streams of this in the logs:
```
2021-10-15 18:06:17.994 UTC [7] LOG S-0x55832986dcf0: db/background@address new connection to server (from 172.30.121.106:56060)
2021-10-15 18:06:17.993 UTC [7] LOG S-0x558329870e70: db/background@address closing because: evicted (age=0s)
```

And apparently PG logs indicates that there are many short lived connections (like a close immediately after the connection is created). So this appears to be causing connection thrashing.

After analyzing the code a bit, I believe that this is caused because when the janitor is trying to ensure there are a minimum number of connections in the pool, it does not consider max_user_connections:
https://github.com/pgbouncer/pgbouncer/blob/aace87ef7e7cd9642ed76f2eea1fc0476355b411/src/janitor.c#L490

Could we change the code to also cap pool size by max_user_connections ? My rationale is basically that if a pool has max_user_connections = X, it real pool size can never be more than X. In fact it may practically become lower than X if there are multiple DBs configured with the same user, but at least this prevents connection thrashing in the simple use case of a single DB per user.

Thoughts? Does this seem like a reasonable solution?

Happy to try to contribute a PR, just wanted your opinion on this solution and to make sure I didn't miss anything. Thanks in advance!

petere:
I'm on the fence about this one. If the user says, I want a minimum pool size of X, and I want at most Y connections, and Y <= X, then who says that Y should win? If this is a nonsensical configuration, maybe the user should just fix it? Is there a real use case that would require this kind of configuration to work?

My concern is, PgBouncer by its nature has a lot of settings like this, how many connections, various timeouts, and some of those naturally conflict with each other. If we start addressing each of those conflicts individually, then we might never see the end of it. I'm not opposed to this one in particular, but I wonder if there is a real use for it or if it's just a user mistake.

[...]

PR #648, opened 2021-10-21
Don't evict connections to achieve min_pool_size

waltercacau:
This should prevent us from evicting older idle connections unnecessarily if the max connection limit for the user or db has been reached.

This should address issue: https://github.com/pgbouncer/pgbouncer/issues/647

I have ran `test/test.sh` and all tests are passing (including two newly added ones).
I have also tested locally reverting the checks I added to janitor and verified that the newly added tests fail.

### DIFF
--- a/include/objects.h
+++ b/include/objects.h
@@ -39,6 +39,7 @@ bool find_server(PgSocket *client)		_MUSTCHECK;
 bool release_server(PgSocket *server)		/* _MUSTCHECK */;
 bool finish_client_login(PgSocket *client)	_MUSTCHECK;
 bool check_fast_fail(PgSocket *client)		_MUSTCHECK;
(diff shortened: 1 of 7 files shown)


mmontagna:
+1

atanner27:
+1

petere:
I wonder whether a different way of structuring this might be possible. I don't like that we have to repeat the whole logic of checking the parameters in your `check_max_connection_limits()`. (Also, by the way, the return value of that function is not obvious. What does "true" or "false" as a return from "check" mean?) Perhaps we could add an argument to `launch_new_connection()` that specifies whether we want to run the "evict" code.

The tests are nice, but they each add >=7 seconds to the total run time, which seems quite a lot. Is there a way to get that shorter?

PR #652, opened 2021-11-10
Fail sbuf_send_pending operation if dst socket is closed

kmichel-aiven:
If the dst socket is closed, `sbuf_send_pending` will fail but not mark the operation as failed.

It will then retry repeatedly, consuming a lot of CPU until pgbouncer is restarted.

### DIFF
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -520,6 +520,7 @@ try_more:
 
 	if (sbuf->dst->sock == 0) {
 		log_error("sbuf_send_pending: no dst sock?");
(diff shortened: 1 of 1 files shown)


JelteF:
I'm not sure how it happens. It doesn't happen often. But when it happens it's really bad. pgbouncer starts using 100% of it's CPU core and starts flooding the logs with errors. The only way to stop it is by restarting it.

mtuncer:
@petere we are looking forward to this fix

guybrowne:
Is this likely to make it into the next release? we started hitting this error in version 1.12 have then updated to 1.16 & 1.17 where we are still experiencing it... its very intermittent.

PR #666, opened 2022-01-04
Add PgBouncer peering support

JelteF:
For both performance and availability reasons it can be useful to load
balance across multiple PgBouncer processes, possibly on different
machines. It's very easy to do, by simply putting a generic TCP load
balancer in front of these pgbouncer processes. The `so_reuseport`
option is an example of such a load balancer, which uses the OS kernel
to do the load balancing.

However, there is a problem with this simple approach. Query
cancellations will not work most of the time. The reason is that a
cancellation will open another TCP connection, which might be handled by
another pgbouncer process than the one that is processing the query that
should be cancelled. I talked about this problem in more detail at
the Uptime conference: https://www.youtube.com/watch?v=M585FfbboNA

This change adds support for pgbouncer peering. To enable peering two
things need to be added to the configuration of each pgbouncer:
1. A peer_id that's unique for each pgbouncer that is peered with
2. A new "peers" section that should get an entry for each of the other
   peers

When pgbouncer is configured like this, then whenever it receives a
cancellation request for an unknown session, it will forward it to the
peer that owns this session.

To configure three pgbouncers in peering mode using `so_reuseport`, you
can use these example configs.

pgbouncer1.ini
```ini
[databases]
postgres = host=localhost dbname=postgres

[pgbouncer]
peer_id=1
pool_mode=transaction
listen_addr=127.0.0.1
auth_type=trust
admin_users=jelte
auth_file=auth_file.conf
so_reuseport=1
unix_socket_dir=/tmp/pgbouncer1

[peers]
2 = host=/tmp/pgbouncer2
3 = host=/tmp/pgbouncer3
```

pgbouncer2.ini
```ini
[databases]
postgres = host=localhost dbname=postgres

[pgbouncer]
peer_id=2
pool_mode=transaction
listen_addr=127.0.0.1
auth_type=trust
admin_users=jelte
auth_file=auth_file.conf
so_reuseport=1
unix_socket_dir=/tmp/pgbouncer2

[peers]
1 = host=/tmp/pgbouncer1
3 = host=/tmp/pgbouncer3
```

pgbouncer3.ini
```ini
[databases]
postgres = host=localhost dbname=postgres

[pgbouncer]
peer_id=3
pool_mode=transaction
listen_addr=127.0.0.1
auth_type=trust
admin_users=jelte
auth_file=auth_file.conf
so_reuseport=1
unix_socket_dir=/tmp/pgbouncer3

[peers]
1 = host=/tmp/pgbouncer1
2 = host=/tmp/pgbouncer2
```

To then try out the cancellation forwarding. Simply run the following in
`psql`:
```sql
select pg_sleep(20);
```

And then press ctrl+c

Related to #345

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -212,9 +212,11 @@ If you need to specify multiple values, use a comma-separated list (e.g.
 
 Default: empty
 
-## peer_id
-The peer id used to identify this pgbouncer process in a group of pgbouncer
-processes that are peered together. When set to 0 pgbouncer peering is disabled
+### peer_id
+The peer id used to identify this PgBouncer process in a group of PgBouncer
+processes that are peered together. The `peer_id` value should be unique within
+a group of peered PgBouncer processes. When set to 0 pgbouncer peering is
+disabled. See the docs for the `[peers]` section for more information.
 
 Default: 0
 
@@ -861,6 +863,8 @@ might have implications for monitoring and metrics collection.
 To make sure query cancellations keep working, you should set up PgBouncer
 peering between the different PgBouncer processes. For details look at docs
 for the `peer_id` configuration option and the `peers` configuration section.
+There's also an example that uses peering and so_reuseport in the example
+section of these docs.
 
 Default: 0
 
@@ -1085,6 +1089,79 @@ database or default `pool_mode` is used.
 Configure a maximum for the user (i.e. all pools with the user will
 not have more than this many server connections).
 
+## Section [peers]
+
+The section `[peers]` defines the peers that PgBouncer can forward cancellation
+requests to and where those cancellation requests will be routed.
+
+PgBouncer processes can be peered together in a group by defining a `peer_id`
+value and a `[peers]` section in the configs of all the PgBouncer processes.
+These PgBouncer processes can then forward cancellations requests to the process
+that it originated from.  This is needed to make cancellations work when
+multiple PgBouncer processes (possibly on different servers) are behind the same
+TCP load balancer.  Cancellation requests are sent over different TCP
+connections than the query they are cancelling, so a TCP load balancer might
+send the cancellation request connection to a different process than the one
+that it was meant for.  By peering them these cancellation requests eventually
+end up at the right process. A more in-depth explanation is provided in this
+[recording of a conference talk][cancel-problem-video].
+
+[cancel-problem-video]: https://www.youtube.com/watch?v=M585FfbboNA
+
+The section contains key=value lines like
+
+    peer_id = connection string
+
+Where the key will be taken as a `peer_id` and the value as a connection string,
+consisting of key=value pairs of connection parameters, described below (similar
+to libpq, but the actual libpq is not used and the set of available features is
+different).  Example:
+
+    1 = host=host1.example.com
+    2 = host=/tmp/pgbouncer-2  port=5555
+
+Note: For peering to work, the `peer_id` of each PgBouncer process in the group
+must be unique within the peered group.  And the `[peers]` section should
+contain entries for each of those peer ids.  An example can be found in the
+examples section of these docs.  It **is** allowed, but not necessary, for the
+`[peers]` section to contain the `peer_id` of the PgBouncer that the config is
+for. Such an entry will be ignored, but it is allowed to config management easy.
+Because it allows using the exact same `[peers]` section for multiple
+configs.
+
+### host
+
+Host name or IP address to connect to.  Host names are resolved at connection
+time, the result is cached per `dns_max_ttl` parameter.  If DNS returns several
+results, they are used in a round-robin manner.  But in general it's not
+recommended to use a hostname that resolves to multiple IPs, because then the
+cancel request might still be forwarded to the wrong node and it would need to
+be forwarded again (which is only allowed up to three times).
+
+If the value begins with `/`, then a Unix socket in the file-system namespace is
+used.  If the value begins with `@`, then a Unix socket in the abstract
+namespace is used.
+
+Examples:
+
+	host=localhost
+	host=127.0.0.1
+	host=2001:0db8:85a3:0000:0000:8a2e:0370:7334
+	host=/var/run/pgbouncer-1
+
+### port
+
+Default: 6432
+
+### pool_size
+
+Set the maximum number of cancel requests that can be in flight to the peer at
+the same time.  It's quite normal for cancel requests to arrive in bursts, e.g.
+when the backing Postgres server slow or down.  So it's important for
+`pool_size` to not be so low that it cannot handle these bursts.
+
+If not set, the `default_pool_size` is used.
+
 
 ## Include directive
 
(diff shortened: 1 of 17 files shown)


src/objects.c
```diff
@@ -518,6 +572,35 @@ static PgPool *new_pool(PgDatabase *db, PgUser *user)
 	return pool;
 }
 
+
+/* create new pool object */
```
marcocitus, line 576: maybe add: for forwarding cancellations to other pgbouncers behind the same load balancer

src/objects.c
```diff
@@ -1323,57 +1423,100 @@ void accept_cancel_request(PgSocket *req)
...
+	} else {
+		PgDatabase *peer = find_peer(peer_id);
+		if (!peer) {
+			disconnect_client(req, false, "could not find peer to forward request to");
 			return;
 		}
+		log_debug("forwarding cancellation request to peer %d", peer_id);
 
-		disconnect_client(req, false, "cancel request for idle client");
+		pool = get_peer_pool(peer);
```
marcocitus, line 1507: could use a comment, e.g.
```c
/*
 * When using peering (multiple pgbouncers behind the same load balancer), we may receive 
 * cancellation messages that were intended for a peer via the load balancer. We propagate the
 * cancellation via the peer's pool, instead of the server pool.
 */
```

src/objects.c
```diff
@@ -874,7 +968,9 @@ void disconnect_server(PgSocket *server, bool notify, const char *reason, ...)
 	free_scram_state(&server->scram_state);
 
 	server->pool->db->connection_count--;
-	server->pool->user->connection_count--;
+	if (server->pool->user) {
```
marcocitus, line 971: should we add a comment somewhere that user is NULL in case of peer pools, because we only use them to propagate cancellations, which do not have a specific user.

marcocitus:
> When pgbouncer is configured like this, then whenever it receives a cancellation request for an unknown session, it will forward it to the peer that owns this session.

Worth noting that the peer ID is encoded into the first 16 bits of the cancellation key sent to the client (which is normally a random 64-bit number), and the last bit is used to indicate whether the cancellation was forwarded. This reduces entropy somewhat, but it is still a lot higher than for a regular postgres server which use the first 32-bit to encode the (quite predictable) PID.

src/objects.c
JelteF: mentioned this in the struct definition in `bouncer.h` and also in the function comment of `new_peer_pool`

JelteF:
I updated various comments to make the code and the reasoning clearer, including one about the reduced entropy.

marcocitus:
> [peers]
> 1 = host=/tmp/pgbouncer1
> 2 = host=/tmp/pgbouncer2

Is there any downside when users include the peer itself? (to keep config consistent)

src/loader.c
```diff
@@ -182,6 +183,78 @@ static bool set_autodb(const char *connstr)
...
+		goto fail;
+	}
+
+	/* tag the peer as alive */
+	peer->db_dead = false;
+
+	free(peer->host);
+	peer->host = host;
+	peer->port = port;
+	peer->pool_size = pool_size;
```
marcocitus, line 250: are there cases in which a pool_size > 1 is desirable? (should that be disallowed?)

src/objects.c
```diff
@@ -543,6 +607,37 @@ static PgPool *new_pool(PgDatabase *db, PgUser *user)
...
+	list_init(&pool->map_head);
+
+	pool->db = db;
+
+	statlist_init(&pool->new_server_list, "new_server_list");
+	statlist_init(&pool->waiting_cancel_req_list, "waiting_cancel_req_list");
+	statlist_init(&pool->active_cancel_req_list, "active_cancel_req_list");
+	statlist_init(&pool->active_cancel_server_list, "active_cancel_server_list");
+
+	/* keep pools in db/user order to make stats faster */
```
marcocitus, line 636: actually peer_id order

src/objects.c
```diff
@@ -1380,18 +1488,20 @@ void launch_new_connection(PgPool *pool, bool evict_if_needed)
 		}
 	}
 
-	max = user_max_connections(pool->user);
-	if (max > 0) {
-		/* try to evict unused connection first */
-		while (evict_if_needed && pool->user->connection_count >= max) {
-			if (!evict_user_connection(pool->user)) {
-				break;
+	if (pool->user) {
```
marcocitus, line 1491: this could use a comment, or an explicit boolean to make the code self-explanatory `if (!pool->is_peer)`

JelteF:
> Is there any downside when users include the peer itself? (to keep config consistent)

No, that works just fine. When receiving a cancellation, the code first checks if the cancellation its peer_id is its own peer_id. Only after that it checks if it knows the peer that it should forward the message to.

src/objects.c
JelteF: Removed this check. I added dedicated pool_size checking for peer pools now and if that succeeds it skips over this section.

src/loader.c
JelteF: Yes, I added user facing docs (which were partially missing) to explain why. If there are bursts of cancel requests you likely don't want to send those sequentially to the other process, so pool_size should almost always be larger than 1.

marcocitus:
I took another look at this, and it seems pretty solid. The only issue I ran into is that if the peer is unreachable then cancellation can take up to query_wait_timeout (2 minutes by default) to return, which seems a little unexpected, though also not specific to the peering problem.

doc/config.md
```diff
@@ -1202,6 +1288,40 @@ Example of a secure function for `auth_query`:
...
+    1 = host=/tmp/pgbouncer1
+    2 = host=/tmp/pgbouncer2
+
+    [pgbouncer]
+    listen_addr=127.0.0.1
+    auth_file=auth_file.conf
+    so_reuseport=1
+    unix_socket_dir=/tmp/pgbouncer1
+    peer_id=1
+
```
marcocitus, line 1307: I also had to add a pidfile to make the example work

achanda:
@JelteF I am not sure I understand why this is needed. In your example pgbouncer config, all the three pgbouncer instances talk to the same postgres host. In that case, the same postgres host will always receive the cancel request. I saw that your slides show two different postgres instances each with a pgbouncer in front, that setup makes sense to me. But as far as I can tell, this is not useful when a bunch of different pgbouncer instances all talk to the same postgres instance. Did I misunderstand this?

JelteF:
@achanda Yes, you misunderstood. This peering feature is useful if there's a load balancer in front of these different pgbouncer instances (or if you're using `so_reuseport=`, which is effectively turns the linux kernel into a load balancer). The reason is that the cancellation might end up at a different PgBouncer instance. This is a problem even when the backing postgres server is identical, because cancellation requests are not forwarded directly to Postgres. Instead the PgBouncer instance translates the cancellation token it receives from the client, to the specific client, and then it send the cancellation token for the server that's attached to that client to the server. So it sends a different cancellation token to the server than it received from the client. 

The client and server cancellation tokens cannot be the same, because when the client connects to PgBouncer, no server connection is assigned to it yet. And even if a connection would be assigned instantly, in transaction pooling mode the client will likely use another server connection for the next transaction, and the protocol does not allow updating the cancellation token. The cancellation token is static for the duration of the connection and is only sent during the start-up phase of the connection.

achanda:
Thanks for explaining, makes sense 👍

PR #675, opened 2022-01-25
Increase max password length

shamwow:
AWS produces passwords via the `aws rds generate-db-auth-token` command that are 1150 characters in length, which are incompatible with pgbouncer because the limit is set to 996.

This change bumps up that limit.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -147,11 +147,11 @@ extern int cf_sbuf_len;
 
 /*
  * Some cloud services use very long generated passwords, so give it
- * plenty of room.  Up to PostgreSQL 13, the server can handle
+ * plenty of room.  Note that, up to PostgreSQL 13, the server can handle
  * passwords up to 996 bytes, after that it's longer.  Also, libpq
  * maxes out around 1024, so going much higher is not straightforward.
  */
-#define MAX_PASSWORD	996
+#define MAX_PASSWORD	2048
 
 /*
  * AUTH_* symbols are used for both protocol handling and


include/bouncer.h
```diff
@@ -147,11 +147,11 @@ extern int cf_sbuf_len;
 
 /*
  * Some cloud services use very long generated passwords, so give it
- * plenty of room.  Up to PostgreSQL 13, the server can handle
+ * plenty of room.  Note that, up to PostgreSQL 13, the server can handle
  * passwords up to 996 bytes, after that it's longer.  Also, libpq
  * maxes out around 1024, so going much higher is not straightforward.
```
shamwow, line 152: @petere could you point to where this limit is defined in libpq? I wasn't able to find any evidence and in testing, a password of 1150 characters worked fine.

petere: The relevant libpq code has been whacked around a few times since I looked into this, so the situation might have changed. Which versions did you check?

shamwow: I was looking here: https://github.com/postgres/postgres/tree/master/src/backend/libpq. I see 1024 defined [here](https://github.com/postgres/postgres/blob/master/src/backend/libpq/auth.c#L2847) but that looks like its only for RADIUS auth.

I was able to change this to 2048 and everything works fine. So that got me curious where this limit is actually being enforced

josnyder-rh: So far as I can tell, the limit on password length was 995 bytes (1000 bytes - 4 byte packet length header - 1 byte null terminator) since 2002 (https://github.com/postgres/postgres/commit/3f63787cbfe0f1e837c92cd8ac35fd7ab811c18b#). It was then briefly changed to be unlimited (https://github.com/postgres/postgres/commit/67a472d71c98c3d2fa322a1b4013080b20720b98) and then changed again to be PG_MAX_AUTH_TOKEN_LENGTH (65536 bytes) (https://github.com/postgres/postgres/commit/9626325da5e8e23ff90091bc96535495d350f06e)

In the client code, I haven't been able to find an explicit limit. When reading passwords from a file, libpq will [enlarge its buffer in 128 byte chunks](https://github.com/postgres/postgres/blob/98e93a1fc93e9b54eb477d870ec744e9e1669f34/src/interfaces/libpq/fe-connect.c#L7080) up to INT_MAX (2147483647) . When I sent oversized passwords to my Postgres 9.6 server, the backend hung up on me.

Based on my limited reading of the source code, my guess is that the Postgres community intends to support passwords of length up to 65531 (65536 - 4 length bytes - 1 null byte) moving forward.

petere:
In the test suite, the test `test_password_client` tests with a maximum-length password. The test suite on Cirrus CI currently runs with a PostgreSQL version 11 server, which only supports up to length 996, which is why all the tests again this present patch fail. So we need to make changes to the test suite that it continues to pass against all supported PostgreSQL versions.

Another problem I have noticed is that if you set `MAX_PASSWORD` too high, you will get

    FATAL:  client sent partial pkt in startup phase

For me, the threshold appears to be around 4096 minus a few bytes. But I don't know what exactly determines this, so it might be lower in other environments. Maybe this is not a problem yet, but if we're going to keep increasing `MAX_PASSWORD` again and again, we will need to face this problem at some point.

PR #683, opened 2022-02-12
Close connection when check query errors

jfrconley:
Currently, pgbouncer will ignore errors returned by the check query. This is mostly ok for traditional Postgres, but creates issues when using pgbouncer in front of high availability pg compatible databases (ex: CockroachDB). Essentially, the pooler will hand out broken connections if the database is draining. This change simply closes SV_TESTED connections that receive errors, allowing for seamless failover. Additionally, newly logged in connections are now checked when check delay is set to zero.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -771,6 +771,13 @@ bool release_server(PgSocket *server)
 	case SV_LOGIN:
 		pool->last_login_failed = false;
 		pool->last_connect_failed = false;
+
+        /*
+         * Run check on fresh connection if check_delay is zero
+         */
+        if (cf_server_check_delay == 0 && *cf_server_check_query) {
+            newstate = SV_USED;
+        }
 		break;
 	default:
 		fatal("bad server state: %d", server->state);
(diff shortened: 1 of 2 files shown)


JelteF (changes requested):
I'm not entirely sure what issue this is fixing, could you describe the scenario a bit more clearly? I haven't used CockroachDB a lot, so I likely just need a bit more background to understand.

TBH, I'm not even sure I really understand the purpose of the `server_check_query` at all. From my look at the code of server_check_query, it's run after a previous query ends to check if the server is still working. I guess this is mostly to find out quickly if the connection got completely broken during the previous query, but even if it succeeds, it's still very possible that the connection breaks inbetween the `server_check_query` and the next query that's sent over the connection.

Also regarding this: 
>  Additionally, newly logged in connections are now checked when check delay is set to zero.

What is this necessary for?

JelteF:
Okay, with some more years of experience I now get the point of checking the error. That indeed seems sensible. It would be nice to have some tests though. Also this needs a rebase.

Issue #484, opened 2020-05-08
pgbouncer didn't even try auth_query to verify user password


bagafoot:
Hi all, I'm stuck and don't know where I'm doing wrong.
I'm try to configure pgbouncer to auth_type = hba and auth_user with auth_query but pgbouncer didn't send request to database. Pgbouncer and Postgres on the same server. What I'm missing

```
pgbouncer --version
PgBouncer 1.13.0
libevent 2.0.21-stable
adns: libc-2.17
tls: OpenSSL 1.0.2k-fips  26 Jan 2017
PostgreSQL 12.2 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-39), 64-bit
```
pgbouncer.ini

```
[databases]
test = dbname=test host=127.0.0.1 port=5432
test = auth_user=pgbouncer
[users]
test_app = pool_mode=session max_user_connections=3
[pgbouncer]
logfile = /var/log/pgbouncer/pgbouncer.log
pidfile = /var/run/pgbouncer/pgbouncer.pid
listen_addr = 127.0.0.1
listen_port = 6432
auth_type = hba
auth_file = /etc/pgbouncer/userlist.txt
auth_hba_file = /etc/pgbouncer/pg_hba.conf
auth_query = SELECT p_user, p_password FROM public.lookup($1)
admin_users = pgbouncer
stats_users = stats
server_reset_query = DISCARD ALL
server_reset_query_always = 1
max_client_conn = 100
default_pool_size = 20
log_connections = 1
log_disconnections = 1
log_pooler_errors = 1
log_stats = 1
verbose = 1
tcp_keepalive = 1
job_name = pgbouncer

```
userlist file
`"pgbouncer" "md5be5544d3807b54dd0637f2439ecb03b9"`

pgbouncer pg_hba.conf file
```
host    all     pgbouncer       127.0.0.1/32    trust
host    test    test_app        127.0.0.1/32    md5
```

postgres pg_hba.conf file
```
 # TYPE  DATABASE        USER            ADDRESS                 METHOD
 
 # "local" is for Unix domain socket connections only
 local   all             all                                    trust
 # IPv4 local connections:
 host    all             all             127.0.0.1/32            md5
```

lookup function 
```
test=# \df
                                 List of functions
 Schema |  Name  | Result data type |          Argument data types           | Type
--------+--------+------------------+----------------------------------------+------
[...]

Issue #314, opened 2018-08-06
pgbouncer crashes when user is equal to auth_user in database config


sidot3291:
For example:
```
test_database =  host=127.0.0.1 port=5432 auth_user=test_user user=test_user password=password
```

- If you attempt to login with a user defined in auth_file, the connection will succeed.
- If you attempt to login with a user not defined in auth_file and not returned by auth_query, you will receive a no such user error.
- If you attempt to login with a user not defined in auth_file but is returned by auth_query, pgbouncer crashes with:

> FATAL @src/objects.c:310 in function put_in_order(): put_in_order: found existing elem

This is true for both 'md5' and 'trust' auth types.

petere:
I cannot reproduce this. Can you send your complete setup, `pgbouncer.ini`, `userlist.txt`, `pg_authid` contents, etc.?

gclough:
Similar to #90 , and likely the same cause.

nikolaev-rd:
Not sure, but maybe this issue relates to #134

kenanrizvic:
> We have the same problem. Another problem, that we have: it seems like equal user from database used password from `auth_file` on PgBouncer, not from PostgreSQL database.

That sentence helped me to find problem in my case. We had in database section, lines with both user=XY and auth_user=XY  and password in userlist.txt  for XY user  and it worked ok untill someone try to login with unexisting user where pgbouncer failed with:

> 2019-09-03 13:49:03.155 29836 FATAL @src/objects.c:312 in function put_in_order(): put_in_order: found existing elem

Solution was to remove auth_user from database section.

petere:
The problem is that in PgBouncer internally, users with the same name can exist multiple times in different contexts:
- globally (from `auth_file`)
- PAM user
- attached to a database (via `auth_query`)
- as a "forced user" (`user=` entry in a database)

This is in principle correct, since for example users obtained via `auth_query` from different databases could have different passwords.

At the same time, however, PgBouncer maintains a global list of pools (global variable `pool_list`) that it thinks is uniquely identifiable by database name and user name. You can already see how this would be a problem.

In the particular case, where you have

    mydb = ... user=foo auth_user=foo

[...]

PR #706, opened 2022-04-14
Fix shadow auth for users defined under [users] section

Justin-Kwan:
**Description**

When a user is not defined within `userlist.txt` and exists as a user in the Postgres server, PgBouncer should send an auth query to retrieve the user's md5 password hash in order to log the user in.

However, as outlined in issue: https://github.com/pgbouncer/pgbouncer/issues/484, if a user's password is not defined in `userlist.txt` and the user has overridden configurations under the `[users]` section in PgBouncer's ini file, the user is unable to login to the Postgres server through PgBouncer.

**Root Cause**

This happens because any users defined under `[users]` are parsed and added as new users in `loader.c`'s `parse_user`. As result, when the real and live user connects, PgBouncer thinks they already exist (not NULL) and skips the shadow authorization process where the user's md5 password hash is queried from `pg_shadow`.

**Fix**

This PR introduces a new flag to track whether a user was created from ini file parsing or from a real live login. Then, we determine whether a new user logging in requires shadow authorization.

This fix should now allow PgBouncer operators to use the `max_user_connections` per user feature.

cc @viggy28

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -326,6 +326,7 @@ struct PgUser {
 	uint8_t scram_ServerKey[32];
 	bool has_scram_keys;		/* true if the above two are valid */
 	bool mock_auth;			/* not a real user, only for mock auth */
+	bool is_preconfigured;		/* true if user is created from pre-configuration parsing */
 	int pool_mode;
 	int max_user_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
--- a/src/client.c
+++ b/src/client.c
@@ -206,7 +206,7 @@ static bool finish_set_pool(PgSocket *client, bool takeover)
 		if (client->db->forced_user)
 			pool_user = client->db->forced_user;
 		else
-			pool_user = client->login_user;
+			pool_user = find_original_user(client->login_user);
 
 		client->pool = get_pool(client->db, pool_user);
 		if (!client->pool) {
--- a/src/client.c
+++ b/src/client.c
@@ -343,13 +343,12 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 		}
 	} else {
 		client->login_user = find_user(username);
-		if (!client->login_user) {
+		if (!client->login_user || !user_is_authenticated(client->login_user)) {
 			/*
-			 * If the login user specified by the client
-			 * does not exist, check if an auth_user is
-			 * set and if so send off an auth_query.  If
-			 * no auth_user is set for the db, see if the
-			 * global auth_user is set and use that.
+			 * If the login user specified by the client does not exist or only exists as
+			 * a pre-configuration, check if an auth_user is set and if so send off an
+			 * auth_query. If no auth_user is set for the db, see if the global auth_user
+			 * is set and use that.
 			 */
 			if (!client->db->auth_user && cf_auth_user) {
 				client->db->auth_user = find_user(cf_auth_user);
--- a/src/loader.c
+++ b/src/loader.c
@@ -407,11 +407,13 @@ bool parse_user(void *base, const char *name, const char *connstr)
 
 	user = find_user(name);
 	if (!user) {
+		/* represents a user pre-configuration, not a connected logged-in user */
 		user = add_user(name, "");
 		if (!user) {
 			log_error("cannot create user, no memory?");
 			goto fail;
 		}
+		user->is_preconfigured = true;
 	}
 
 	user->pool_mode = pool_mode;
--- a/src/objects.c
+++ b/src/objects.c
@@ -387,6 +387,36 @@ PgUser *add_user(const char *name, const char *passwd)
 	return user;
 }
 
+/* find original or preconfigured user given the login user */
+PgUser *find_original_user(PgUser *login_user)
+{
+	/*
+	 * The original user will either be the current login user, a preconfigured user via
+	 * ini file [users] section, a preconfigured user via global auth_file, a PAM user,
+	 * a user attached to a database via auth_query, a forced user with user= entry via
+	 * ini file [databases] section, or not exist yet if the user has never connected.
+	 */
+	PgUser *original_user = find_user(login_user->name);
+	char *tmp_passwd;
+
+	/* add new login user to reload their configurations at runtime */
+	if (original_user == NULL) {
+		put_in_order(&login_user->head, &user_list, cmp_user);
+		aatree_insert(&user_tree, (uintptr_t)login_user->name, &login_user->tree_node);
+		return login_user;
+	}
+
+	/* original user is no longer preconfigured and won't send auth queries on future login */
+	if (original_user->is_preconfigured)
+		original_user->is_preconfigured = false;
+
+	tmp_passwd = strdup(login_user->passwd);
+	safe_strcpy(original_user->passwd, tmp_passwd, sizeof(original_user->passwd));
+	free(tmp_passwd);
+
+	return original_user;
+}
+
 /* add or update db users */
 PgUser *add_db_user(PgDatabase *db, const char *name, const char *passwd)
 {
--- a/src/server.c
+++ b/src/server.c
@@ -243,6 +243,14 @@ int user_max_connections(PgUser *user)
 	}
 }
 
+bool user_is_authenticated(PgUser *user)
+{
+	/* authenticated users cannot be in a non-logged in or preconfigured state */
+	return cf_auth_type == AUTH_TRUST ||
+		   (cf_auth_user != NULL && strcmp(cf_auth_user, user->name) == 0) ||
+		   !user->is_preconfigured;
+}
+
 /* process packets on logged in connection */
 static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 {
--- a/test/test.sh
+++ b/test/test.sh
@@ -185,6 +185,8 @@ psql -X -p $PG_PORT -d p0 -c "select * from pg_user" | grep pswcheck > /dev/null
 	psql -X -o /dev/null -p $PG_PORT -c "create user pswcheck with superuser createdb password 'pgbouncer-check';" p0 || exit 1
 	psql -X -o /dev/null -p $PG_PORT -c "create user someuser with password 'anypasswd';" p0 || exit 1
 	psql -X -o /dev/null -p $PG_PORT -c "create user maxedout;" p0 || exit 1
+	psql -X -o /dev/null -p $PG_PORT -c "create user shadowuser1 with password 'bar';" p0 || exit 1
+	psql -X -o /dev/null -p $PG_PORT -c "create user shadowuser2 with password 'foo';" p0 || exit 1
 	psql -X -o /dev/null -p $PG_PORT -c "create user longpass with password '$long_password';" p0 || exit 1
 	if $pg_supports_scram; then
 		psql -X -o /dev/null -p $PG_PORT -c "set password_encryption = 'md5'; create user muser1 password 'foo';" p0 || exit 1
@@ -931,6 +933,48 @@ test_password_server() {
 	return 0
 }
 
+# test password authentication from PgBouncer to PostgreSQL server using
+# auth_query to retrieve user's shadow password from pg_shadow
+test_shadow_password_server_login() {
+	$have_getpeereid || return 77
+
+	admin "set auth_type='md5'"
+	admin "set auth_user='pswcheck'"
+
+	# plain-text password of auth_user in userlist.txt
+	curuser=`psql -X -d "dbname=authdb user=pswcheck password=pgbouncer-check" -tAq -c "select current_user;"`
+	echo "curuser=$curuser"
+	test "$curuser" = "pswcheck" || return 1
+
+	# user with correct password from PostgreSQL server
+	curuser=`psql -X -d "dbname=authdb user=shadowuser1 password=bar" -tAq -c "select current_user;"`
+	echo "curuser=$curuser"
+	test "$curuser" = "shadowuser1" || return 1
+	# user with wrong password from PostgreSQL server
+	curuser=`psql -X -d "dbname=authdb user=shadowuser1 password=badpasswd" -tAq -c "select current_user;"`
+	echo "curuser=$curuser"
+	test "$curuser" = "" || return 1
+
+	# user defined in ini [users] section with correct password from PostgreSQL server
+	curuser=`psql -X -d "dbname=authdb user=shadowuser2 password=foo" -tAq -c "select current_user;"`
+	echo "curuser=$curuser"
+	test "$curuser" = "shadowuser2" || return 1
+	# user defined in ini [users] section with wrong password from PostgreSQL server
+	curuser2=`psql -X -d "dbname=authdb user=shadowuser2 password=badpasswd" -tAq -c "select current_user;"`
+	echo "curuser2=$curuser2"
+	test "$curuser2" = "" || return 1
+
+	# auth_user defined in ini [users] section with correct password from PostgreSQL server
+	admin "set auth_user='shadowuser2'"
+	curuser=`psql -X -d "dbname=authdb user=shadowuser2 password=foo" -tAq -c "select current_user;"`
+	echo "curuser=$curuser"
+	test "$curuser" = "shadowuser2" || return 1
+
+	admin "set auth_type='trust'"
+
+	return 0
+}
+
 # test plain-text password authentication from client to PgBouncer
 test_password_client() {
 	$have_getpeereid || return 77
@@ -1445,6 +1489,7 @@ test_server_lifetime
 test_server_idle_timeout
 test_query_timeout
 test_idle_transaction_timeout
+test_shadow_password_server_login
 test_server_connect_timeout_establish
 test_server_connect_timeout_reject
 test_server_check_delay
--- a/include/objects.h
+++ b/include/objects.h
@@ -47,6 +47,7 @@ void disconnect_client(PgSocket *client, bool notify, const char *reason, ...) _
 
 PgDatabase * add_database(const char *name) _MUSTCHECK;
 PgDatabase *register_auto_database(const char *name);
+PgUser * find_original_user(PgUser *login_user) _MUSTCHECK;
 PgUser * add_user(const char *name, const char *passwd) _MUSTCHECK;
 PgUser * add_db_user(PgDatabase *db, const char *name, const char *passwd) _MUSTCHECK;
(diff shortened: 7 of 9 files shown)


test/test.sh
```diff
@@ -931,6 +933,40 @@ test_password_server() {
 	return 0
 }
 
+# test password authentication from PgBouncer to PostgreSQL server using
+# auth_query to retrieve user's shadow password from pg_shadow
+test_shadow_password_server_login() {
```
Justin-Kwan, line 938: These test cases are reproducible examples of bug: https://github.com/pgbouncer/pgbouncer/issues/484

src/loader.c
```diff
@@ -407,11 +407,13 @@ bool parse_user(void *base, const char *name, const char *connstr)
 
 	user = find_user(name);
 	if (!user) {
+		/* represents a user pre-configuration, not a connected logged-in user */
 		user = add_user(name, "");
 		if (!user) {
 			log_error("cannot create user, no memory?");
 			goto fail;
 		}
+		user->is_preconfigured = true;
```
Justin-Kwan, line 416: We mark a user as preconfigured if they are created within the parser, and not from a real login.

viggy28 (approved):

src/objects.c
```diff
@@ -387,6 +387,42 @@ PgUser *add_user(const char *name, const char *passwd)
...
+	/* original user is no longer preconfigured and won't send auth queries on future login */
+	if (original_user->is_preconfigured)
+		original_user->is_preconfigured = false;
+
+	char *tmp_passwd = strdup(login_user->passwd);
+	safe_strcpy(original_user->passwd, tmp_passwd, sizeof(original_user->passwd));
+	free(tmp_passwd);
+
+	if (login_user != original_user) {
+		statlist_remove(&user_list, &login_user->head);
```
Justin-Kwan, line 418: We deallocate any new login user that does not match the initial user reference parsed from `parse_user`. We always keep the original reference in the global `user_list`.

src/objects.c
```diff
@@ -387,6 +387,42 @@ PgUser *add_user(const char *name, const char *passwd)
...
+		aatree_remove(&user_tree, (uintptr_t)login_user->name);
+		slab_free(user_cache, login_user);
```
Justin-Kwan, line 420: We deallocate any new login user that does not match the initial user reference parsed from parse_user. We always keep the original reference in the global user_list.

src/objects.c
```diff
@@ -387,6 +387,42 @@ PgUser *add_user(const char *name, const char *passwd)
...
+	 * The original user will either be the current login user, a preconfigured user via
+	 * ini file [users] section, a preconfigured user via global auth_file, a PAM user,
+	 * a user attached to a database via auth_query, a forced user via user= entry via
+	 * ini file [databases] section, or not exist yet if the user has never connected.
+	 */
+	PgUser *login_user = *user;
+	PgUser *original_user = find_user(login_user->name);
+
+	/* add new login user to reload their configurations at runtime */
+	if (original_user == NULL) {
```
Justin-Kwan, line 403: If the new login user is not defined in ini configuration `[users]`/`[databases]` section, global `auth_file`, etc. then we store it in the global `user_list` to ensure that any admin `RELOAD`s can access the user to update it's configuration fields at runtime.

src/objects.c
```diff
@@ -387,6 +387,36 @@ PgUser *add_user(const char *name, const char *passwd)
...
+		aatree_insert(&user_tree, (uintptr_t)login_user->name, &login_user->tree_node);
+		return login_user;
+	}
+
+	/* original user is no longer preconfigured and won't send auth queries on future login */
+	if (original_user->is_preconfigured)
+		original_user->is_preconfigured = false;
+
+	tmp_passwd = strdup(login_user->passwd);
+	safe_strcpy(original_user->passwd, tmp_passwd, sizeof(original_user->passwd));
```
Justin-Kwan, line 414: Copy over password of new login user to account for a single user that may have different passwords for multiple databases.

JelteF: So you are copying every time in case a user has a different password in different databases? I think even with this code that could lead to problems, because the `original_user` is shared between clients of different databases. So maybe we should just document that a user should not have different passwords for different databases when using `auth_query`/`auth_user`.

src/objects.c
```diff
@@ -387,6 +387,36 @@ PgUser *add_user(const char *name, const char *passwd)
 	return user;
 }
 
+/* find original or preconfigured user given the login user */
+PgUser *find_original_user(PgUser *login_user)
```
JelteF, line 391: This function name isn't really descriptive of what this function is doing, especially the `put_in_order` call.

JelteF (commented):
The new tests that you added also pass with the following simpler version of `find_original_user`. Are the rest of things that you added to this function really necessary?
```c
/* find original or preconfigured user given the login user */
PgUser *find_original_user(PgUser *login_user)
{
	/*
	 * The original user will either be the current login user, a preconfigured user via
	 * ini file [users] section, a preconfigured user via global auth_file, a PAM user,
	 * a user attached to a database via auth_query, a forced user with user= entry via
	 * ini file [databases] section, or not exist yet if the user has never connected.
	 */
	PgUser *original_user = find_user(login_user->name);

	/* add new login user to reload their configurations at runtime */
	if (original_user == NULL) {
		return login_user;
	}

	/* original user is no longer preconfigured and won't send auth queries on future login */
	if (original_user->is_preconfigured) {
		original_user->is_preconfigured = false;
		if (original_user->passwd[0] == '\0') {
			safe_strcpy(original_user->passwd, login_user->passwd, sizeof(original_user->passwd));
		}
	}

	return original_user;
}
```

test/test.sh
```diff
@@ -931,6 +933,48 @@ test_password_server() {
...
+	$have_getpeereid || return 77
+
+	admin "set auth_type='md5'"
```
JelteF, line 941: Do you think it makes sense to include the same tests also for scram auth?

Justin-Kwan:
Hi @JelteF! Thanks for taking a look at this PR. This PR is actually very outdated, I'm going to repush everything.

src/objects.c
Justin-Kwan: Updated the PR. `find_original_user` no longer exists. To support a user logging into different databases with different passwords (usually when databases have different hosts), each `PgDatabase*` now stores a list of user passwords.

```
struct PgUserPassword {
	struct AANode tree_node;	/* used to attach user password to tree */
	char username[MAX_USERNAME];
	char passwd[MAX_PASSWORD];
};
```

Semantically, this should be clearer since it models reality, where each database may have different passwords per user. `PgUser*` still retains a password field to store a password from `auth_file` which will be used if it exists by default.

Justin-Kwan:
Some more details around refactoring work done here:

**Context:**

In the OSS PgBouncer, PgUser* may exist in multiple contexts: (see OSS discussion) https://github.com/pgbouncer/pgbouncer/issues/314#issuecomment-1042807960

- If a user is defined in userlist.txt, then they were stored in global user_tree
- Any user not defined in userlist.txt (requiring an auth_query to PG) would be stored locally under PgDatabase->user_tree
- From this, you can imagine that a single user connecting to multiple databases would be duplicated in memory
- This is confusing to understand, because semantically, one user exists across all databases in a single Postgres cluster (host server)

**Issue:**

- In order to **correctly** support global PgUser configuration settings (`max_user_connections`, etc.), all users now reside globally in `user_tree` to simplify centrally storing/enforcing configuration values against
- PgBouncer supports connecting to databases with different hostnames/ports (ex. connection pooling and proxying queries to different cluster endpoints)
- Each user (uniquely denoted by name) may have a different password per each different database (on separate clusters)
Currently, we only store the single password defined in `auth_file`, or the latest password retrieved from the last database that the user connected to, which causes the user to login using a password for a different database (failed login)

**This PR:**

- Refactors PgBouncer to handle multiple passwords per user
- `PgDatabase->user_passwds` contains all passwords for users to log into it (user with multiple passwords rather than multiple user instances)
- This approach was chosen because in real life, each database keeps track of all passwords for users all users
It is cleaner to store user to password pairs per database because each user is unique by name (easily identifiable), storing the other way around (database to password pairs in PgUser would be more convoluted because database names aren't unique, and would require appending port, etc)
- When the database is periodically deallocated (in kill_database), all passwords will be cleanly freed as well, as opposed to having to make an expensive lookup through all users to find which have passwords for the freed database, and removing those passwords

src/admin.c
```diff
@@ -345,12 +347,12 @@ static bool show_one_fd(PgSocket *admin, PgSocket *sk)
...
 
-	if (sk->pool && sk->pool->db->auth_user && sk->login_user && !find_user(sk->login_user->name))
-		password = sk->login_user->passwd;
+	is_auth_user_present = sk->pool && sk->pool->db->auth_user && sk->login_user && !find_user(sk->login_user->name);
+	is_pam_auth_used = cf_auth_type == AUTH_PAM && !find_user(sk->login_user->name);
 
 	/* PAM requires passwords as well since they are not stored externally */
-	if (cf_auth_type == AUTH_PAM && !find_user(sk->login_user->name))
-		password = sk->login_user->passwd;
+	if (is_auth_user_present || is_pam_auth_used)
```
Justin-Kwan, line 354: This is done to prevent unnecessary extra call to `find_user`.

src/client.c
```diff
@@ -339,13 +343,12 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 		}
 	} else {
 		client->login_user = find_user(username);
-		if (!client->login_user) {
+		if (!client->login_user || user_requires_auth_query(client->login_user)) {
```
Justin-Kwan, line 346: Existence of `login_user` does not imply authentication of the user, as the user could exist due to being parsed from `[users]` section, without a password in `auth_file`. This motivates the use of method `user_requires_auth_query`.

src/client.c
```diff
@@ -394,7 +398,9 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
 		}
 		break;
 	case 'D':	/* DataRow */
-		memset(&user, 0, sizeof(user));
+		memset(&real_user, 0, sizeof(real_user));
```
Justin-Kwan, line 401: Code cleanup to use two variables (simpler) rather than partial fields of `PgUser*` struct.

src/client.c
```diff
@@ -427,23 +433,29 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
...
 		}
-		if (sizeof(user.passwd)  - 1 < length)
-			length = sizeof(user.passwd) - 1;
-		memcpy(user.passwd, password, length);
+		if (sizeof(real_passwd) - 1 < length)
+			length = sizeof(real_passwd) - 1;
+		memcpy(real_passwd, passwd, length);
 
-		client->login_user = add_db_user(client->db, user.name, user.passwd);
+		client->login_user = add_user(real_user, "");
```
Justin-Kwan, line 448: We can leave `PgUser*->passwd` empty here because we've just received the user's password from Postgres `pg_shadow`. This means that the user's password must not be in `auth_file`.

src/client.c
```diff
@@ -584,6 +596,7 @@ static bool scram_client_first(PgSocket *client, uint32_t datalen, const uint8_t
 	char *input;
 	int res;
 	PgUser *user = client->login_user;
+	const char *real_passwd = user_password(user, client_database(client));
```
Justin-Kwan, line 599: Lookup user's password from the database.

src/janitor.c
```diff
@@ -703,6 +703,7 @@ void kill_database(PgDatabase *db)
 			kill_pool(pool);
 	}
 
+	aatree_destroy(&db->user_passwds);
```
Justin-Kwan, line 706: Free user passwords associated to the database being killed.

src/loader.c
```diff
@@ -584,7 +612,8 @@ bool load_auth_file(const char *fn)
 		*p++ = 0; /* tag password end */
 
 		/* send them away */
-		unquote_add_user(user, password);
+		user = unquote_add_user(username, password);
+		user->from_auth_file = true;
```
Justin-Kwan, line 616: Flag that user was from `auth_file` to know whether to send off `auth_query` upon login.

src/server.c
```diff
@@ -243,6 +243,69 @@ int user_max_connections(PgUser *user)
...
+bool user_requires_auth_query(PgUser *user)
+{
+	/* users defined in auth_file with password do not require an auth_query */
+	return cf_auth_type != AUTH_TRUST && !user->from_auth_file;
+}
+
+void user_passwd_free(struct AANode *node, void *arg)
+{
+	PgUserPassword *user_passwd = container_of(node, PgUserPassword, tree_node);
+	free(user_passwd);
```
Justin-Kwan, line 255: `PgUserPassword` allocation/deallocation could be handled by slab allocation pooler if you think that's necessary.

test/test.sh
```diff
@@ -768,6 +770,22 @@ test_enable_disable() {
 	return 0
 }
 
+test_database_kill() {
```
Justin-Kwan, line 773: This test ensures that `PgUser*` still globally exists even if a database (which contains one of its passwords) is freed.

test/test.sh
```diff
@@ -986,6 +986,28 @@ test_shadow_password_server_login() {
 	echo "curuser=$curuser"
 	test "$curuser" = "shadowuser2" || return 1
 
+	# user defined in ini [users] section with good recently changed password from PostgreSQL server
```
Justin-Kwan, line 989: More test cases to ensure that users can still login with new password when it's changed at Postgres.

src/objects.c
Justin-Kwan: `find_original_user` no longer exists now.

src/objects.c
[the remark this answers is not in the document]
Justin-Kwan: `find_original_user` no longer exists now.

src/objects.c
```diff
@@ -387,6 +387,36 @@ PgUser *add_user(const char *name, const char *passwd)
...
+		put_in_order(&login_user->head, &user_list, cmp_user);
+		aatree_insert(&user_tree, (uintptr_t)login_user->name, &login_user->tree_node);
+		return login_user;
+	}
+
+	/* original user is no longer preconfigured and won't send auth queries on future login */
+	if (original_user->is_preconfigured)
+		original_user->is_preconfigured = false;
+
+	tmp_passwd = strdup(login_user->passwd);
```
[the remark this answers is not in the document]
Justin-Kwan: `find_original_user` no longer exists now.

include/bouncer.h
```diff
@@ -335,6 +336,7 @@ struct PgUser {
 struct PgDatabase {
 	struct List head;
 	char name[MAX_DBNAME];	/* db name for clients */
+	struct AATree user_passwds;	/* user passwords for this database from auth_query */
```
Justin-Kwan, line 339: A hashtable to store user passwords is more ideal but is more difficult (would have to pull in glibc).

src/client.c
```diff
@@ -35,38 +35,39 @@ static const char *hdr2hex(const struct MBuf *data, char *buf, unsigned buflen)
 	return bin2hex(bin, dlen, buf, buflen);
 }
 
-static bool check_client_passwd(PgSocket *client, const char *passwd)
+static bool check_client_passwd(PgSocket *client, const char *provided_passwd)
```
Justin-Kwan, line 38: More cleanup to easily discern between the user's real password (from Postgres) vs client provided password.

src/server.c
```diff
@@ -249,6 +249,63 @@ bool user_requires_auth_query(PgUser *user)
...
+
+char *user_password(const PgUser *user, const PgDatabase *db)
+{
+	PgUserPassword *user_passwd = NULL;
+	struct AANode *node;
+
+	if (db == NULL || user == db->forced_user || user->from_auth_file)
+		return (char *)user->passwd;
+
+	node = aatree_search((struct AATree *)&db->user_passwds, (uintptr_t)user->name);
```
Justin-Kwan, line 275: Cast away from const* here. Alternatively, can add `const*` modifier to signature `aatree_search`.

JelteF:
Thanks for sharing the updated code, I'll try to look at it soonish. If you rebase with master the Alpine tests should hopefully start passing again. And the valgrind tests are currently failing, but seemingly a pretty unrelated test, so maybe it's just a flaky test.

goksgie:
If I am not misinterpreting the final version of this PR, it does not contain the proposed design of maintaining user password in databases. Instead, it seems to be storing users in `user_tree` and update when new login is attempted. Is that the intended behavior? 

Additionally, I do not think `user_tree` is getting cleared unlike databases. Thus, question becomes, what is the resource implications of this design, especially for those systems that rather have limited resources?

Justin-Kwan:
Please take a look at this commit in our official fork of PgBouncer, this implements the newly proposed design (which has been running smoothly under the load of roughly 20K QPS in production): https://github.com/cloudflare/cf-pgbouncer/commit/999133ef3f3a328acca81c1119aee01d9c2bbd6e

Hi @goksgie! In our workloads, tenants typically stay connected for long periods, and so the number of users should not be frequently changing or linearly increasing. Struct allocation should be fairly cheap even for a few thousand `PgUser` structs. However, if you're still concerned about active heap usage given **many many** users that become stale frequently (IMO an uncommon usage pattern), an eviction callback could be registered with libevent to periodically purge all users without custom configurations.

I won't have any cycles to work on this PR as I am back in school now.

src/objects.c
JelteF: This change seems to somehow have been reverted on your branch. Possibly due to a rebase.

JelteF (changes requested):
Sorry for taking way too long to look at this again. But looking at the current change it seems that it has reverted back to one of the earliest versions of the PR because `find_original_user` is back (even though you mentioned that it is removed now).

Other than that:
1. This needs a rebase and should use the new testing infrastructure
2. We should probably have some cleanup mechanism for the users like @goksgie described

But mostly it seems problematic that this PR is containing the old code again, which means I cannot review the latest version.

veshant:
> Sorry for taking way too long to look at this again. But looking at the current change it seems that it has reverted back to one of the earliest versions of the PR because `find_original_user` is back (even though you mentioned that it is removed now).
> 
> Other than that:
> 
> 1. This needs a rebase and should use the new testing infrastructure
> 2. We should probably have some cleanup mechanism for the users like @goksgie described
> 
> But mostly it seems problematic that this PR is containing the old code again, which means I cannot review the latest version.

It might be worth considering to merge the [cf-pgbouncer](https://github.com/cloudflare/cf-pgbouncer) fork. Probably requires some heavy refactoring but they have some useful fixes and features including the users section.

PR #707, opened 2022-04-16
Configure and enforce max connections per user at runtime

Justin-Kwan:
**Description**

This PR introduces a new command to enforce max user connections at runtime. In addition to existing `max_user_connections` per user on `pgbouncer.ini` file reload, PgBouncer operators can now execute `SET USER <user> = 'max_user_connections=<new limit> pool_mode=<session|transaction|statement>'` to update a user's max connections across all pools that the user is in.

When the `pgbouncer.ini` file is reloaded, or the above command is submitted, PgBouncer will now drop the user's connections as a reaction to these new values. PgBouncer will prioritize killing any idle and unused connections, and then kill the oldest active connections across all pools owned by the user to achieve the configured connection limit.

In addition, this PR also updates the `SHOW USERS;` command to display the maximum configured connections per user to increase visibility when throttling users.

In a production multi-tenant cluster, we have seen the need for PgBouncer to enforce maximum connections per user across all pools. When our upstream services (PG users) acquires too many connections while executing expensive and/or repetitive queries, this often puts strain on our Postgres primary server and starves CPU/Disk IO resources at the server. To remediate this, currently we manually constrain the user to maximum number of Postgres connections at server side via `ALTER USER "some_bad-user" WITH CONNECTION LIMIT 123;`. We've seen a clear correlation between the connection pool size (concurrency) per tenant denoted in green and database resource load denoted in yellow:

production connections to load correlation (1)

Having this option in PgBouncer would be very convenient because:

1. Postgres does not actively kill existing connections that surpass the new limit (which means load does not immediately reduce)
2. This opens the option for operators to build custom scripts that can automate per user connection throttling in PgBouncer via the `SET USER ...` command

**Note**

This PR follows from https://github.com/pgbouncer/pgbouncer/pull/706 and has some overlapping commits. The previous PR should merge before this one.

cc @viggy28

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -387,6 +387,36 @@ PgUser *add_user(const char *name, const char *passwd)
 	return user;
 }
 
+/* find original or preconfigured user given the login user */
+PgUser *find_original_user(PgUser *login_user)
+{
+	/*
+	 * The original user will either be the current login user, a preconfigured user via
+	 * ini file [users] section, a preconfigured user via global auth_file, a PAM user,
+	 * a user attached to a database via auth_query, a forced user with user= entry via
+	 * ini file [databases] section, or not exist yet if the user has never connected.
+	 */
+	PgUser *original_user = find_user(login_user->name);
+	char *tmp_passwd;
+
+	/* add new login user to reload their configurations at runtime */
+	if (original_user == NULL) {
+		put_in_order(&login_user->head, &user_list, cmp_user);
+		aatree_insert(&user_tree, (uintptr_t)login_user->name, &login_user->tree_node);
+		return login_user;
+	}
+
+	/* original user is no longer preconfigured and won't send auth queries on future login */
+	if (original_user->is_preconfigured)
+		original_user->is_preconfigured = false;
+
+	tmp_passwd = strdup(login_user->passwd);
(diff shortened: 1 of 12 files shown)


src/objects.c
```diff
@@ -1207,6 +1232,56 @@ bool evict_user_connection(PgUser *user)
...
+	}
+}
+
+void notify_user_cf_updated(PgUser *user)
+{
+	if (user->connection_count <= user_max_connections(user) || pgb_event_base == NULL)
+		return;
+
+	struct event ev_connection_limit_enforcement;
+	event_assign(&ev_connection_limit_enforcement, pgb_event_base, -1, EV_TIMEOUT, enforce_user_connection_limit, user);
```
Justin-Kwan, line 1281: We queue a new event callback to drop all of the user's connections that surpass their newly configured limit, upon `RELOAD` or `SET USER ...`.

viggy28 (approved):

JelteF:
With the amount of merge conflicts and @Justin-Kwan moving on to working on other projects I'm going to close this PR. I think it's a sensible feature. If someone wants implement something similar based on the user management refactoring that was done in #1052, then that would be a welcome addition.

PR #717, opened 2022-05-12
Handle race condition between cancel requests and server reuse

JelteF:
Cancel requests happen out of band with regular queries. Because
PgBouncer reuses server connections across clients its possible to get
race conditions here if you're not careful.

Without this patch there were two possibilities for a race condition
where a cancel request for client A could cancel a query that was sent
by client B. These race conditions happened when:
1. A cancel request on client A was forwarded to the server only after
   the server was reused by another client.
2. A cancel request for client A was already forwarded to the server,
   but not yet handled by the server and then the server was reused for
   client B.

The first race conditions was especially likely if connection
establishment takes a while, for example because of TLS roundtrips.
Because a new connection needs to be established for every forwarded
cancellation request. While establishing this connection the query
running on the server might finish by itself already, and then the
server could be reused by another already waiting client.

To resolve these race conditions the following two fixes are implemented:
1. We now only forward the cancel request if the server connection over
   which the query was sent hasn't become idle or disconnected in the
   meantime. This is done by doubly-linking cancel requests and the
   original server connection that was used for the query.
2. Once a cancel request is forwarded to the server, we do not allow reuse of
   the server until the cancel request completes. This required tracking
   the response of the server for cancel requests too, instead of
   closing the connection right after forwarding the cancel.

Fixes #544

In passing this also adds more comprehensive comments to the different
`StatList`s in the `PgPool` struct. Because understanding how the state
machine worked was the part of this PR that took me the longest.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -49,21 +49,24 @@ extern struct event_base *pgb_event_base;
 
 /* each state corresponds to a list */
 enum SocketState {
-	CL_FREE,		/* free_client_list */
+	CL_FREE,			/* free_client_list */
 	CL_JUSTFREE,		/* justfree_client_list */
-	CL_LOGIN,		/* login_client_list */
-	CL_WAITING,		/* pool->waiting_client_list */
+	CL_LOGIN,			/* login_client_list */
+	CL_WAITING,			/* pool->waiting_client_list */
 	CL_WAITING_LOGIN,	/*   - but return to CL_LOGIN instead of CL_ACTIVE */
-	CL_ACTIVE,		/* pool->active_client_list */
-	CL_CANCEL,		/* pool->cancel_req_list */
+	CL_ACTIVE,			/* pool->active_client_list */
+	CL_WAITING_CANCEL,	/* pool->waiting_cancel_req_list */
+	CL_ACTIVE_CANCEL,	/* pool->active_cancel_req_list */
 
-	SV_FREE,		/* free_server_list */
+	SV_FREE,			/* free_server_list */
 	SV_JUSTFREE,		/* justfree_server_list */
-	SV_LOGIN,		/* pool->new_server_list */
-	SV_IDLE,		/* pool->idle_server_list */
-	SV_ACTIVE,		/* pool->active_server_list */
-	SV_USED,		/* pool->used_server_list */
-	SV_TESTED		/* pool->tested_server_list */
+	SV_LOGIN,			/* pool->new_server_list */
+	SV_WAIT_CANCELS,	/* pool->wait_cancels_server_list */
+	SV_IDLE,			/* pool->idle_server_list */
+	SV_ACTIVE,			/* pool->active_server_list */
+	SV_ACTIVE_CANCEL,	/* pool->active_cancel_server_list */
+	SV_USED,			/* pool->used_server_list */
+	SV_TESTED			/* pool->tested_server_list */
 };
 
 enum PauseMode {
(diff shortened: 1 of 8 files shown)


JelteF:
PgBouncer also generates unique cancellation tokens per client. So, that alone won't protect you against the same issue in Odyssey.

To clarify the issue: These unique cancellation tokens, when received by the connection pooler, are handled by by sending a cancellation token to the actual Postgres server. This cancellation token should be the one that matches the postgres connection on which the query of the client is currently being sent. The problem is that while sending this cancellation token to Postgres, the query on that connection might finish, and the postgres connection might be reused for a query from another client. If that's possible in Odyssey then you almost certainly have the same bug.

One easy thing to try is run the test I wrote for this PR against Odyssey and see what happens.

> Anyway the behavior of cancels is somewhat problematic. Sometimes server's CPU is starving due to large amount of fork()s. Clients start to cancel their queries. But cancels produce more fork()s! And ISTM this PR addressess the issue, doesn't it?

I'm not sure why you think this PR addresses the issue. It's definitely not the purpose of the PR, but it might reduce cancellation load as a side-effect.

include/bouncer.h
```diff
@@ -55,13 +55,16 @@ enum SocketState {
...
 	CL_WAITING_LOGIN,	/*   - but return to CL_LOGIN instead of CL_ACTIVE */
 	CL_ACTIVE,		/* pool->active_client_list */
-	CL_CANCEL,		/* pool->cancel_req_list */
+	CL_WAITING_CANCEL,	/* pool->waiting_cancel_req_list */
+	CL_ACTIVE_CANCEL,	/* pool->active_cancel_req_list */
 
 	SV_FREE,		/* free_server_list */
 	SV_JUSTFREE,		/* justfree_server_list */
 	SV_LOGIN,		/* pool->new_server_list */
+	SV_WAIT_CANCELS,/* pool->wait_cancels_server_list */
```
sumedhpathak, line 64:
```suggestion
	SV_WAIT_CANCELS,       /* pool->wait_cancels_server_list */
```

src/objects.c
```diff
@@ -959,16 +1013,28 @@ void disconnect_client(PgSocket *client, bool notify, const char *reason, ...)
 				release_server(server);
 			}
 		}
+		break;
+	case CL_ACTIVE_CANCEL:
+	case CL_WAITING_CANCEL:
+		if (client->cancel_server) {
+			PgSocket *cancel_server = client->cancel_server;
+			statlist_remove(&cancel_server->cancel_clients, &client->cancel_head);
```
sumedhpathak, line 1021: I am a bit unsure about the cancel_server part. Could we chat Monday to discuss this?
I'm not sure how cancel requests are queued, and new server connections given. If that is the case, I'm also now now sure where cancel_server is getting cancel requests queued, since it seems like the *new* server connection given to a cancel request has these cancel clients queued?

sumedhpathak:
The other comment we discussed was to add this behind a feature flag, so we could toggle this off in case there is an issue.

src/objects.c
```diff
@@ -782,6 +807,19 @@ bool release_server(PgSocket *server)
 		fatal("bad server state: %d", server->state);
 	}
 
+	statlist_for_each_safe(cancel_item, &server->cancel_clients, tmp) {
+		PgSocket *cancel_client = container_of(cancel_item, PgSocket, cancel_head);
+		/*
+		 * If the cancel request is not in flight yet we can simply unlink
+		 * the cancel_client. That way it will not send the cancelation
+		 * request to the server, once it gets a server assigned to it.
```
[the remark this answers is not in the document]
JelteF: Yes, updated the comment to make this clearer.

include/bouncer.h
```diff
@@ -293,6 +371,7 @@ struct PgPool {
 
 #define pool_connected_server_count(pool) ( \
 		statlist_count(&(pool)->active_server_list) + \
+		statlist_count(&(pool)->wait_cancels_server_list) + \
```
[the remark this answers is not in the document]
JelteF: I took a look at the usages of these three macros and added comments about how they are used. I added `active_cancels_server_list` to `pool_server_count` but not `pool_connected_server_count`.

src/objects.c
JelteF: I added some more comments to hopefully make this logic clearer.

include/bouncer.h
```diff
@@ -419,6 +522,9 @@ struct PgSocket {
 	usec_t wait_start;	/* client: waiting start moment */
 
 	uint8_t cancel_key[BACKENDKEY_LEN]; /* client: generated, server: remote */
+	struct StatList cancel_clients;	/* clients trying to cancel the query on this connection */
+	PgSocket *cancel_server; /* server that is being canceled by this request */
```
JelteF, line 526: Not the most obvious name, I'll try to think of a better one.

sumedhpathak (approved):
LGTM!

petere:
Is there a reason why `active_cancel` is singular and `wait_cancels` is plural? Could that be more consistent?

JelteF:
> Is there a reason why `active_cancel` is singular and `wait_cancels` is plural? Could that be more consistent?

Yes, `active_cancel` is a server connection on which **a single** active cancelation request. While `wait_cancels` is an idle server connection that is waiting on **one or more** cancellation requests to finish before it is released to the pool.

However, I do agree that the naming of wait_cancels is quite suboptimal. If you have a better idea for a name, feel free to change it.

petere:
> > Is there a reason why `active_cancel` is singular and `wait_cancels` is plural? Could that be more consistent?
> 
> Yes, `active_cancel` is a server connection on which **a single** active cancelation request. While `wait_cancels` is an idle server connection that is waiting on **one or more** cancellation requests to finish before it is released to the pool.

The documentation has this:

```
sv_active_cancel
:   Server connections that are currently forwarding a cancel request
```

So from that it seems it can be more than one.

PR #722, opened 2022-06-03
CI: Install bash on Alpine

JelteF:
It seems the default Alpine image doesn't have bash installed anymore,
but the test.sh script requires it. This installs it on CI.

An example of a failed CI run can be found here: https://cirrus-ci.com/task/6439380507164672


andriisoldatenko:
Looks good!

@petere could you please take another look?

Issue #337, opened 2018-10-29
Feature: support DNS SRV records


atzoum:
Instead of mere IP resolution, SRV records contain port information. This would allow to dynamically change both elements of a backend database via DNS lookup.
Would this be easy to implement? Whenever an SRV record is returned from a DNS lookup I would expect the resolved port to override the configured one.

petere:
This is not an unreasonable idea, but I don't see that anyone is going to work on it.

PR #725, opened 2022-06-13
Add minimal support for SRV records

rodrigol-chan:
This PR adds minimal support for SRV records (#337).


JelteF:
@rodrigol-chan are you still interested in this feature? If so is it possible to add some tests for this?

PR #726, opened 2022-06-15
Fix incorrect/confusing docs

jcoleman:
During investigation of client wait time we came across several pieces of documentation that were either incomplete, confusing, or just plain wrong. This fixes several of those.

I'd like mark the existing `avg_wait_time` calculation as deprecated, but I'll open up something separate for that (because the alternative is fixing it).

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -364,7 +364,7 @@ user
 :   User name.
 
 cl_active
-:   Client connections that are linked to server connection and can process queries.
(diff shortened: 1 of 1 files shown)


jcoleman:
The "Linux (Alpine)" build seems broken separately:
```
./test.sh
env: can't execute 'bash': No such file or directory
make[1]: *** [Makefile:36: check] Error 127
make: *** [Makefile:137: check] Error 2
```

PR #727, opened 2022-06-17
Make avg_wait_time a meaningful metric

### LINKED ISSUE #316: Average Wait time / Units
Hi,

I have some trouble with the metric "average wait time".
I am seeing values vastly exceeding the lifetime of the pgbouncer instance, seeing something like "212233866028us" for an instance, which has been active for a couple of minutes.

I am slightly confused about the calculation of wait time here:
> https://github.com/pgbouncer/pgbouncer/blob/master/src/stats.c#L73

`old` and `cur` hold the accumulated `wait_time` of all the clients at a point in time.
The unit of `wait_time` is time, like `query_time`, and `xact_time`.
But unlike the other time metrics, it is divided over a different time (`dur`), which gives us a fraction.

I would assume, in order to get the average `wait_time`, we would have to divide by a number.
I would suggest to accumulate the number of waiters in stats, and average over the delta at the two points in time.

Sorry, if I am mistaken.
Otherwise, I am willing to offer a patch to implement said behaviour.

Thanks and cheers,
  Fabian

jcoleman:
Currently avg_wait_time is calculated as the wait time accumulated by
clients that were assigned a backend in the current stats_period and
averaged over the time in that period. That means that a connection
that's been waiting for 10 minutes will cause (assuming the default
stats_period of 60s) a reported 10s average wait time in the
stats_period it is assigned a connection (10m = 600s then 600s / 60s).
It's fairly obvious that's not meaningful, because the reported value
has no relation to the wait time experienced by an average client.

If we instead keep track of how many times backends have been assigned
for the current pool (if we were limited to transaction pooling, we
could use transaction count, for example, but we want something that's
generic across all pooling modes) then we can report the wait time
experienced on average by a client assigned a backend during the current
stats period.

As cleanup first (to make sure I understand the stats calculation) I uncovered
and removed dead code (in separate commits).

Fixes #316


include/bouncer.h
```diff
@@ -235,6 +235,7 @@ int pga_cmp_addr(const PgAddr *a, const PgAddr *b);
  * Stats, kept per-pool.
  */
 struct PgStats {
+	uint64_t backend_assignment_count;
```
JelteF, line 238: Do we really need this new counter? Isn't `xact_count` effectively the same as `backend_assignment_count`?

JelteF: I guess it's close but not always the same. Transaction count only gets incremented at the end of the transaction and this gets incremented at the start. So yeah this new counter is needed. I do think we should expose this counter to users too though, otherwise it's impossible to calculate the total wait time for users.

JelteF (changes requested):
This needs a rebase and preferably also some tests, but other than that I think this is a good change.

JelteF (approved):

jcoleman:
@JelteF Thanks for putting the work in to get this over the finish line as I'd been unable to devote the necessary time to do so. Apologies for my being incommunicado here; I've very happy to see this merged.

PR #730, opened 2022-06-26
Prevent Segfault when connecting to non-existent DB with Scram Authentication

mbsabath:
We were seeing segfaults occur when using scram authentication to connect to non-existent databases. Looks like the issue arises when attempting to save the Scram authentication keys. This PR adds a check that the DB isn't fake before attempting to save the credentials

### DIFF
--- a/test/test.sh
+++ b/test/test.sh
@@ -1272,16 +1272,29 @@ test_no_database_auth_user() {
 	return 0
 }
 
-test_no_database_md5_auth_success() {
-	# Testing what happens on successful md5 auth connection to non-existent DB
+test_no_database_scram_auth_success() {
+	# Testing what happens on successful SCRAM auth connection to non-existent DB
 	# Segfaults have been seen after mock authentication was put in place
 
 	admin "set auth_type='md5'"
 	PGPASSWORD=foo psql -X -U scramuser1 -d nosuchdb1 -c "select 1" && return 1
 	grep -F "no such database: nosuchdb1" $BOUNCER_LOG || return 1
 
-	return 0
+    return 0
 }
+
+test_no_database_md5_auth_success() {
+	# Testing what happens on successful SCRAM auth connection to non-existent DB
+	# Segfaults have been seen after mock authentication was put in place
+
+	admin "set auth_type='md5'"
+	PGPASSWORD=foo psql -X -U muser1 -d nosuchdb1 -c "select 1" && return 1
+	grep -F "no such database: nosuchdb1" $BOUNCER_LOG || return 1
+
+    return 0
+}
+
+
 
 test_cancel() {
 	case `uname` in MINGW*) return 77;; esac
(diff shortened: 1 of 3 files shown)


test/test.sh
```diff
@@ -1272,6 +1272,30 @@ test_no_database_auth_user() {
 	return 0
 }
 
+test_no_database_scram_auth_success() {
+	# Testing what happens on successful SCRAM auth connection to non-existent DB
+	# Segfaults have been seen after mock authentication was put in place
+
+	admin "set auth_type='md5'"
```
JelteF, line 1279: After testing, it seems that `md5` was used here on purpose to trigger the bug. Using `scram-256` as auth_type doesn't crash. Would be good to add this detail to the comment above.

test/test.sh
```diff
@@ -1272,6 +1272,30 @@ test_no_database_auth_user() {
...
+	grep -F "no such database: nosuchdb1" $BOUNCER_LOG || return 1
+
+    return 0
+}
+
+test_no_database_md5_auth_success() {
+	# Testing what happens on successful SCRAM auth connection to non-existent DB
+	# Segfaults have been seen after mock authentication was put in place
+
+	admin "set auth_type='md5'"
```
JelteF, line 1290: This test didn't fail before.  I agree that it makes sense for completeness, but maybe swap the auth_type here too (so use `scram-256` instead of md5). Or add all 4 different combinations:
1. scramuser1 + md5
2. scramuser1 + scram-256
3. muser1 + md5
4. muser1 + scram-256

JelteF (commented):
Code looks good. I left some comments on the tests. 

Also CI is failing on Windows. I'm not sure why, but various tests are being skipped on windows. I think we could do the same for the offending test, which seems to be: `test_no_database_md5_auth_success`

Adding this line at the top of the test should skip it on windows:
```
case `uname` in MINGW*) return 77;; esac
```

PS. Alpine is also failing, but that's expected because of: https://github.com/pgbouncer/pgbouncer/pull/722

test/test.sh
mbsabath: Good suggestion. Added all combinations here except for `muser1 + scram` since that failed authentication (as its supposed to).

test/test.sh
mbsabath: I reran the tests using the current master branch of `pgbouncer` and saw the segfault occur when `auth_type` was set to scram or md5 with a scram PW. I've added more detail about the source of the bug to the tests though

test/test.sh
```diff
@@ -1272,6 +1272,30 @@ test_no_database_auth_user() {
...
+	PGPASSWORD=foo psql -X -U scramuser1 -d nosuchdb1 -c "select 1" && return 1
+	grep -F "no such database: nosuchdb1" $BOUNCER_LOG || return 1
+
+    return 0
+}
+
+test_no_database_md5_auth_success() {
+	# Testing what happens on successful SCRAM auth connection to non-existent DB
```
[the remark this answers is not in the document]
mbsabath: Good catch, fun times with lazy and copy and pasting

mbsabath:
Thanks @JelteF for the review! I missed some conditions when I set up those tests so hopefully things will no longer hang on the Windows tests.

Issue #288, opened 2018-04-20
kill_database() and kill_pool() relations


jakubjedelsky:
Hello,

we are using pgbouncer 1.6 (will have 1.8.1 soon). But according to code, it looks the same (to me).

I'd like to ask in which situation function kill_database() runs kill_pool().  We're using `autodb_idle_timeout` with 10 minutes timeout set (see the config below). From time to time, we met this log:

```
Apr 18 12:16:02 pgnode pgbouncer[20380]: dropping database 'db1' as it does not exist anymore or inactive auto-database severity=warning
Apr 18 12:16:02 pgnode pgbouncer[20380]: C-0xb25780: db1/user@<IP>:41951 Pooler Error: database removed severity=warning
Apr 18 12:16:02 pgnode pgbouncer[20380]: C-0xb25780: (nodb)/(nouser)@<IP>:42502 registered new auto-database: db = db1 severity=info
```
At the time, when database is removed, client returns connection error. I have an idea, that the new connection/query is starting right in the moment, when pgbouncer is cleaning inactive dbs, thus it meet the error. Is that possible? I can imagine, there is a few miliseconds, between kill_database and kill_pool, where a new connection can appear and thus can be killed. Or am I completely wrong?

Thanks.

```
pgbouncer=# show config;
            key             |                         value                          | changeable 
----------------------------+--------------------------------------------------------+------------
 job_name                   | pgbouncer                                              | no
 conffile                   | /etc/pgbouncer/pgbouncer.ini                           | yes
 logfile                    |                                                        | yes
 pidfile                    | /var/run/pgbouncer/pgbouncer.pid                       | no
 listen_addr                | *                                                      | no
 listen_port                | 6432                                                   | no
 listen_backlog             | 128                                                    | no
 unix_socket_dir            | /tmp                                                   | no
 unix_socket_mode           | 511                                                    | no
 unix_socket_group          |                                                        | no
 auth_type                  | md5                                                    | yes
 auth_file                  | /etc/pgbouncer/userlist.txt                            | yes
 auth_query                 | SELECT usename, passwd FROM pg_shadow WHERE usename=$1 | yes
 pool_mode                  | session                                                | yes
 max_client_conn            | 1000                                                   | yes
 default_pool_size          | 40                                                     | yes
 min_pool_size              | 0                                                      | yes
 reserve_pool_size          | 30                                                     | yes
 reserve_pool_timeout       | 5                                                      | yes
[...]

PR #734, opened 2022-06-29
Track connection waiting for password in a pool

janchochol:
After receiving `PKT_STARTUP` packet, client is associated to database
(result of `find_database` is stored in `client->db`).

For some authentication mechanisms (like plain or MD5), pgbouncer is
waiting for password packet, before it switches client to `CL_ACTIVE`
state).

If there is enough delay between these two packets, it can happen that
`do_full_maint` is called.
These clients are not linked anywhere in pool, so it can happen that
auto database can be marked as inactive.

When password packet arrives, there is no mechanism to un-mark such
database as inactive (in case of no new connections), so it can be
later killed leading to errors like:

```
Jun 27 08:46:13 node001 pgbouncer[16819]: C-0x115b218: db_001/user01@10.1.1.1:43132 pooler error: database removed severity=warning
```

Even worse is situation, when password packet arrives after
`autodb_idle_timeout`, which leads to use-after-free situation (as used
database is free). Fortunately this is not probable, as
`autodb_idle_timeout` is generally larger that delay between first two
packets.

This fix ensures, that after `PKT_STARTUP` client connection is tracked
in a pool for all cases (base on code analysis, this seemd as last place).

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -52,6 +52,7 @@ enum SocketState {
 	CL_FREE,		/* free_client_list */
 	CL_JUSTFREE,		/* justfree_client_list */
 	CL_LOGIN,		/* login_client_list */
+	CL_AUTH,		/* pool->auth_client_list */
 	CL_WAITING,		/* pool->waiting_client_list */
 	CL_WAITING_LOGIN,	/*   - but return to CL_LOGIN instead of CL_ACTIVE */
 	CL_ACTIVE,		/* pool->active_client_list */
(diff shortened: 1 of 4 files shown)


janchochol:
I am not sure, if proposed way is correct fix.

As we are periodically hit by this issue in the production, we are prepared to change used approach upon request.

janchochol:
This should also fix https://github.com/pgbouncer/pgbouncer/issues/288, as we had more time for analysis this time.

JelteF:
Do you think there's a way to create a test for the issue you're describing?

JelteF (changes requested):
Would be good to have a regression test for this issue included in the PR. Also many existing tests are failing so this code likely has a bug somewhere.

janchochol:
Sorry for late response, but we unfortunately have currently other priorities.
We will definitely investigate how to add some regression tests.

JelteF:
Writing regression tests should be a lot simpler now. Because the framework is now using python. Another thing I realised is that this new client type should probably also be made visible by `SHOW CLIENTS` in admin.c

JelteF:
@janchochol are you intending to finish this?

eulerto:
Closed due to inactivity.

PR #448, opened 2019-12-18
add server_shuffle_hosts

ericjster:
This feature shuffles the host list returned from DNS, providing a simple way to balance connections to postgres servers.

This feature was motivated by a DNS server that always returns results in the same order, resulting in one postgres server overloaded and others not being used. This shuffling feature might be useful if the DNS server does not support or is not configured to do round robin, or in lieu of a load balancer.

It may be useful to combine with server_round_robin, which cycles clients through all postgres server connections. Also server_lifetime might be reduced from its default of 3600 (1 hour) to 600 (10 minutes) to more quickly adjust to infrastructure changes.

The implementation shuffles the order of the linked list of addrinfo returned by getaddrinfo. We think that reordering the linked list is a safe operation based on RFC 3493: "The freeaddrinfo() function must support the freeing of arbitrary sublists of an addrinfo list originally returned by getaddrinfo()." We also reviewed several implementations of freeaddrinfo and all we examined traverse the linked list and free memory for each addrinfo independently.

PR #736, opened 2022-07-05
Add `load_balance_hosts` parameter to db config

dpirotte:
PGBouncer's default behavior is to round-robin between comma-separated
hosts on each new connection attempt. This is desirable when all hosts
are similar, such as load balancing over multiple replicas, but
undesirable when only one host out of the list is expected to
successfully connect and login. In this scenario, every other connection
attempt would hit an invalid host and fail.

The `host_strategy` parameter controls this host selection behavior. The
default behavior is labeled `round_robin` and a new strategy called
`last_successful` is introduced. This new strategy instructs PGBouncer
to prefer the host with most recent successful connections. When that
host fails, new connections will round-robin through the list until a
successful connection occurs.


src/loader.c
```diff
@@ -154,12 +154,14 @@ bool parse_database(void *base, const char *name, const char *connstr)
...
 	PgDatabase *db;
 	struct CfValue cv;
+	struct CfValue host_strategy_lookup;
 	int pool_size = -1;
 	int min_pool_size = -1;
 	int res_pool_size = -1;
 	int max_db_connections = -1;
 	int dbname_ofs;
 	int pool_mode = POOL_INHERIT;
+	int host_strategy = ROUND_ROBIN;
```
JelteF, line 164:
```suggestion
	enum HostStrategy host_strategy = ROUND_ROBIN;
```

src/loader.c
```diff
@@ -236,6 +241,11 @@ bool parse_database(void *base, const char *name, const char *connstr)
 			res_pool_size = atoi(val);
 		} else if (strcmp("max_db_connections", key) == 0) {
 			max_db_connections = atoi(val);
+		} else if (strcmp("host_strategy", key) == 0) {
```
JelteF, line 244: This needs an entry in `config.md` to describe its usage.

src/objects.c
```diff
@@ -1069,6 +1069,9 @@ static void dns_connect(struct PgSocket *server)
 		int count = 1;
 		int n;
 
+		if (!server->pool->last_connect_failed && server->pool->db->host_strategy == LAST_SUCCESSFUL)
+			server->pool->rrcounter = server->pool->last_successful_rrcounter;
```
JelteF, line 1073: The current PR seems to have the desired effect based on the tests you wrote, but I have the  feeling like the same could probably be achieved without the two new fields you added: `server->rrcounter` and `pool->last_successful_rrcounter`.

How about we simply increase the `pool->rr_counter` whenever a connection attempt fails.

JelteF (commented):
Thank you for the contribution, it looks useful, postgres connection strings work in a similar way so it makes sense to support that as well. 

I left some comments. And can you please rebase/merge this PR because there's a merge conflict.

src/loader.c
dpirotte: Documentation is in. Feedback welcome.

src/objects.c
dpirotte: Thanks for the feedback. I simplified the logic as you suggest and it looks better to me.

src/loader.c
```diff
@@ -299,6 +309,7 @@ bool parse_database(void *base, const char *name, const char *connstr)
 	db->res_pool_size = res_pool_size;
 	db->pool_mode = pool_mode;
 	db->max_db_connections = max_db_connections;
+	db->host_strategy = host_strategy;
```
goksgie, line 312: if the `host_strategy` of a database can change, should we tag it as [changed](https://github.com/pgbouncer/pgbouncer/blob/b35c9a67345e2b7789724f8974f16be565827992/src/loader.c#L300)?

dpirotte: That's a good call. Yes, I think we should tag it so behavior changes after a reload. I will make that change. Thanks for the feedback!

dpirotte: `host_strategy` is now reloadable.

`show databases`/`show pools` also now include `host_strategy` in their output, which makes this testable and is useful on its own.

dpirotte:
@petere @JelteF I rebased this PR and made @goksgie's suggested change above.

Please let me know if anything else needs to happen to get this into a mergeable state.

JelteF:
I think the code in the PR is fine now (apart from the merge conflict that occured due to another merged PR). But after playing with these changes a bit I found one other issue, and I'm wondering how useful this feature is without also addressing this issue:
`server_login_retry` creates a significant pause anytime a single node cannot be connected to.

@dpirotte How are you using this change, that it is useful in its current form?

JelteF:
Hi @dpirotte, github notified me that you rebased this on a branch of your fork. I think a rebased version of this can be merged pretty easily. Could you still explain how you use this in practice? Because when I played with it, another bug in pgbouncer was causing serious issues for me.

dpirotte:
Hi @JelteF. Yeah, I’m working on rebasing this and another patch on top of 1.19. The code works but I haven’t updated to your new (much improved!) test framework.

The purpose here is to support faster failovers in environments where 1/N hosts in the list will be the primary and (N-1)/N hosts will be replicas _and_ where DNS updates are potentially too slow. RDS/Aurora is one such environment, and I believe some Patroni configurations are as well.

Given the following configurations:
```
; pg-primary is a CNAME to whichever of pg-0 or pg-1 is the primary
current = host=pg-primary.cluster.local connect_query=‘select disconnect_self() if pg_is_in_recovery()’…
proposed = host=pg-0.cluster.local,pg-1.cluster.local host_strategy=last_successful connect_query=‘select disconnect_self() if pg_is_in_recovery()’…
```

After a failover, `current` may take some time to point to the correct new primary, i.e. due to DNS caching. To speed up failovers, `proposed` takes the full list of potential primaries so pgbouncer can iterate through until it finds the primary. (And, to your earlier comment, this does encourage a low value for `server_login_retry`.)

A key part here is the `connect_query`, which asserts that pgbouncer is only ever talking to a writable database. Without `host_strategy`, using the full list of potential primaries means that (N-1)/N new connections will fail and wait for `server_login_retry` before trying another host. With `host_strategy=last_successful`, we only fail connections until a new primary is found, and direct all new connections there until the next failure. This minimizes failed connection noise and time waiting for `server_login_retry`.

I have another patch that adds lightweight `target_session_attrs` support albeit only on PG 14+ hosts where the server informs the client of its status as a primary or standby. This replaces the `connect_query` above, but is not a replacement for `host_strategy`. One could imagine having three hosts (primary and two replicas) where you want `host_strategy=last_successful` for the primary db connection string but still want to load balance on the replica connection string with `host_strategy=round_robin`.

Another option would be to keep the round-robin behavior and make `target_session_attrs` loop through hosts and gracefully disconnect until it finds one matching the requested state. I prefer having both configuration options so that `target_session_attrs=primary host_strategy=last_successful` is opening/closing as few connections on replicas as possible.

I’ll bump once the tests are ported to pytest and we’ll go from there.

dpirotte:
@JelteF Rebased `host_strategy` with pytests if you want to take a look while I sort out the macOS build.

doc/config.md
```diff
@@ -1060,6 +1060,20 @@ they are logged but ignored otherwise.
 Set the pool mode specific to this database. If not set,
 the default `pool_mode` is used.
 
+### host_strategy
```
JelteF, line 1063: I think it would make sense to change the name of the option to `load_balance_hosts`. This option got added to libpq for PG16 (by me). It seems nice to use consistently named options across projects. The `last_successful` option name should then be changed to `disable`. And lets change `round_robin` to `round-robin`, to be consistent with other arguments like `verify-full` and `read-write`.

PS. I especially think consistency is nice here, because you mentioned you also want to add `target_session_attrs` to PgBouncer. Since `target_session_attrs` and `load_balance_hosts` can be used nicely together, I think it's especially useful to have the same names for both Postgres and PgBouncer.

JelteF: Btw, adding a `random` as an option would be a nice future improvement to make sure PgBouncer supports the same as libpq. But that should definitely be another follow up PR. We can merge this PR easily without the random functionality.

dpirotte: > I think it would make sense to change the name of the option to load_balance_hosts. This option got added to libpq for PG16 (by me).

Ha, good call. I noticed that you contributed this feature in the release notes. +100 to consistent naming with libpq. I'll take care of that.

dpirotte:
@jeltef Updated naming per discussion above.

The macOS build passes now. The mingw{32,64} builds failed a couple times but seem to pass now, so those might have been random failures. The most recent build failures look like Cirrus issues: `Failed to start an instance: INTERNAL: API response null(0): An error has occurred. Failed to connect to /34.122.100.68:443`.

test/test.ini
```diff
@@ -30,6 +30,9 @@ authdb = port=6666 host=127.0.0.1 dbname=p1 auth_user=pswcheck
 
 hostlist1 = port=6666 host=127.0.0.1,::1 dbname=p0 user=bouncer
 hostlist2 = port=6666 host=127.0.0.1,127.0.0.1 dbname=p0 user=bouncer
+hostlist_good_first = port=6666 host=127.0.0.1,127.0.0.3 dbname=p0 user=bouncer load_balance_hosts=disable
+hostlist_bad_first = port=6666 host=unresolvable-hostname,127.0.0.1 dbname=p0 user=bouncer load_balance_hosts=disable
+load_balance_hosts_update = port=6666 host=127.0.0.1,127.0.0.3 dbname=p0 user=bouncer load_balance_hosts=disable
```
JelteF, line 35: The load_balance_hosts_update database isn't used (afaict). Did you have a test in mind for it?

doc/config.md
```diff
@@ -1060,6 +1060,20 @@ they are logged but ignored otherwise.
 Set the pool mode specific to this database. If not set,
 the default `pool_mode` is used.
 
+### load_balance_hosts
+
+When a comma-separated list is specified in `host`, `load_balance_hosts` controls
+which entry is chosen for a new connection.
```
JelteF, line 1066: It would be great if this same logic would be used when receiving multiple IP adresses from DNS. The easiest way to test this is to create lines in `/etc/host`. Libpq its load_balance_hosts version does this too. So if we don't implement this, we should at least mention in the docs that this setting currently doesn't impact DNS based load balancing.

doc/config.md
```diff
@@ -1060,6 +1060,20 @@ they are logged but ignored otherwise.
...
+
+round-robin
+:   A new connection attempt chooses the next host entry in the list.
+
+disable
+:   A new connection continues using the same host entry until a connection
+    fails, after which the next host entry is chosen.
```
JelteF, line 1073: It would be good to mention that it is recommended to set `server_login_retry` lower than the default to ensure fast retries, because of issue https://github.com/pgbouncer/pgbouncer/issues/872

doc/config.md
eulerto: Yeah. I'm wondering if this feature also includes #448 or at least to have it in mind.

JelteF: Interesting, I didn't see that PR before (I guess I didn't go more than 2 years into PR history when joining the project). Right now #448 is completely distinct from this PR:
1. This PR does not implement the `disable` mode for DNS load balancing. It still uses round robin even if `disable` is used.
2. This PR does not implement the `random` mode. PR #448 implements `random` mode for DNS load balancing only, not for host lists (certainly because host lists were not available in PgBouncer in the time that PR was written).

Th design of the config option would allow for implementing both 1 and 2 though. I think 2 isn't required to implement for this PR. A completely missing option isn't very confusing. But I think 1 is quite confusing because the existing option does not apply to all the places that you would expect, so preferably this PR would implement 1 too. If that's hard for some reason, then I think a note in the docs is also sufficient.

doc/config.md
petere: I prefer the initial naming. It seems clear. load_balance_hosts seems to be a misnomer, since no "load" and no "balancing" is involved?

JelteF: > I prefer the initial naming. It seems clear.

I think the initial name was a fine name too. But I think being in line with naming of libpq, jdbc and npgsql is a big advantage of the load_balance_hosts naming.

> load_balance_hosts seems to be a misnomer, since no "load" and no "balancing" is involved?

The "load" is the number of connection that are made by pgbouncer to the database, and those connections are "balanced" across the different servers in a round-robin fashion (or random fashion when implemented in the future). And you can enable this balancing by using `disable` which will select a working host and send all traffic there. 

IMHO, the old naming describes how pgbouncer chooses hosts. The new naming describes what the user will see happening to their setup by changing the setting. I think I personally like names that describe the effects instead of the method.

doc/config.md
cosgroveb: Hey @JelteF,

I work with the author and he's been a bit too busy to finish responding to your feedback so I'd like to take a stab at it. 

I don't quite understand 1. Maybe I don't understand pgbouncer internals well enough yet but can you give me some pointers on implementing the `disable` mode for DNS load balancing? Sorry for such a vague ask for clarification but I do want to complete this feature as soon as I can.

JelteF: So to clarify: PgBouncer does round-robin load balancing in two ways:
1. If you have multiple hostnames in the `host` key, then it will round robin across these.
2. If a single DNS entry returns multiple IPs, then it load balanaces using round-robin across these IPs: https://www.pgbouncer.org/faq.html#how-to-load-balance-queries-between-several-servers

This same thing is true for `libpq` and `load_balance_hosts` can be used to control the behaviour for both of these there. The implementation of `load_balance_hosts` in this PR only controls the behaviour of 1, not 2. And I think it should control both.

cosgroveb: Aha thank you for the clarification. I'll look at working up 2.

cosgroveb: @JelteF I re-based this and took a look but I was wondering: is it possible to merge this with a prominent disclaimer in the docs as you suggested here:

> If that's hard for some reason, then I think a note in the docs is also sufficient.

I would strongly prefer implementing this in a way that's more congruent with libpq and so I would commit to working on a follow-on PR that does that.

The reason I'm so interested in merging this as soon as possible is because:

1. We've been using this and the `target_session_attrs` patch mentioned above in production for some time now. I'm maintaining that and...
2. I detest running an internal fork
3. I'm working through an improvement to our `target_session_attrs` but quite honestly I don't want to change even our internal patch without trying to get it upstreamed
4. Essentially, I want your feedback on my improvements to `target_session_attrs` but to do that we need to get your feedback on our current implementation :)
5. PR-ing that patch doesn't make sense to me without this being incorporated first

Let me know! https://github.com/cosgroveb/pgbouncer/tree/host-strategy-cos-rebased

JelteF: > is it possible to merge this with a prominent disclaimer in the docs as you suggested here

I think it's acceptable, but it would be nice if you could explain why it's not easy to implement it.

cosgroveb: Hey @JelteF I think that I'm finding this to be a a bit of a yak-shave to start because to test it properly I feel that I need to add a DNS server to the test infrastructure. It's _doable_ don't get me wrong but it does make it difficult to start test-driving the work IMHO.

JelteF: It's actually pretty easy to test this without setting up a full featured DNS server. You can add the same hostname multiple times to the `/etc/hosts` file. i.e. if you add the following there and connect to `myhost` pgbouncer should load balance across 127.0.0.1 and 127.0.0.2
```
127.0.0.1   myhost
127.0.0.2   myhost
```

This trick is actually used by the Postgres tests for `load_balance_hosts` (but to be clear, no automated tests would be needed for PgBouncer for that in this PR).

cosgroveb: That's not standard behavior for /etc/hosts. Interesting

JelteF: Afaict it's standard behaviour for `getaddrinfo` to return both IP addresess. The DNS level load balancing is then implemented by PgBouncer over this list of IP addresses that it receives (based on my reading of the Pgbouncer docs, i've never tested this feature myself). This is also how `load_balance_hosts` is implemented by libpq.

cosgroveb: I was mistaken. Looks like `getent hosts XX` returns all IPs listed in /etc/hosts for XX.

cosgroveb: OK then the only thing that makes this hard @JelteF is my own understanding. Is this where the DNS rr needs to be disabled? https://github.com/cosgroveb/pgbouncer/blob/host-strategy-cos-rebased/src/objects.c#L1647-L1649

cosgroveb: Okay while `getent hosts XX` returns all the IPs listed in /etc/hosts should I expect to see all the IPs in /etc/hosts for my database host? [I put together this demonstration](https://github.com/pgbouncer/pgbouncer/commit/7b1ff7e69e8906699d94972905dc45b02fd9e42e):

```python
def test_load_balance_hosts_disable_with_dns(bouncer, pg):
    bouncer.default_db = "dns_load_balance_hosts_disable"
    hosts = f"""
127.0.0.53       dnsdbhost
127.0.0.54       dnsdbhost
127.0.0.55       dnsdbhost
    """

    # I have modified utils.py to set listen_addresses='*' for this
    #  demonstrationsbut would use pg.configure/pg.reload for a real test
    with bouncer.run_with_appended_etc_hosts(hosts):
        subprocess_result = capture(
            ["getent", "hosts", "dnsdbhost"],
        )
        getent_result = subprocess_result.split("\n")
        assert "127.0.0.53      dnsdbhost" == getent_result[0]
        assert "127.0.0.54      dnsdbhost" == getent_result[1]
        assert "127.0.0.55      dnsdbhost" == getent_result[2]

        bouncer.sql(query="SELECT 1", user="bouncer", password="zzzz", dbname="dns_load_balance_hosts_disable")
        with bouncer.admin_runner.cur() as cur:
            results = cur.execute("SHOW DNS_HOSTS").fetchall()
            result = [r for r in results if r[0] == 'dnsdbhost'][0]

            assert result[2] != "127.0.0.53:0"
```

```====================================================== FAILURES =======================================================
______________________________________ test_load_balance_hosts_disable_with_dns _______________________________________
/home/admin/code/pgbouncer/test/test_load_balance_hosts.py:70: in test_load_balance_hosts_disable_with_dns
    assert result[2] != "127.0.0.53:0"
E   AssertionError: assert '127.0.0.53:0' != '127.0.0.53:0'
        bouncer    = <test.utils.Bouncer object at 0x7f0499fbb2e0>
        cur        = <psycopg.Cursor [closed] [BAD] at 0x7f049b445460>
        getent_result = ['127.0.0.53      dnsdbhost', '127.0.0.54      dnsdbhost', '127.0.0.55      dnsdbhost', '127.0.0.53      dnsdbhost', '127.0.0.54      dnsdbhost', '127.0.0.55      dnsdbhost', ...]
        hosts      = '\n127.0.0.53       dnsdbhost\n127.0.0.54       dnsdbhost\n127.0.0.55       dnsdbhost\n    '
        pg         = <test.utils.Postgres object at 0x7f0499fbbac0>
        result     = ('dnsdbhost', 14, '127.0.0.53:0')
        results    = [('dnsdbhost', 14, '127.0.0.53:0')]
        subprocess_result = '127.0.0.53      dnsdbhost\n127.0.0.54      dnsdbhost\n127.0.0.55      dnsdbhost\n127.0.0.53      dnsdbhost\n127.0.0.54      dnsdbhost\n127.0.0.55      dnsdbhost\n'
```

Because the work on the related feature, `target_session_attrs` is substantially complete and in need of review I do favor merging this branch as soon as possible. As you can see, the testing situation is a _bit icky_ at the moment for the DNS sub-feature of load_balance_hosts but with some time I can complete it. It would be super beneficial to me if we can start working on getting `target_session_attrs` and I think it would be to other users as well.

JelteF: Sounds reasonable to me.

cosgroveb: I'll get push access to @dpirotte's fork, push up a more recent rebase, and address your docs feedback below `it is recommended to set server_login_retry lower than the default`

cosgroveb:
@JelteF I rebased this and addressed the docs feedback above. Let me know if you run into any issues merging this!

doc/config.md
```diff
@@ -1237,6 +1237,23 @@ they are logged but ignored otherwise.
 Set the pool mode specific to this database. If not set,
 the default `pool_mode` is used.
 
+### load_balance_hosts
+
+When a comma-separated list is specified in `host`, `load_balance_hosts` controls
+which entry is chosen for a new connection.
```
JelteF, line 1243:
```suggestion
which entry is chosen for a new connection.

Note: This setting currently only controls the load balancing behaviour when providing multiple hosts in the connection string, but not when a single host its DNS record references multiple IP addresses. This is a missing feature, so in a future release this setting might start to to control both methods of load balancing.
```

doc/usage.md
```diff
@@ -563,6 +566,9 @@ server_lifetime
 pool_mode
 :   The database's override pool_mode, or NULL if the default will be used instead.
 
+host_strategy
```
JelteF, line 569:
```suggestion
load_balance_hosts
```

JelteF (changes requested):
I'll need to give this another round of testing, but from what I remember the functionality worked as advertised in the past.

I left a few small comments on the documentation. Also formatting check is still failing.

doc/config.md
cosgroveb: Ah yes I forgot to disclaim. Thank you.

cosgroveb:
Thanks @JelteF. I pushed another fix for the linting and will keep an eye on CI.

JelteF:
Sorry for the delay here. I'll try to play around with this soonish. 

Aside from that, since today there's now a discord channel to discuss PgBouncer development. It would be nice if you could join that: https://discordapp.com/channels/1258108670710124574/1300532992304742481

cosgroveb:
@JelteF can you link the Discord server? I wonder if I need to join the server first. Your channel link takes me here:

image

cosgroveb:
Aha it must be PostgreSQL Hackers https://lnkd.in/g8n3dZfx

JelteF (approved):

PR #756, opened 2022-08-18
Apply PG Bouncer Server Side Prepared Statements Patch

dashorst:
The patch was developed against 1.14, so this PR (which is against 1.16) probably won't work properly. However, github doesn't allow me to craft a PR against a tag and there's no 1.14 branch available :man_shrugging:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,12 +10,14 @@ pgbouncer_SOURCES = \
 	src/hba.c \
 	src/janitor.c \
 	src/loader.c \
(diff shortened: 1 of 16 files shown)


JelteF:
Thank you, I tried to see if there were merge conflicts with the current master branch and there is only a tiny one that is easy to fix.  Could you fix that merge conflict (preferably by rebase, but a merge commit is also fine) and open the PR against the master branch instead? Then CI will work as expected.

Issue #695, opened 2022-03-16
Yet Another Server Side Prepared Statements Patch


dashorst:
I know this is completely unsolicited, but I'd like to ask for your consideration to evaluate whether and under what conditions [these commits][patch] can be integrated with pgbouncer.

## TL;DR

- we have [patched pgbouncer to support prepared statements][patch]
- the proposed [changes to pgbouncer can be seen in this diff][diff]
- we want to contribute the patch to the pgbouncer community
- do you want this feature at all (e.g. prepared statement support)
- do you want this patch at all?
- what must we do to make this patch acceptable?

## The Case for Server Side Prepared statements support

One of the biggest issues when we migrated from Oracle to Postgresql and using pgbouncer was the lack of server-side prepared statement caching. This resulted in an enormous amount of inbound network traffic from our application servers due to the queries being constantly sent over for parsing, and in a really high CPU usage on the database server due to having to parse the statements.

After deploying the code in this prepared statements patch project our CPU usage was considerably lower, and the network traffic went from 400mbit/s to 20mbit/s IIRC (it was in summer 2020 so specific details are becoming vague)

### Figure 1: Database CPU usage before and after using the patched PG Bouncer

![db-cpu-usage](https://user-images.githubusercontent.com/46626/158636579-15f18ace-65c9-4ad3-b4d2-c88e44804b38.png)

As you can see in this graph the CPU usage after the upgrade to the patched pgbouncer is significantly lower than before.

### Figure 2: Network traffic before and after using the patched PG Bouncer

![network](https://user-images.githubusercontent.com/46626/158638296-65fffe04-f0b5-4c2c-ad6e-59db933e4d41.png)

The contrast between running without server side prepared statements and using server side prepared statements is very apparent in the "incoming" network traffic graph in Figure 2. The amount of incoming traffic on the database server goes from WOW to practically non-existent.

### Figure 3: application response times before and after the patch

** Before the patch **
![response-times-before-ps](https://user-images.githubusercontent.com/46626/158636607-f59fa223-20dc-40d3-a909-bc3c57442ff9.png)

** After the patch **
![response-times-after-ps](https://user-images.githubusercontent.com/46626/158636605-c3656b78-7238-4c16-b258-0513090e5e36.png)

As you can see using the patch the response times of our application dropped significantly in a good way. Lower is better.

## Proposal

[...]

PR #757, opened 2022-08-18
Add Server-side Prepared Statements Cache

dashorst:
Unfortunately [PgBouncer](https://www.pgbouncer.org/) does not support prepared statements in transaction pooling mode, as stated in the FAQ:

> To make prepared statements work in this mode would need PgBouncer to keep track of them internally, which it does not do. So the only way to keep using PgBouncer in this mode is to disable prepared statements in the client.

This patch introduces [prepared statement tracking](#how-it-works), allowing to run PgBouncer in transaction pooling mode with server side prepared statements. It **only** supports the extended query protocol parse, bind, execute and close [message formats](https://www.postgresql.org/docs/12/protocol-message-formats.html) E.g. the flow used by the PostgreSQL JDBC driver to create server prepared statements.

## How it works

PgBouncer has client connections (application :left_right_arrow: PgBouncer) and server connections (PgBouncer :left_right_arrow: PostgreSQL). Prepared statements received with a 'parse' command are stored on the receiving client connection together with a hash of the SQL. Parse commands sent to PostgreSQL are modified to have a unique prepared statement name for the server connection. The server connection stores the SQL hash and the prepared statement name unique to that server connection.

Whenever a client connection receives a command involving a prepared statement, it does a lookup to get the prepared statement metadata (SQL sent by the parse command and the SQL hash). The server connection maps this SQL hash to the prepared statement name unique to that server connection and modifies the prepared statement name before sending the command to PostgreSQL. If the SQL hash cannot be found for the server connection, this means we are on a different server connection than the one we originally issued the parse command on. In this case, a parse command is generated and sent first.

## Configuration

Create a configuration file, using `./etc/pgbouncer.ini` as a starting point.

By default, prepared statement support is not enabled. To enable it just add the following to your pgbouncer-ps configuration:

```
disable_prepared_statement_support = 0
```

The number of prepared statements kept active on a single backend connection is defined by the following configuration:

```
prepared_statement_cache_queries = 100
```

Note: keep in mind that this will increase the memory footprint of each client connection on your PostgreSQL server.

Tracking prepared statements does not only come with a memory cost, but also with increased CPU usage. Multiple PgBouncer instances can be run to use more than one core for processing, see `so_reuseport` socket option.

## Connect the application

Configure your client application as though you were connecting directly to a PostgreSQL database.

Example - JDBC driver

```
jdbc:postgresql://pgbouncer-ps:6432/postgres?prepareThreshold=10&preparedStatementCacheQueries=512&preparedStatementCacheSizeMiB=10
```

Note: Cached prepared statements by the JDBC driver will increase the memory footprint of each JDBC connection in your application and each front-end connection in PgBouncer.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -880,6 +880,8 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 {
 	SBuf *sbuf = &client->sbuf;
 	int rfq_delta = 0;
+  uint8_t ps_action = PS_IGNORE;
+  PgClosePacket *close_packet;
 
 	switch (pkt->type) {
 
@@ -913,10 +915,60 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 	 * to buffer packets until sync or flush is sent by client
 	 */
 	case 'P':		/* Parse */
+    if (!cf_disable_prepared_statement_support) {
+      if (!inspect_parse_packet(client, pkt, &ps_action))
+        return false;
+      
+      if (ps_action == PS_HANDLE && incomplete_pkt(pkt)) {
+        client->packet_cb_state.flag = CB_WANT_COMPLETE_PACKET;
+        sbuf_prepare_fetch(sbuf, pkt->len);
+        return true;
+      }
+    }
+    break;
+
 	case 'E':		/* Execute */
+    break;
+
 	case 'C':		/* Close */
+    if (!cf_disable_prepared_statement_support) {
+      if (!unmarshall_close_packet(client, pkt, &close_packet))
+        return false;
+
+      if (close_packet && is_close_statement_packet(close_packet)) {
+        if (!handle_close_statement_command(client, pkt, close_packet))
+          return false;
+
+        client->pool->stats.client_bytes += pkt->len;
+
+        free(close_packet->name);
+        free(close_packet);
+
+        /* No further processing required */
+        return true;
+      }
+
+      if (close_packet) {
+        free(close_packet->name);
+        free(close_packet);
+      }
+    }
+    break;
+
 	case 'B':		/* Bind */
+    if (!cf_disable_prepared_statement_support) {
+      if (!inspect_bind_packet(client, pkt, &ps_action))
+        return false;
+    }
+    break;
+
 	case 'D':		/* Describe */
+    if (!cf_disable_prepared_statement_support) {
+      if (!inspect_describe_packet(client, pkt, &ps_action))
+        return false;
+    }
+    break;
+
 	case 'd':		/* CopyData(F/B) */
 		break;
 
@@ -930,7 +982,7 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 		return false;
 	}
 
-	/* update stats */
+  /* update stats */
 	if (!client->query_start) {
 		client->pool->stats.query_count++;
 		client->query_start = get_cached_time();
@@ -960,10 +1012,58 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 	client->link->ready = false;
 	client->link->idle_tx = false;
 
+  if (ps_action == PS_HANDLE) {
+    switch (pkt->type)
+    {
+      case 'P':
+        if (!handle_parse_command(client, pkt))
+          return false;
+
+        if (client->packet_cb_state.flag == CB_HANDLE_COMPLETE_PACKET) {
+          /* pkt already read from sbuf with cb */
+          client->packet_cb_state.flag = CB_NONE;
+        } else {
+          sbuf_prepare_skip(sbuf, pkt->len);
+        }
+        return true;
+
+      case 'B':
+        client->packet_cb_state.flag = CB_REWRITE_PACKET;
+        sbuf_prepare_fetch(sbuf, pkt->len);
+        return true;
+
+      case 'D':
+        if (!handle_describe_command(client, pkt))
+          return false;
+
+        return true;
+    }
+  }
+
 	/* forward the packet */
 	sbuf_prepare_send(sbuf, &client->link->sbuf, pkt->len);
 
 	return true;
+}
+
+static bool handle_partial_client_work(PgSocket *client, PktHdr *pkt_hdr, unsigned offset, struct MBuf *data)
+{
+	struct PktBuf *pkt = pktbuf_temp();
+
+  switch (pkt_hdr->type) {
+    case 'B':
+      if (offset == 0) {
+        if (!handle_bind_command(client, pkt_hdr, offset, data, pkt))
+          return false;
+      }
+      pktbuf_put_bytes(pkt, data->data + mbuf_consumed(data) , mbuf_avail_for_read(data));
+      break;
+  }
+
+  if (!pktbuf_send_immediate(pkt, client->link))
+    return false;
+
+  return true;
 }
 
 /* callback from SBuf */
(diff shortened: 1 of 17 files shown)


achanda:
I tried running this patch with the Go `pgx` driver and this code

```
db, err := sql.Open("pgx", "postgres://user:pass@localhost:5433/user")
if err != nil {
    panic(err)
}

rows, err := db.Query("SELECT * FROM foobar")
if err != nil {
    panic(err)
}
defer rows.Close()
```

Since pgx does unnamed prepared statements by default, and this fails with `panic: FATAL: prepared statement '' not found (SQLSTATE 08P01)`

Interestingly, the following code with libpq works fine

```
db, err := sql.Open("pgx", "postgres://user:pass@localhost:5433/user")
if err != nil {
    panic(err)
}

stmt, err := db.Prepare("SELECT * FROM foobar")
if err != nil {
    panic(err)
}
defer stmt.Close()

rows, err := stmt.Query()
if err != nil {
    panic(err)
}
```

Looking at packet capture for both cases, I see `P`, `D` and `B` packets in both cases. In the case when it works, the client gets back a `BindComplete` message, which it does not get in the error case. Both captures are between client and pgbouncer.

Good case:
image

Bad case:
image

I did not see any errors in Postgres logs. PgBouncer logs ends with these

```
NOISE postgres/user@[::1]:61695 read pkt='B' len=27
NOISE postgres/user@[::1]:61695 inspect_bind_packet: type=B, len=27, statement=lrupsc_1_0, portal=, action=1
DEBUG postgres/user@[::1]:61695 handle_bind_command: mapped statement 'lrupsc_1_0' (query hash '154723704867692259631581161632731655014') to 'B_1'
NOISE postgres/user@[::1]:61695 read pkt='D' len=7
NOISE postgres/user@[::1]:61695 inspect_descibe_packet: type=D, len=7, P/S=P, action=1
LOG postgres/user@[::1]:61695 closing because: prepared statement '' not found (age=0s)
LOG postgres/user@[::1]:5432 closing because: client disconnect while server was not ready (age=0s)
WARNING postgres/user@[::1]:61695 pooler error: prepared statement '' not found
```

which seems to be from this line https://github.com/pgbouncer/pgbouncer/pull/757/files#diff-496884af24ad5ffd6995005241d3348a0eb1f7613b028a323fd2aac4d9819889R248

What might be the best way to handle the case when the prepared statement is not found in PgBouncer cache?

carukc:
@thecodingrobot ... did you ever find a solution to this issue on PgBouncer 1.17?

rzavadski-ms:
any updates here?

ziegenberg:
Can this PR be considered a viable solution for #695?

dashorst:
> Can this PR be considered a viable solution for https://github.com/pgbouncer/pgbouncer/issues/695?

yes

adamCo1:
Hey, is there an expectation for the PR to be merged?

cyberbudy:
That's great news, guys!
Waiting this feature so much
Tested this with diesel on Rust and had the same error as @thecodingrobot

sboschman:
Running pgbouncer with `-vv` gives some additional 'NOISE' logging, which might help to determine at which point stuff breaks.

Executing the testcase with standard pg (running as a local container) seems to work fine:
```
$ RUST_LOG=debug cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.05s
     Running `target/debug/tokyo-test`
[2023-03-30T15:53:58Z DEBUG tokio_postgres::prepare] preparing query s0: SELECT $1::TEXT
[2023-03-30T15:53:58Z DEBUG tokio_postgres::query] executing statement s0 with parameters: ["hello world"]
```

kotik:
Hey, can I help somehow to get this patch finalized and merged?

src/messages.c
```diff
@@ -0,0 +1,278 @@
...
+
+bool is_close_statement_packet(PgClosePacket *close_packet)
+{
+  return close_packet->type == 'S' && strlen(close_packet->name) > 0;
+}
+
+PktBuf *create_parse_packet(char *statement, PgParsePacket *pkt)
+{
+  PktBuf *buf;
+  buf = pktbuf_dynamic(pkt->len - strlen(pkt->name + strlen(statement)));
```
knizhnik, line 222:
```suggestion
  buf = pktbuf_dynamic(pkt->len - strlen(pkt->name) + strlen(statement));
```

src/client.c
```diff
@@ -1052,8 +1152,67 @@ bool client_proto(SBuf *sbuf, SBufEvent evtype, struct MBuf *data)
...
+        }
+        break;
+      case CB_REWRITE_PACKET:
+      {
+        unsigned avail = mbuf_avail_for_read(data);
+        if (first) {
+          client->packet_cb_state.pkt_offset = 0;
+        }
+      
+        res = handle_partial_client_work(client, client->packet_cb_state.pkt_header, client->packet_cb_state.pkt_offset, data);
```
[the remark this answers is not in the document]
knizhnik: I have trier to fix this problem: hand bind message in the same way as prepare.
You can find my version here: git@github.com:knizhnik/pgbouncer.git

src/ps.c
```diff
@@ -0,0 +1,336 @@
...
+		initialized = true;
+		rand_seed = csrandom();
+	}
+
+  s = (PgParsedPreparedStatement *)malloc(sizeof(*s));
+  s->name = parse_packet->name;
+  s->pkt = parse_packet;
+  s->query_hash[0] = rand_seed;
+	s->query_hash[1] = 0;
+	spookyhash(parse_packet->query, strlen(parse_packet->query), &s->query_hash[0], &s->query_hash[1]);
```
knizhnik, line 23: I wonder why don't you take in account hash collisions here?
Yes, probability of collision of spooky hash is assumed to be quite small, but still it is non-zero.
And if it happens, then pgbouncer locates wrong query and application receives wrong result.

src/ps.c
```diff
@@ -0,0 +1,336 @@
...
+
+static bool register_prepared_statement(PgSocket *server, PgServerPreparedStatement *stmt)
+{
+  struct PgServerPreparedStatement *current, *tmp;
+  PktBuf *buf;
+  unsigned int cached_query_count = HASH_COUNT(server->server_prepared_statements);
+
+  if (cached_query_count >= (unsigned int)cf_prepared_statement_cache_queries)
+  {
+    HASH_SORT(server->server_prepared_statements, bind_count_sort);
```
knizhnik, line 57: Such replacement algorithm will cause throwing away newly added statements just because their bond_count  is small.
It will be better to use Clock or LRU replacement algorithm.
Also it will help to eliminate sort overhead on each new item which may become bottleneck for large number of prepared statements.

arianf:
@dashorst If I wanted to turn on prepared statements purely for security reasons, would it be possible to set `prepared_statement_cache_queries` to `0` or `1`?

Seems like memory footprint would become negligible, but would there still be an increase in CPU usage?

JelteF:
@arianf You should be able to use anonymous prepared statements just fine with PgBouncer already. Those are fine for security reasons. See this blog for a detailed explanation: https://jpcamara.com/2023/04/12/pgbouncer-is-useful.html#prepared-statements

JelteF:
@dashorst Thank you for all this work, I finally looked at it and really like the overall design. I left some comments on #845, which was a follow up PR on this one by @knizhnik, that contains a few small fixes/improvements. If you still feel like working on this further, I'd very much appreciate if you respond on that PR. If both you and @knizhnik want to work on it, we should make sure to coordinate so that both no-one is doing double work. I considered copy pasting the comments I made to that #845 to this PR too, but that seemed like it could only lead to split discussions, so I refrained from doing that.

Again thanks for your great work on this initial implementation!

JelteF:
@sboschman @dashorst 
On the PR #845 @sboschman wrote this:
> Fix:
> Instead of direct malloc/free calls the Slab memory store from libusual is used. After a recent os update (glibc maybe?) we sometimes see that pgbouncer keeps claiming memory from the os (until OOM). It seems like freed memory is never again reused for a future malloc call, and malloc keeps requesting more and more memory from the os.
>
>Feat:
> The admin console has two new commands to show some parse/bind statistics for client and server sockets. Which at least give a starting point to reason about how effect the ps caching is for a workload.

I think both those changes make sense. I checked the patch repo that you shared: https://github.com/topicusonderwijs/pgbouncer-ps-patch But the license there is different than the PgBouncer one (or at least its ammended). Would it be possible to update this PR to contain your latest changes that are made on the patch repo? That way there's definitely no licensing weirdness. And I also think it will make it easier to bring #845 up to date with these new changes.

letrthang:
this mean that until this fix has been done. we still need to config as below when connects to PgBouncer ?

```
jpa:
    open-in-view: false
    properties:
      hibernate:
        cache:
          use_query_cache: false
        default_schema: navigation_schema_service
```

JelteF:
To avoid confusion, I'm going to close this in favor of #845

PR #764, opened 2022-10-08
Introduce auth_dbname as an option

goksgie:
This commit implements the auth_dbname option introduced by the following PR: https://github.com/pgbouncer/pgbouncer/pull/645

In essence, users should be able to specify an authentication database to execute `auth_query` against. This will help them to avoid creating the same function on every database on their server.

A user can specify `auth_dbname` in the connection string, or they can set it globally. In case `auth_dbname` is specified in the connection string (under database) it will override the global configuration.

Thank you to the original contributors (@gbartolini - Gabriele Bartolini and @leonardoce - Leonardo Cecchi) for implementing this feature.


gbartolini:
Hi thanks, just to mention that the original PR was co-authored by Leonardo Cecchi as well. Please make sure you mention him (more than me) in the PR. Thanks.

goksgie:
Hey @gbartolini ! Thank you for pointing that out. I have updated the description. By the way, please feel free to use this PR if needed for reviving the original PR.

src/loader.c
```diff
@@ -127,6 +127,31 @@ static char * cstr_get_pair(char *p,
...
+static bool set_auth_dbname(PgDatabase *db, const char *new_auth_dbname)
+{
+	const char *old_auth_dbname = db->auth_dbname;
+
+	if (old_auth_dbname && new_auth_dbname) {
+		if (strcmp(old_auth_dbname, new_auth_dbname) == 0) {
+			return true;
+		}
+
+		free((void *)old_auth_dbname);
```
JelteF, line 139: nit: no need for this cast
```suggestion
		free(old_auth_dbname);
```

src/loader.c
```diff
@@ -127,6 +127,31 @@ static char * cstr_get_pair(char *p,
...
+		db->auth_dbname = strdup(new_auth_dbname);
+	} else if (new_auth_dbname) {
+		db->auth_dbname = strdup(new_auth_dbname);
+	} else {
+		db->auth_dbname = NULL;
+	}
+
+	if (new_auth_dbname && !db->auth_dbname) {
+		log_error("auth_dbname %s could not be set for database %s", new_auth_dbname, db->name);
```
JelteF, line 148:
```suggestion
		log_error("auth_dbname %s could not be set for database %s, out of memory", new_auth_dbname, db->name);
```

src/client.c
```diff
@@ -35,6 +35,51 @@ static const char *hdr2hex(const struct MBuf *data, char *buf, unsigned buflen)
...
+ *   1) client->db->auth_dbname: Client's target database may alraedy have
+ * 		an auth_database defined. Prefer this when available.
+ *   2) cf_auth_dbname: Global authentication database name.
+ *   3) client->db: Client's target database. 
+ * NOTE: If authentication database is found but it is disabled, then
+ * 		 client will get disconnected. 
+ * Returns:
+ *   auth_database, based on the above preference.
+ */
+PgDatabase *prepare_auth_database(PgSocket *client)
```
JelteF, line 50: I think the many nested ifs make this function quite hard to read. I think readability here would benefit a lot from early returns. pseudo code:
```
if (!client->db->auth_dbname && !cf_auth_dbname)
   return client->db

auth_dbname = client->db->auth_dbname ? client->db->auth_dbname : cf_auth_dbname


auth_db = find_database(auth_dbname)
if (!auth_db)
    return client->db

if auth_db->db_disabled
    disconnect_client
    return NULL
return auth_db;
```

goksgie: Thank you for the suggestion. Updated the code.

src/loader.c
goksgie: The explicit cast seem to be required here, as otherwise we face with following warning message during compilation:

`warning: passing 'const char *' to parameter of type 'void *' discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
                free(old_auth_dbname);`

src/loader.c
goksgie: Done.

src/loader.c
JelteF: Ah, I didn't realize the const qualifier required it. Maybe add a comment about it.

test/test.sh
```diff
@@ -948,6 +948,68 @@ test_auth_user() {
...
+
+	admin "set auth_type='md5'"
+
+	# test if invalid globally set auth_dbname results in fallback to
+	# user's target database. 
+	admin "set auth_dbname='pinvalid'"
+
+	curuser=`psql -X -d "dbname=authdb user=someuser password=anypasswd" -tAq -c "select current_user;"`
+	echo "revert test curuser=$curuser"
+	test "$curuser" = "$curuser" || return 1
```
JelteF, line 964:
```suggestion
	test "$curuser" = "someuser" || return 1
```

test/test.sh
```diff
@@ -948,6 +948,68 @@ test_auth_user() {
...
+	grep -q "Reverting to client database" $BOUNCER_LOG  || return 1
+
+	# test if auth_dbname specified in connection string takes precedence over global setting
+	curuser=`psql -X -d "dbname=pauthz user=someuser password=anypasswd" -tAq -c "select current_user;"`
+	echo "conn string test curuser=$curuser"
+	test "$curuser" = "$curuser" || return 1
```
JelteF, line 970:
```suggestion
	test "$curuser" = "someuser" || return 1
```

test/test.sh
goksgie: oh what a miss. Updated for all comparisons for this test case.

JelteF (approved):

eulerto (commented):
This is not a complete review. I didn't check tests yet.

src/loader.c
```diff
@@ -286,6 +315,10 @@ bool parse_database(void *base, const char *name, const char *connstr)
 		} else if (!!connect_query != !!db->connect_query
 			   || (connect_query && strcmp(connect_query, db->connect_query) != 0))	{
 			changed = true;
+		} else if ((db->auth_dbname && !auth_dbname)
+			   || (!db->auth_dbname && auth_dbname)
+			   || (auth_dbname && strcmp(auth_dbname, db->auth_dbname) != 0)) {
+			changed = true;
```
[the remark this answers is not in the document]
JelteF: That's not done in for the `connect_query` check above either.
IMHO it is quite obvious that you shouldn't rearrange these checks. 

I do think that for consistency these two checks should be done in the same way.
So: `!!auth_dbname != !!db->auth_dbname`. Or honestly even `!auth_dbname != !db->auth_dbname` (the double in the `connect_query` check `!!` seem unecessary)

src/client.c
```diff
@@ -35,6 +35,56 @@ static const char *hdr2hex(const struct MBuf *data, char *buf, unsigned buflen)
...
+ *   1) client->db->auth_dbname: Client's target database may alraedy have
+ * 		an auth_database defined. Prefer this when available.
+ *   2) cf_auth_dbname: Global authentication database name.
+ *   3) client->db: Client's target database. 
+ * NOTE: If authentication database is found but it is disabled, then
+ * 		 client will get disconnected. 
+ * Returns:
+ *   auth_database, based on the above preference. May return NULL when
+ *   				auth_db is disabled.
+ */
```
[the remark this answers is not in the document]
JelteF: > The chosen variable names are very similar (auth_db and auth_dbname)
Any suggestions for better ones? To me they seem quite clear, one is the db object, the other is a name. They are also inline with other functions that we already have like use_client_socket, where we have `db` and `dbname`.

> There are several returns.

Why is this an issue? I suggested the early-return pattern because I thought the initial version was quite hard to follow.

> I added a comment on the other PR https://github.com/pgbouncer/pgbouncer/pull/645 that I'm concerned about a fallback option. IMO it should be predictable. If the authentication database could not be found, it should fail.

I don't have a very strong opinion either way. But failing hard in case of wrong config, indeed has my slight preference.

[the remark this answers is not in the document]
eulerto: The early-return pattern is not an issue per se. I just said this in hopes that someone else could come up with something easier to read. We can always make hard-to-read code easier to read by adding comments.

src/loader.c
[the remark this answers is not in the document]
JelteF: I agree that the current code isn't great to read. An improvement would be to have a dedicated function for it like `strings_equal(char *left, char *right)` that can handle `NULL.

src/client.c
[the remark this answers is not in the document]
goksgie: To confirm, is the ask here to make `auth_dbname` a breaking change between versions that when not set (in connection string or globally), it will cause disconnection?

The current behavior in that case is to use `dbname` in the connection string.

src/janitor.c
```diff
@@ -729,6 +729,11 @@ void kill_database(PgDatabase *db)
 	} else {
 		statlist_remove(&database_list, &db->head);
 	}
+	
+	if (db->auth_dbname) {
```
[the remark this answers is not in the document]
goksgie: Mainly added to avoid tiny mistakes that we sometimes see in production, but can remove brackets.

src/client.c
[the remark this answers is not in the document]
JelteF: No, (afaik) the ask is: if it is set, but the database is not defined, it should error. Right now it it falls back to use the `dbname` in that case.

So keep this:
```c
	/* use the client's target database if auth_dbname is not configured. */
	if (!auth_dbname) {
		return auth_db;
	}
```
but remove this
```c
	/* if auth_dbname is configured, attemp to find it.
	 * when database is found but it is disabled, it
	 * will result in client disconnection.
	 */
	auth_db = find_database(auth_dbname);
	if (!auth_db) {
		slog_info(client, "Authentication database %s could not be found. Reverting to client database.", auth_dbname);
		return client->db;
	}
```

JelteF (changes requested):
for clarity marking as request changes because of the feedback from @eulerto

src/loader.c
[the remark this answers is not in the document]
goksgie: Thank you for the discussion and suggestions. Added `strings_equal` function that covers `NULL` case.

src/loader.c
```diff
@@ -222,7 +249,9 @@ bool parse_database(void *base, const char *name, const char *connstr)
 			password = val;
 		} else if (strcmp("auth_user", key) == 0) {
 			auth_username = val;
-		} else if (strcmp("client_encoding", key) == 0) {
+		} else if (strcmp("auth_dbname", key) == 0) {
+			auth_dbname = val;
+		}else if (strcmp("client_encoding", key) == 0) {
```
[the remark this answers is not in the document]
goksgie: thank you for pointing that, reverted.

src/loader.c
```diff
@@ -127,6 +127,32 @@ static char * cstr_get_pair(char *p,
...
+{
+	const char *old_auth_dbname = db->auth_dbname;
+
+	if (old_auth_dbname && new_auth_dbname) {
+		if (strcmp(old_auth_dbname, new_auth_dbname) == 0) {
+			return true;
+		}
+
+		/* The cast here is required to strip const qualifier to avoid build warnings. */
+		free((char *)old_auth_dbname);
```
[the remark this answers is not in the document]
goksgie: removed `old_auth_dbname` and reiterated over the function.

src/janitor.c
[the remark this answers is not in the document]
goksgie: removed brackets covering one line.

src/client.c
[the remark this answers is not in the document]
goksgie: removed the logic of falling back. Instead, it will disconnect the client if `auth_db` cannot be found.

etc/pgbouncer.ini
```diff
@@ -122,6 +127,9 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; must have 2 columns - username and password hash.
 ;auth_query = SELECT usename, passwd FROM pg_shadow WHERE usename=$1
 
+;; Authentication database that can be set globally to run "auth_query".
+;auth_dbname
```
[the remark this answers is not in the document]
goksgie: Thank you for pointing that out! Added the `"`.

etc/pgbouncer.ini
```diff
@@ -27,6 +27,11 @@
 ;; auth_user must exist in auth_file
 ; foodb = auth_user=bar
 
+;; use auth_dbname to run auth_query on a specific database.
+;; when auth_dbname is set in connection string, it will override
+;; the global setting.
+; bardb = auth_dbname=foo
```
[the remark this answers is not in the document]
goksgie: Thank you, applied the suggested change.

src/loader.c
```diff
@@ -127,6 +127,35 @@ static char * cstr_get_pair(char *p,
 	return cstr_skip_ws(p);
 }
 
+static bool strings_equal(const char *str_left, const char *str_right)
+{
+	if (!str_left != !str_right)
+		return false;
+	
+	return !str_left == !str_right || strcmp(str_left, str_right) == 0;
```
JelteF, line 135: I think this function should be optimized for readability a bit more instead of for succinctness.
```suggestion
	if (str_left == NULL && str_right == NULL)
		return true;
	
	if (str_left == NULL || str_right == NULL)
	    return false;
	
	return strcmp(str_left, str_right) == 0;
```

src/loader.c
```diff
@@ -127,6 +127,35 @@ static char * cstr_get_pair(char *p,
 	return cstr_skip_ws(p);
 }
 
+static bool strings_equal(const char *str_left, const char *str_right)
```
JelteF, line 130: Can you add a comment? Something like this
```suggestion
/* Same as strcmp, but handles NULLs. If both sides are NULL this returns true */
static bool strings_equal(const char *str_left, const char *str_right)
```

goksgie: Done with commit [a111808](https://github.com/pgbouncer/pgbouncer/pull/764/commits/a111808a8910e8a14eb0eb57992d198cb0f75c26)

src/loader.c
goksgie: done with commit [a111808](https://github.com/pgbouncer/pgbouncer/pull/764/commits/a111808a8910e8a14eb0eb57992d198cb0f75c26)

src/loader.c
```diff
@@ -127,6 +127,41 @@ static char * cstr_get_pair(char *p,
 	return cstr_skip_ws(p);
 }
 
+/*
+ * Same as stcmp, but handles NULLs. If both sides are NULL, returns "true".
```
JelteF, line 131: small typo
```suggestion
 * Same as strcmp, but handles NULLs. If both sides are NULL, returns "true".
```

JelteF (approved):
LGTM

src/loader.c
goksgie: done with commit [de31d44](https://github.com/pgbouncer/pgbouncer/pull/764/commits/de31d44a80be844fd3ed6816420b152ed2483fb9)

JelteF:
@eulerto did you want to look at this once more? Or are you okay with merging it now?

eulerto (approved):
I have no additional comments.

kotik-adjust:
Hey, wondering when you're gonna release the new version of pgbouncer with this feature. :)

3nprob:
Cheers everyone involved for getting this merged <3

PR #788, opened 2022-12-05
Rename wait_cancels state to being_canceled

JelteF:
After merging #717 it was noted that the `wait_cancels` name was quite
confusing. This changes that name to `being_canceled`. This makes it
more clear that this is not a server that is canceling itself, but
instead it's the target of a cancellation request.

Another option would be to rename it to waiting_cancel to be symetric 
with the client state. However, not being symetric is kind of the point 
of the newly suggested name. 

sv_wait_cancel would suggest that it's waiting for a cancel that it sent itself:
- active_cancel_req_list, is a list of clients that are sending a cancel themselves
- waiting_cancel_req_list, is a list of clients that are waiting for a server connection needed to for the cancel
- active_cancel_server_list, is the list of server connections on which a cancel is currently being sent

Thus (to me) waiting_cancel_server_list would suggest that any server in this list is waiting to send a cancel. But that is not what it means to be in this list. Servers in the list in question are currently being cancelled by another server connection. To make this difference clear is why I suggested `being_cancelled_server_list`.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -253,7 +253,7 @@ database
 state
 :   State of the pgbouncer server connection, one of **active**,
     **idle**, **used**, **tested**, **new**, **active_cancel**,
-    **wait_cancels**.
+    **being_canceled**.
 
 addr
 :   IP address of PostgreSQL server.
@@ -394,7 +394,7 @@ sv_active
 sv_active_cancel
 :   Server connections that are currently forwarding a cancel request.
 
-sv_wait_cancels
+sv_being_canceled
 :   Servers that normally could become idle but are waiting to do so until
     all in-flight cancel requests have completed that were sent to cancel
     a query on this server.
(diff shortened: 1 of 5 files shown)


JelteF:
Another name that comes to mind with a similar vibe is `cancel_target`

JelteF:
Not being symetric is kind of the point of the newly suggested name. 

sv_wait_cancel would suggest that it's waiting for a cancel that it sent itself:
- active_cancel_req_list, is a list of clients that are sending a cancel themselves
- waiting_cancel_req_list, is a list of clients that are waiting for a server connection needed to for the cancel
- active_cancel_server_list, is the list of server connections on which a cancel is currently being sent

Thus (to me) waiting_cancel_server_list would suggest that any server in this list is waiting to send a cancel. But that is not what it means to be in this list. Servers in the list in question are currently being cancelled by another server connection. To make this difference clear is why I suggested `being_cancelled_server_list`.

petere (approved):
Difficult to find good brief terms, but this seems fine to me.

PR #789, opened 2022-12-08
Add a python testing framework

JelteF:
Testing in bash is very error prone and the tests are hard to review.
This is an initial and unfinished attempt at adding a python based
testing framework for PgBouncer.


.cirrus.yml
```diff
@@ -34,12 +34,12 @@ task:
...
-      env:
-        use_valgrind: yes
-        CFLAGS: -O0 -g
-        PGVERSION: 9.6
+    # - container:
+    #     image: ubuntu:20.04
+    #   env:
+    #     use_valgrind: yes
+    #     CFLAGS: -O0 -g
+    #     PGVERSION: 9.6
```
JelteF, line 42: @petere what do you think, 9.6 is EOL for more than a year now. It seems reasonable to me to stop officially supporting it (and thus running tests for it). Stopping to support it would allow to make the scram tests always enabled, since any supported PG version has scram support.

petere: Maybe. Is it a problem?

petere:
Very interesting!

PR #793, opened 2022-12-20
Run make in CI in parallel

JelteF:
To speed up compilation. All CI machines have 2 or 4 cores, it's unclear
if those numbers include hyperthreads or not. For simplicity we start
using 4 threads everywhere. In my experience it's much worse if you less
parallelism than can be handled, but having a bit more doesn't matter
too much for the total compilation time. Also if any of these tasks are
I/O bound instead of CPU bound, then you want more parallelism than your
amount of cores anyway.

As an example of speedup this improves the speed of the build step 
on windows CI from 3 minutes to 2 minutes.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -28,7 +28,5 @@ pgbouncer_1.md: filter.py frag-usage-man.md usage.md
 pgbouncer_5.md: filter.py frag-config-man.md config.md
 	$(PYTHON) $^ >$@
 
-.INTERMEDIATE: pgbouncer_1.md pgbouncer_5.md
-
 web:
 	$(MAKE) -C ../../pgbouncer.github.io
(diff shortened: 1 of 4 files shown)


doc/Makefile
```diff
@@ -28,7 +28,5 @@ pgbouncer_1.md: filter.py frag-usage-man.md usage.md
 pgbouncer_5.md: filter.py frag-config-man.md config.md
 	$(PYTHON) $^ >$@
 
-.INTERMEDIATE: pgbouncer_1.md pgbouncer_5.md
```
JelteF, line 31: In CI for windows these files were being cleaned up before pandoc was done using them. Since windows is one of the slowest platforms to compile on, removing this seemed worth the advantage of being able to compile in parallel.

eulerto:
LGTM. Did you observe regression with the other platforms? Out of curiosity, did you experiment with 2 cores? Regarding the `doc/Makefile` change, my suggestion is to use [.NOTPARALLEL](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html) to avoid the issue you described.

JelteF:
I just tried `.NOTPARALLEL` by adding the following to the makefile:
```make
.INTERMEDIATE: pgbouncer_1.md pgbouncer_5.md
.NOTPARALLEL: .INTERMEDIATE pgbouncer_1.md pgbouncer_5.md pgbouncer.1 pgbouncer.5
```
But it didn't help and the windows build still failed: https://cirrus-ci.com/task/5095241439510528

So I'm removing that commit again.

PR #797, opened 2022-12-23
Update SCRAM using Postgres 15.1

eulerto:
Since SCRAM support as copied from Postgres 11.5, SCRAM infrastructure changed quite a bit. This update uses Postgres 15.1 as base. New files need to be added due to code rearrangement.

I split this work in 3 steps:

1. update `etc/import-common.sh`. It adds additional files that needs to be copied for this update.
2. copy the original files from PostgreSQL source code. It uses minor version 15.1 (latest stable minor version). As is, it overrides the changes made by previous commits. The required changes are reintroduced in the step 3.
3. apply the required changes. Some previous logic was kept but a few were changed.

Although Postgres mixes some header files (such as `sha2_int.h`) with *.c files, I moved all header files to include/common and updated the references to them. The reason is that `Makefile` does not include the current directory.

After copying (over) the required files, a few changes are required to compile the code. The challenge was to apply minimal changes to make it work. `system.h` and `postgres_compat.h` does the trick plus changing some header file references that is now in the `include/common` directory. Makefile was updated to include the new required files.

TODO:

- [ ] avoid new files if possible
- [ ] fix compiler warnings
- [ ] document that this update is based on Postgres 15.1
- [ ] `scram_XXX` functions should use `errstr` for error handling


src/scram.c
```diff
@@ -1081,8 +1087,8 @@ scram_verify_plain_password(PgSocket *client,
 		password = prep_password;
 
 	/* Compute Server Key based on the user-supplied plaintext password */
-	scram_SaltedPassword(password, salt, saltlen, iterations, salted_password);
-	scram_ServerKey(salted_password, computed_key);
+	scram_SaltedPassword(password, salt, saltlen, iterations, salted_password, &errstr);
+	scram_ServerKey(salted_password, computed_key, &errstr);
```
JelteF, line 1091: Shouldn't we be doing something with errstr?

eulerto: Yes. I forgot to add a TODO. A few error/warning messages was added to these `scram_XXX` functions in Postgres and we should do the same in PgBouncer as well.

JelteF:
Do you still want to work on this?

eulerto:
It is on my list for some time. I will post an updated PR soon.

PR #804, opened 2023-01-30
Add Python style checks

petere:
Now that we have a lot more Python code, add some project-wide style checking configuration, using flake8.

A couple of easy violations are fixed here, but some that are more difficult are just disabled and might be addressed later (or never).


petere:
> I think the next step would be to add black too: https://github.com/psf/black

Didn't know that one, but looks interesting.

Issue #770, opened 2022-10-20
Deployment on Heroku: add options to handle SIGTERM


PetrDlouhy:
I am using PGBouncer on Heroku in Docker container along with Gunicorn/Nginx/Supervisor.

When Heroku scales down it first sends SIGTERM to all processes in the container and after 30 seconds it sends SIGKILL. I don't think, that I can change this behavior.

The problem is, that at the time of SIGTERM being send, there are still requests being processed. I am receiving following error during scaledown event:
```
TransactionManagementError
An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.
```
It happens because PGBouncer gets terminated before the remaining requests finish being processed.

I wonder if there could be added (or is) any possibility to block/redirect SIGTERM (and let it wait for the SIGKILL or let me terminate it by other process) or add some waiting time before PGBouncer shutdown.

PetrDlouhy:
Just if anyone is searching for workaround solution, I solved this by patching PGBouncer and compiling from source:
```shell
curl -L https://github.com/pgbouncer/pgbouncer/releases/download/pgbouncer_1_17_0/pgbouncer-1.17.0.tar.gz -o pgbouncer.tar.gz \                                                       
tar -xvzf pgbouncer.tar.gz \                                                                                                                                                         
cd pgbouncer-1.17.0 \                                                                                                                                                                
sed -i "s/got SIGTERM, fast exit/got SIGTERM, do nothing/" src/main.c \                                                                                                              
sed -i "s/\texit(1);$//g" src/main.c \                                                                                                                                                
./configure \                                                                                                                                                                        
make \                                                                                                                                                                               
make install
```

petere:
You should use the `SIGINT` signal for gentle shutdown.

PetrDlouhy:
@petere I just can't. It is not possible to change the shutdown process on Heroku.

aaabramov:
> @petere I just can't. It is not possible to change the shutdown process on Heroku.

How about changing the root process to Dockerfile that handles SIGTERM and sends SIGINT to pgbouncer?

PetrDlouhy:
Unfortunately Heroku sends SIGTERM to ALL processes in the Docker container.

I sent them an ticket about this, but so far they are refusing to make any change. I am trying to convince them to change this behavior, but I not giving it much change of success.

eulerto:
[...]

PR #806, opened 2023-02-05
Add option to ignore SIGTERM

mjayasuriya:
This is useful for running in cloud environments like AWS & Heroku where TERM is sent to all processes expecting them to shutdown gracefully.  AWS stops routing requests to these instances when TERM is sent, so pgbouncer can serve any requests in-flight and then terminate once the container exits.

https://github.com/pgbouncer/pgbouncer/issues/770

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -212,6 +212,15 @@ If you need to specify multiple values, use a comma-separated list (e.g.
 
 Default: empty
 
+### ignore_sigterm
+
+Ignore any received TERM signals, useful for running in cloud environments like Herko & AWS which
+send TERM signals on container shutdown while rerouting new requests.  Ignoring the signal allows
+process to respond to last queries instead of terminating immediately.  AWS provides 30s after sending
+TERM to finish responding to in-flight requests.
+
+Default: 0
+
 ### disable_pqexec
 
 Disable the Simple Query protocol (PQexec).  Unlike the Extended Query protocol, Simple Query
--- a/etc/pgbouncer.ini
+++ b/etc/pgbouncer.ini
@@ -161,6 +161,9 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; packet.  Newer JDBC versions require the extra_float_digits here.
 ;ignore_startup_parameters = extra_float_digits
 
+;; If set then ignore TERM signals and continue to serve
+;ignore_sigterm = 0
+
 ;; When taking idle server into use, this query is run first.
 ;server_check_query = select 1
 
--- a/src/main.c
+++ b/src/main.c
@@ -138,6 +138,7 @@ char *cf_resolv_conf;
 unsigned int cf_max_packet_size;
 
 char *cf_ignore_startup_params;
+int  cf_ignore_sigterm;
 
 char *cf_autodb_connstr; /* here is "" different from NULL */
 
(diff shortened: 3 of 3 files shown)


JelteF:
What AWS service are you using that sends SIGTERM? Does that service that you use allow you to change the signal to SIGINT? Because that would solve the issue you're seeing as described in #770 

In any case completely ignoring TERM seems like the wrong solution. It would be better to have the SIGINT behaviour for SIGTERM in this case, than not do anything at all.  So an option like `pause_on_sigterm` would be better

mjayasuriya:
> What AWS service are you using that sends SIGTERM? Does that service that you use allow you to change the signal to SIGINT? Because that would solve the issue you're seeing as described in #770
> 
> In any case completely ignoring TERM seems like the wrong solution. It would be better to have the SIGINT behaviour for SIGTERM in this case, than not do anything at all. So an option like `pause_on_sigterm` would be better

Thanks for the feedback!

We're using auto-scaling EC2 instances, I don't think AWS gives us control over the shutdown process - it broadcasts a TERM to all processes on the box and then 30 seconds later kills them.

One issue with `pause_on_sigterm` is we'd like to continue getting the resiliency pgbouncer provides during that 30s window.  We run pgbouncer in the same container as our worker, so when AWS is down-scaling they'll stop sending requests to that instance and send TERM - but our worker might be generating a report or some other work that requires multiple DB queries.  So having the reconnect & pooling functionality of pgbouncer for those 30s would be helpful to the worker.

What do you think?  The other guy in https://github.com/pgbouncer/pgbouncer/issues/770 also reported issues with Heroku & TERM signal - so probably more cloud providers operate this way.

JelteF:
> it broadcasts a TERM to all processes on the box and then 30 seconds later kills them.

It doesn't seem to actually do that: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#what-happens-terminate 

I'm guessing the behaviour you're seeing happens indirectly, because some part of the OS does that. I'd expect this part to be systemd or docker. Both of these can be configured to send different signals on stop. `KillSignal` for systemd and `STOPSIGNAL` for docker.

Just to clarify your setup: Are you manually running docker containers on EC2 VMs?

JelteF:
> We run pgbouncer in the same container as our worker, so when AWS is down-scaling they'll stop sending requests to that instance and send TERM - but our worker might be generating a report or some other work that requires multiple DB queries. So having the reconnect & pooling functionality of pgbouncer for those 30s would be helpful to the worker.

So are you also ignoring SIGTERM in your application? Is your goal simply to let both your application and pgbouncer run as if no SIGTERM was sent? If so, you can probably have script that's the `ENTRYPOINT` catch the SIGTERM signal and don't forward it to pgbouncer and your app. Another option would be to set `STOPSIGNAL` to `SIGHUP`

mjayasuriya:
> > it broadcasts a TERM to all processes on the box and then 30 seconds later kills them.
> 
> It doesn't seem to actually do that: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/terminating-instances.html#what-happens-terminate
> 
> I'm guessing the behaviour you're seeing happens indirectly, because some part of the OS does that. I'd expect this part to be systemd or docker. Both of these can be configured to send different signals on stop. `KillSignal` for systemd and `STOPSIGNAL` for docker.
> 
> Just to clarify your setup: Are you manually running docker containers on EC2 VMs?

One correction here we're running ECS, but I think the same thing applies - docs say it sends TERM to only PID1.  I need to dig in a bit more and figure out what's PID1 and see if it's something like systemd that like you mention can be configured differently.

mjayasuriya:
> > We run pgbouncer in the same container as our worker, so when AWS is down-scaling they'll stop sending requests to that instance and send TERM - but our worker might be generating a report or some other work that requires multiple DB queries. So having the reconnect & pooling functionality of pgbouncer for those 30s would be helpful to the worker.
> 
> So are you also ignoring SIGTERM in your application? Is your goal simply to let both your application and pgbouncer run as if no SIGTERM was sent? If so, you can probably have script that's the `ENTRYPOINT` catch the SIGTERM signal and don't forward it to pgbouncer and your app. Another option would be to set `STOPSIGNAL` to `SIGHUP`

Ya we're catching SIGTERM in our app too.  I did try to make a bash wrapper around pgbouncer that TRAP sigterm, but that didn't work in our environment.  Like the earlier discussion probably the startup system is notifying each running process independently - not going through the process tree.

Thanks for the discussion!  I need to dig into our startup system more and see if there's options there.

JelteF:
Did you figure out what was happening in your environment in the end?

mjayasuriya:
> Did you figure out what was happening in your environment in the end?

Not exactly, but making supervisord pid1 has been working for us!  Then using the signal config in supervisord.

PetrDlouhy:
@mjayasuriya I have provided some way to work around this here: https://github.com/pgbouncer/pgbouncer/issues/770#issuecomment-1285637655
(although it means compiling PgBouncer from source and automatically patching it)

At least for Heroku, I don't think there is another solution until PgBouncer adds some option to handle this.

PR #807, opened 2023-02-07
Autoformat python code

JelteF:
This uses `black` and `isort` to get consistent codestyle in all our python
files in the repo.

It also adds the `flake8-bugbear` plugin to `flake8`

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,3 @@
 [flake8]
-extend-ignore = E265, F401, F541
+extend-ignore = E203, E265, F401, F541
 max-line-length = 150
--- a/test/utils.py
+++ b/test/utils.py
@@ -556,7 +556,7 @@ class Postgres(QueryRunner):
         """Remove any HBA rules that were added after the last call to commit_hba"""
         with self.hba_path.open() as f:
             hba_contents = f.read()
-        committed = hba_contents[hba_contents.find("# committed-rules\n"):]
+        committed = hba_contents[hba_contents.find("# committed-rules\n") :]
         with self.hba_path.open("w") as f:
             f.write(committed)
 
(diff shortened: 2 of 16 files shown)


.flake8
```diff
@@ -1,3 +1,3 @@
 [flake8]
-extend-ignore = E265, F401, F541
+extend-ignore = E203, E265, F401, F541
```
petere, line 2: This is weird.

test/utils.py
```diff
@@ -556,7 +556,7 @@ def reset_hba(self):
         """Remove any HBA rules that were added after the last call to commit_hba"""
         with self.hba_path.open() as f:
             hba_contents = f.read()
-        committed = hba_contents[hba_contents.find("# committed-rules\n"):]
+        committed = hba_contents[hba_contents.find("# committed-rules\n") :]
```
petere, line 559: This is weird. Why would it do that? Do we know?

petere (approved):
Looks good to me, except that one comment.

.flake8
JelteF: It disallows some valid patterns apparently: https://black.readthedocs.io/en/stable/faq.html#why-are-flake8-s-e203-and-w503-violated

test/utils.py
JelteF: Putting whitespace around the slice operator is recommended by pep8 for complex expressions: https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#slices

Apparently black considers this function call complex enough to trigger adding a single space. 

I don't feel strongly either way. And it's not possible to configure this behaviour differently. Based on your comment it seems like you don't feel strong about it either, mostly finding the change unexpected. So I'll merge the PR.

PR #811, opened 2023-03-01
Don't process buffered payload when EV_CLOSED is detected

bonnefoa:
Some drivers close the connection to PgBouncer on timeout ([pq-timeout](https://github.com/Kount/pq-timeouts) for example).
When PgBouncer is loaded (due to churning TLS connections for example), this can lead to sockets piling up with buffered query and already closed by the client.

```
CLOSE-WAIT 52     0      10.12.16.24:5433   10.12.16.199:48878 users:(("pgbouncer",pid=2047465,fd=629)) timer:(keepalive,119min,0)
CLOSE-WAIT 52     0      10.12.16.24:5433   10.12.16.199:41310 users:(("pgbouncer",pid=2047465,fd=537)) timer:(keepalive,119min,0)
CLOSE-WAIT 52     0      10.12.16.24:5433   10.12.16.199:48910 users:(("pgbouncer",pid=2047465,fd=633)) timer:(keepalive,119min,0)
CLOSE-WAIT 52     0      10.12.16.24:5433   10.12.16.199:45086 users:(("pgbouncer",pid=2047465,fd=514)) timer:(keepalive,119min,0)
```

PgBouncer will still pop the buffered event, open a server connection, send the query and immediately close the server connection as it's not in a reusable state.
Detecting early closure with EV_CLOSED when it's available can help recover from this situation by avoiding wasting time on queries that are not gonna be used.

### DIFF
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -33,6 +33,13 @@
 #define USE_TLS
 #endif
 
+/* EV_CLOSED was introduced in libevent 2.1
+ * Centos7 is using libevent 2.0 and won't have it defined
+ */
+#ifndef EV_CLOSED
+#define EV_CLOSED 0
+#endif
+
 /* sbuf_main_loop() skip_recv values */
 #define DO_RECV		false
 #define SKIP_RECV	true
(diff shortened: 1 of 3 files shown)


JelteF:
In general this change seems reasonable. But it's causing some existing tests to fail with a connection timeout now. Also on compilation fails on CentOS 7.

bonnefoa:
Yeah, I just saw the failures. I will work on fixing those.

Issue #778, opened 2022-11-11
pgbouncer forwards different SQLState code than PostgreSQL if database does not exist


sleepy-manul:
We have a development environment where lots of PostgreSQL databases are created and dropped all day long, e.g.

```
somedb_branch_PRODUCT-5875
somedb_branch_PRODUCT-6897
somedb_branch_PRODUCT-8292
```

etc. To accommodate this, I set up pgBouncer like this:

``` 
[databases]
*=host=fqdn.of.our.postgresql.instance port=5432

[pgbouncer]
listen_port=6432
listen_addr=0.0.0.0
auth_file=/opt/bitnami/pgbouncer/conf/userlist.txt
auth_type=plain
pidfile=/opt/bitnami/pgbouncer/tmp/pgbouncer.pid
logfile=/opt/bitnami/pgbouncer/logs/pgbouncer.log
admin_users=admin@user-name-of-the-pgsql-instance
client_tls_sslmode=disable
server_tls_sslmode=verify-ca
server_tls_ca_file=/etc/ssl/certs/ca-certificates.crt
server_tls_ciphers=fast
server_tls_protocols=secure
```

The problem is that the PostgreSQL server (direct connection, without pgBouncer) returns this if the database does not exist yet:
```[3D000] FATAL: database "postgres2" does not exist.```

However, when going through pgBouncer, a different SQLState is returned:
```Npgsql.PostgresException (0x80004005): 08P01: database "postgresql2" does not exist ```
(the SQLState here is ```08P01```).

This is a problem for tools like .NET Entity Framework in connection with the npgsql driver, which relies on the SQLState to check if the database already exists. Example:
https://github.com/npgsql/efcore.pg/blob/main/src/EFCore.PG/Storage/Internal/NpgsqlDatabaseCreator.cs#L269

When ```08P01``` is returned instead of ```3D000```, the driver cannot recognise that the database does not exist yet, and thus the start of the application fails.

Due to the number of DBs and the frequency of creates/drops, it is not feasible to create the DBs manually. Also, it would require a lot of effort to change our CI/CD pipeline to create the DBs before the deployment, because our application needs several databases, and different product versions use different databases etc.

Can you change the reply that pgBouncer sends to the client so that in the case of a missing DB, SQLState ```3D000``` is sent instead of ```08P01```?

eulerto:
[...]

PR #814, opened 2023-03-06
Provide SQLSTATE

eulerto:
PgBouncer currently uses SQLSTATE 08P01 (protocol_violation) as default error code to send error messages to client. There are cases that this SQLSTATE is different from what Postgres provides. Since the goal is to be as transparent as possible, PgBouncer should report the SQLSTATE provided by Postgres. Issue #778 exposes one case that if you connect to a database that doesn't exist, Postgres reports 3D000 (invalid_catalog_name) but PgBouncer reports 08P01. The npgsql driver relies on the SQLSTATE to check if the database already exists and this behavior breaks it. PgBouncer will use the SQLSTATE that Postgres provides.

Fix issue #778.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1072,17 +1072,16 @@ void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...
 }
 
 /*
- * close client connection
+ * A wrapper around disconnect_client_sqlstate()
  *
- * notify=true means to send the reason message as an error to the
- * client, notify=false means no message is sent.  The latter is for
- * protocol and communication errors where sending a regular error
- * message is not possible.
+ * The function disconnect_client_sqlstate() inherits the disconnect_client()
+ * content and add a new option that provides a specific SQLSTATE that is
+ * forwarded to client.  PgBouncer used to report SQLSTATE 08P01
+ * (protocol_violation) for all cases but it diverges from what Postgres
+ * reports in some cases.
  */
 void disconnect_client(PgSocket *client, bool notify, const char *reason, ...)
 {
-	usec_t now = get_cached_time();
-
 	if (reason) {
 		char buf[128];
 		va_list ap;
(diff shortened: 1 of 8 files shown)


src/objects.c
```diff
@@ -963,17 +963,16 @@ void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...
...
+ * content and add a new option that provides a specific SQLSTATE that is
+ * forwarded to client.  PgBouncer used to report SQLSTATE 08P01
+ * (protocol_violation) for all cases but it diverges from what Postgres
+ * reports in some cases.
  */
 void disconnect_client(PgSocket *client, bool notify, const char *reason, ...)
 {
-	usec_t now = get_cached_time();
-
 	if (reason) {
```
JelteF, line 976: I think this should be moved to `disconnect_client_sqlstate`.

JelteF (changes requested):
code looks good overall. But I think this should include a regression test. Also there's some compilation warning in CI that should be fixed.

src/objects.c
eulerto: Are you referring to `now` variable? It was moved to `disconnect_client_sqlstate`. I checked a few times after reading the diff. diff seems confusing. `now` is in line 999.

eulerto:
Thanks for checking. I noticed the failure after creating this PR. I will fix that error and create a test case for this issue.

src/objects.c
JelteF: no I meant the `if (reason) {` block

eulerto:
I tried to come up with a test case but it seems psycopg decided that it doesn't use the _Invalid Catalog Name_ (3D000) for the _'database "foo" does not exist'_ case. According to [psycopg2.errors](https://www.psycopg.org/docs/errors.html), class 3D belongs to **ProgrammingError** exception but a single connection to a database that doesn't exist returns **OperationalError** exception.

```
Traceback (most recent call last):
  File "/tmp/c.py", line 13, in <module>
    con2 = psycopg2.connect("host=localhost port=9915 dbname=nonexist")
  File "/usr/lib/python3/dist-packages/psycopg2/__init__.py", line 127, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
psycopg2.OperationalError: FATAL:  database "nonexist" does not exist
```

Unless, I'm missing some trick, I think a test case won't be possible.

JelteF:
All of the `pscopg` error types have an  `sqlstate` field. If you check that manually to be the expected error code, then I think we have tested it well enough.

JelteF (approved):
Okay I tried to create the test that I had in my mind. But indeed it's not possible. The error code is not exposed to us. And the reason for that is actually because the error result is not exposed to psycopg by libpq.

Did you manually test that this fixes the problem with npgsql? If so, I think we should just merge this. The code looks good, but someone should at least test that this fixes the issue manually.

eulerto:
I manually inspected the messages but didn't give it a try with Npgsql. I pinged the other issue to see if someone tested it or can test it.

JelteF:
If you manually inspected the result that's good enough for me. I'll press the merge button now. In the unlikely case that people on the issue still reach out that it didn't fix the problem, then it's probably a minor change.

PR #817, opened 2023-03-22
Fixed connection with unknown stats_users or admin_users which throws…

### LINKED ISSUE #314: pgbouncer crashes when user is equal to auth_user in database config

wwoytenko:
**Description**

This PR contains a bug fix for not existed stats or admin users when auth_query is turned on

This fixes #314 

**Clarification**

I have found in my environment unexpected behavior when some user is trying to connect to pgbouncer. As a result pgbouncer instance crashed with error `@src/objects.c:316 in function put_in_order(): put_in_order: found existing elem`

Debug setting `verbose=2` shows me that some that this user is `stats`. 
```
2023-03-22 11:49:13.588 EET [162420] NOISE C-0x55e9e01f3360: (nodb)/(nouser)@[::1]:33504 read pkt='!' len=82
2023-03-22 11:49:13.588 EET [162420] DEBUG C-0x55e9e01f3360: (nodb)/(nouser)@[::1]:33504 got var: user=stats
2023-03-22 11:49:13.588 EET [162420] DEBUG C-0x55e9e01f3360: (nodb)/(nouser)@[::1]:33504 got var: database=pgbouncer
2023-03-22 11:49:13.588 EET [162420] DEBUG C-0x55e9e01f3360: (nodb)/(nouser)@[::1]:33504 using application_name: psql - [::1]:33504
2023-03-22 11:49:13.588 EET [162420] DEBUG C-0x55e9e01f3360: (nodb)/(nouser)@[::1]:33504 got var: client_encoding=UTF8
2023-03-22 11:49:13.588 EET [162420] FATAL @src/objects.c:334 in function put_in_order(): put_in_order: found existing elem
```

The psql call is
`psql "host=localhost port=6432 dbname=pgbouncer user=stats sslmode=disable"`

I looked through `pgbouncer.ini` and `userlixt.txt` and found that the stats user was set in `pgbouncer.ini` with the setting `stats_users = stats` but it was not in the `userlixt.txt` file.

So the original behavior (in this case) is:
* Get creds from userlist.txt
* Get cred from auth_query
* Reject the connection if cred was not found or mismatch

It is right when we are connecting to the real database (not admin) or `auth_dbname` was defined and set up in the `[databases]` section. But due to we are connecting to the admin DB we are using it as an `auth_dbname` by default and the source code does not contain any checks about the `admin` flag. Eventually, before the login query we are adding existing db into the pool and it throws `@src/objects.c:316 in function put_in_order(): put_in_order: found existing elem`

To repeat this problem use config

pgbouncer.ini
```
[databases]
;; postgres = host=127.0.0.1 port=5432 dbname=postgres
* = host=127.0.0.1 port=5432

[pgbouncer]
auth_query = SELECT usename, passwd FROM pg_shadow where usename = $1
auth_user = pgbouncer
stats_users = stats
listen_addr = localhost
verbose = 2
admin_users = pgbouncer
auth_file = ./etc/userlist.txt
;; auth_dbname = postgres

```

userlist.txt
```
"pgbouncer" "1234"
```


JelteF:
Thanks for the contribution. Could you add a regression test for this as well?

wwoytenko:
@JelteF Surely I'll do this in a few days

wwoytenko:
@JelteF We have done all regression tests and are ready for review

wwoytenko:
> I found another case when we set up `auth_dbname = pgbouncer` we are still able to crash the pgbouncer instance. Going to add additional checks

The check has been moved to `start_auth_query()`  level - it covers both cases.

src/client.c
```diff
@@ -161,6 +161,12 @@ static void start_auth_query(PgSocket *client, const char *username)
 	if (!auth_db)
 		return;
 
+    if (auth_db->admin) {
```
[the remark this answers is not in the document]
goksgie: Additionally, I wonder if we should also prevent users from setting auth_dbname as `pgbouncer` in the query.

The check we perform on [parsing database](https://github.com/pgbouncer/pgbouncer/blob/master/src/loader.c#L216) name can also apply to [auth_dbname](https://github.com/pgbouncer/pgbouncer/blob/master/src/loader.c#L144). This could avoid explicit else block.

test/test_auth.py
```diff
@@ -201,3 +201,117 @@ def test_scram_both(bouncer):
...
+
+    with bouncer.run_with_config(config):
+        with pytest.raises(psycopg.OperationalError, match="bouncer config error"):
+            # bad password
+            # We expect that stats user does not exist ether in userlist.txt and in postgresql DB
+            bouncer.test(user="stats", password="stats", dbname="pgbouncer")
+
+
+@pytest.mark.skipif("WINDOWS", reason="Windows does not have SIGHUP")
+def test_connection_with_missing_stats_user_and_enabled_auth_query_with_global_auth_dbname(
```
goksgie, line 244: I think we should also add a test where auth_dbname is set correctly and `stats` user is not listed in `userlist.txt` file but exists in PostgreSQL. In this case, authentication should work when the target login database is `pgbouncer`

test/test_auth.py
```diff
@@ -201,3 +201,117 @@ def test_scram_both(bouncer):
...
+def test_connection_with_missing_stats_user_and_enabled_auth_query_without_global_auth_dbname(
+    bouncer,
+):
+    """
+    Regression for https://github.com/pgbouncer/pgbouncer/issues/314
+
+    Check that pgbouncer is not crashing on connection to the virtual pgbouncer virtual db when user is missing in
+    userlists.txt and auth_query enabled.
+
+    It covers case when global auth_dbname does not set and target database used for auth_dbname in auth query
```
goksgie, line 216: Suggested description:

It covers `the` case when auth_dbname is not configured and the target database is used as the authentication database.

test/test_auth.py
```diff
@@ -201,3 +201,117 @@ def test_scram_both(bouncer):
 
     # SCRAM password in userlist.txt
     bouncer.test(dbname="p62", user="scramuser1", password="foo")
+
+
+@pytest.mark.skipif("WINDOWS", reason="Windows does not have SIGHUP")
+def test_connection_with_missing_stats_user_and_enabled_auth_query_without_global_auth_dbname(
```
goksgie, line 207: Do we need to specify `enabled_auth_query` in all test names?

test/test_auth.py
wwoytenko: That's a significant remark I will add this test

test/test_auth.py
```diff
@@ -201,3 +201,117 @@ def test_scram_both(bouncer):
...
+def test_connection_with_missing_stats_user_and_enabled_auth_query_with_global_auth_dbname(
+    bouncer,
+):
+    """
+    Regression for https://github.com/pgbouncer/pgbouncer/issues/314
+
+    Check that pgbouncer is not crashing on connection to the virtual pgbouncer virtual db when user is missing in
+    userlists.txt and auth_query enabled.
+
+    It covers case when global auth_dbname is set globally
```
goksgie, line 253: Suggestion:

It covers the case when auth_dbname is set to `pgbouncer` in the configuration file.

src/client.c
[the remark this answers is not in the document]
wwoytenko: > Could be nicer to have this in `prepare_auth_database`. For instance:
> 
> ```
>   if (!auth_dbname)
> 	auth_db = client->db;
>   else
> 	auth_db = find_database(auth_dbname);
> 
> <other checks>
> 
>    if (auth_db->admin) {
>         slog_error(client, "admin database \"%s\" cannot be an authentication database", auth_db->dbname);
>         disconnect_client(client, true, "bouncer config error");
>         return null;
>    }
> ```

That is a reasonable point, I agree that these checks should be within the `prepare_auth_database` function. I will redo it.

test/utils.py
```diff
@@ -815,3 +815,24 @@ def log_contains(self, re_string, times=None):
             else:
                 match_count = len(re.findall(re_string, content))
                 assert match_count == times
+
+    """
+    Run pgbouncer instance with provided config and restore previous config in after execution
```
goksgie, line 820: Should this description be part of the function?

wwoytenko: Definitely! I'll move it inside the function

test/test_auth.py
wwoytenko: Tries to highlight the meaning point of these cases, but I can make the function names simpler. I made it because only when `auth_query` enabled pgbouncer falls apart. Anyway we may write the error condition in the function description

src/client.c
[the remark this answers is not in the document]
wwoytenko: > Additionally, I wonder if we should also prevent users from setting auth_dbname as `pgbouncer` in the query.
> 
> The check we perform on [parsing database](https://github.com/pgbouncer/pgbouncer/blob/master/src/loader.c#L216) name can also apply to [auth_dbname](https://github.com/pgbouncer/pgbouncer/blob/master/src/loader.c#L144). This could avoid explicit else block.


If I understand you clearly, you meant that if we set this check in the set_auth_dbname we will avoid explicit else in the prepare_auth_database. In my point of view, it will throw the same error, when we use fallback databases. For example, we can write in the config `* = host=127.0.0.1 port=6432` and then try to connect to pgbouncer DB using `psql -h localhost -p 6432 -U stats -d pgbouncer`, it will force the pgbouncer exiting with the same error. I suggest keeping this check for both cases (with explicit else) even if we will fix find another solution to prevent using pgbouncer as auth_dbname.

Because of that, I added an additional test when we set up a fallback DB connection with an admin database (pgbouncer DB).

test/test_auth.py
wwoytenko: I assigned simpler test function names and added additional information in the doc-string

test/test_auth.py
wwoytenko: https://github.com/pgbouncer/pgbouncer/pull/817#discussion_r1149849743
The same as in the previous

test/test_auth.py
wwoytenko: https://github.com/pgbouncer/pgbouncer/pull/817#discussion_r1149849743
The same as in the previous

test/utils.py
wwoytenko: Fixed

test/test_auth.py
wwoytenko: Added tests that I hope cover all cases

src/client.c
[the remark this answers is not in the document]
wwoytenko: We still have two inconvenient behaviors:
* No cf_auth_dbname checks
* No checks for fallback DB

I didn't find a convenient solution for parsing [pgbouncer] section which I could implement with the current parser and it wouldn't be a shame. I don't want to mix parsing logic with kind of business logic. Anyway, I should give a few ideas:

1. We could define new type macros in [cfparser](https://github.com/libusual/libusual/blob/b0bc727b793b6a94902b46d06ad87e70eb2a2e0c/usual/cfparser.h#L178) which describes `auth_dbname` constraints. It would be `#CF_AUTHDB`:
```
#define CF_STR	{ cf_set_authdb, cf_get_authdb }
```
And `cf_set_authdb` ([placed here](https://github.com/libusual/libusual/blob/b0bc727b793b6a94902b46d06ad87e70eb2a2e0c/usual/cfparser.c#L441)) would be a wrapper that gives us additional checks, such as:
```
bool cf_set_str(struct CfValue *cv, const char *value)
{
	if (strcmp(dbname, "pgbouncer") != 0) {
		log_error("cannot use the reserved \"%s\" database as an auth_dbname", new_auth_dbname);
		return false;
	}
	return cf_set_str(cv, value);
}
```
I clearly understand that this is not the way to handle it, because it is dependency.

2. Another option is to add an additional check after file parsing. For example, add a check function
```
bool loader_auth_dbname_check(void)
{
    if (cf_auth_dbname && strcmp(cf_auth_dbname, "pgbouncer") == 0) {
        log_error("cannot use the reserved \"%s\" database as an auth_dbname", cf_auth_dbname);
        free(cf_auth_dbname);
        cf_auth_dbname = NULL;
        return false;
    }
    return true;
}
```
And place additional check [before this line](https://github.com/pgbouncer/pgbouncer/blob/master/src/main.c#L409):
```
if (ok){
   ok = loader_auth_dbname_check();
}
```
It looks weak too.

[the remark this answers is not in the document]
wwoytenko: In my point of view, we have two things that should be implemented somehow:
* configuration parameters constraints
* config reloading with shadow copy approach

Additionally, I suppose we should implement a config testing utility, like `nginx -t`. Because now if we add such constraints (or just made a simple mistake in INI) we may partially override the old settings which may bring unpredictable behavior. I suppose we can implement something in the next PR's but It will bring us huge refactoring.

wwoytenko:
Going to add additional checks and tests soon:
1. Check for cf_auth_dbname 
2. Check for `autodb` that it does not contain param `auth_dbname check` with `"pgbouncer"` value

src/client.c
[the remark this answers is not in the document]
wwoytenko: Added validations that cover:
* config:
  * [pgbouncer] section
  * [databases] section
    * validation for all explicitly defined DB's
    * validation for authdb connection str
* runtime:
  * Check inside the `prepare_auth_database` function

wwoytenko:
@goksgie The PR is ready to review. All fixes mentioned above were added in the last commit.

I should mention that this issue we can meet in any version higher that 1.14, even though I didn’t try to perform the same tests with versions not from master branch. But I almost sure that the root cause is the same.

src/loader.c
```diff
@@ -164,13 +169,42 @@ static bool set_auth_dbname(PgDatabase *db, const char *new_auth_dbname)
...
-	}
+    /* Validating constraints for database connection string */
+    tmp_connstr = strdup(connstr);
+    if (!tmp_connstr) {
+        log_error("no mem to change autodb_connstr");
+        return false;
+    }
+
+    p = tmp_connstr;
+    while (*p) {
```
goksgie, line 185: This loop is not required. The "cf_autodb_connstr" will be given to parse_database function, which will detect that "auth_dbname" is set incorrectly.

src/util.c
```diff
@@ -468,3 +469,15 @@ const char *pga_details(const PgAddr *a, char *dst, int dstlen)
 	}
 	return dst;
 }
+
+/*
+ * parser setter for parsing auth_dbname parameter using "cfparser"
+ */
+bool cf_set_authdb(struct CfValue *cv, const char *value)
+{
+    if (value && strcmp(value, "pgbouncer") == 0) {
```
goksgie, line 478: we make this "pgbouncer" check in multiple places. Can we move it to a shared place, e,g., `check_reserved_database`

src/util.c
```diff
@@ -468,3 +469,15 @@ const char *pga_details(const PgAddr *a, char *dst, int dstlen)
 	}
 	return dst;
 }
+
+/*
+ * parser setter for parsing auth_dbname parameter using "cfparser"
+ */
+bool cf_set_authdb(struct CfValue *cv, const char *value)
```
goksgie, line 476: I do not have strong opinion on the correct placement of this parser. On one hand, most of the parsers are placed in the libusual. On the other hand, we can argue that this is not a primitive type we're parsing, but rather an information that is only relevant to PgBouncer. So, I guess that is fine.

+ I like the idea of having a validation during the startup + configuration reloads, though, would be nice to have a senior PgBouncer engineer's opinion on this. @JelteF , @eulerto  what do you think of this change?

src/client.c
```diff
@@ -68,6 +70,13 @@ PgDatabase *prepare_auth_database(PgSocket *client)
 		return NULL;
 	}
 
+    if (auth_db->admin) {
+        slog_error(client, "cannot use the reserved \"%s\" database as an auth_dbname", auth_db->dbname);
+        slog_warning(client, "set global \"auth_dbname\" parameter and define database in the database section");
```
goksgie, line 75: I do not think there is a need to provide a warning here, as the error message is clearly explaining what is wrong.

src/client.c
```diff
@@ -49,10 +49,12 @@ PgDatabase *prepare_auth_database(PgSocket *client)
 	PgDatabase *auth_db = NULL;
 	const char *auth_dbname = client->db->auth_dbname ? client->db->auth_dbname : cf_auth_dbname;
 
-	if (!auth_dbname)
-		return client->db;
+    if (!auth_dbname) {
```
goksgie, line 52: Please double check the indentations, I think there is a tab/spacing issue.

test/test_auth.py
```diff
@@ -201,3 +202,216 @@ def test_scram_both(bouncer):
...
+                # bad password
+                # We expect that stats user does not exist in userlist.txt
+                bouncer.test(user="stats", password="stats", dbname="pgbouncer")
+
+
+@pytest.mark.skipif("WINDOWS", reason="Windows does not have SIGHUP")
+def test_auth_dbname_usage_using_autodb(
+    bouncer,
+):
+    """Regression for https://github.com/pgbouncer/pgbouncer/issues/314
```
goksgie, line 252: some tests referring to this issue do not perform any login with a stats_user. I would suggest removing this remark from those to avoid confusion in the future.

test/test_auth.py
```diff
@@ -201,3 +202,216 @@ def test_scram_both(bouncer):
...
+
+    Check that the pgbouncer does not apply config which contains
+    explicitly pgbouncer database (admin DB) set in auto-database
+    definition
+    """
+
+    config = f"""
+        [databases]
+        * = host={bouncer.host} port={bouncer.port} auth_dbname=pgbouncer
```
goksgie, line 261: would it be better to remove auth_dbname=pgbouncer from [databases] section. As I think the goal of this test is to see if we can detect incorrect [pgbouncer]/auth_dbname values?

test/test_auth.py
```diff
@@ -201,3 +202,216 @@ def test_scram_both(bouncer):
...
+):
+    """Regression for https://github.com/pgbouncer/pgbouncer/issues/314
+
+    Check that the pgbouncer does not apply config which contains
+    explicitly "pgbouncer" database (admin DB) set in [pgbouncer] section
+    """
+
+    config = f"""
+        [databases]
+        * = host={bouncer.host} port={bouncer.port} auth_dbname=pgbouncer
```
goksgie, line 293: if the goal of this test to see if we can detect [pgbouncer]/auth_dbname misconfiguration, then it would be nicer to avoid providing incorrect auth_dbname in the [databases] section.

goksgie (changes requested):
I think mostly looks good. Though, I think we should remove the added loop in the "cf_autodb_connstr" parser, set_autodb.

Edit: I think I left the review incorrectly 😢 , please check the comments

src/loader.c
wwoytenko: I agree, I will move this check inside the `parse_database` function.

src/util.c
wwoytenko: Great point, I will do it soon

src/client.c
wwoytenko: Got it. I'll keep only the error statement

src/client.c
wwoytenko: As I understand we must use tab indentation, if so I have to check the all changes in the PR. I'll do this in the next commit

test/test_auth.py
wwoytenko: The current implementation checks the `auth_dbaname` in the `set_autodb` function and had a different parsing scenario in comparison with the common DB definition. Relying on your comment [here](https://github.com/pgbouncer/pgbouncer/pull/817#discussion_r1161149558) I will keep only one test for [databases] section because if we will use only one loop for DBs (even for autodb), we check the same scenario.

src/loader.c
goksgie: I do not think there is a need to add any other check than one you've already added to set_authdb function, which makes pgbouncer comparision. That should be sufficient.

wwoytenko: Referring to the [line](https://github.com/pgbouncer/pgbouncer/blob/8d0ba3bd8e7e0afe10ebf065e52baa3a14666bbf/src/loader.c#L260) it seems that if we move checks within the `parse_database` function we should partially change parsing logic. 

We could implement a similar as below:
```
bool parse_database(void *base, const char *name, const char *connstr)
{
        /* ...................... */
	bool is_autodb = false;
        /* ...................... */
        if (strcmp(name, "*") == 0) {
                is_autodb = true
                if (!set_autodb(connstr)){
                    return false   
                }
	}
        
        /* ...................... */
	while (*p) {
            else if (strcmp("auth_dbname", key) == 0) {
                        /* Added reserved auth_dbname check */
                        if (check_reserved_database(key)}
                            auth_dbname = val;
                        } else {
                            return false;
                        }
            }
       }
       if (is_autodb) {
           // We do not need to execute further code and allocate memory 
           // for autodb using the `add_database` function. Instead, we have 
           // passed through the all connstr definition and returns true (all is valid)
           return true;
       }
        /* ...................... */
}
```

The main point is - autodb is not a real DB in the pgbouncer terms during the config parsing stage so that we store this value into `cf_autodb_connstr` and do not create a real DB, instead we are creating it when autodb connection is matched. According to that, we can make some checks after the loop (is it autodb or not) in order to avoid memory allocation using the `add_database` function.

I would be glad to see any comments about this concern

goksgie: I see your point that this would prevent "*" being **set** incorrectly. However, I do not see any benefit of performing this additional check with extra complexity. Because, connection attempts landing on the autodb will get rejected due to incorrect configuration anyway once the registered connection string is provided to parse_database.

src/util.c
JelteF: I think it makes sense to have it in pgbouncer itself, since indeed it's not a primitive type.

src/loader.c
wwoytenko: @goksgie you took the point across correctly. But in my point of view, it is better when in the config applying stage we would find any misconfiguration issues. Anyway, I will remove this additional check.

wwoytenko: I closed this thread because I reverted the changes

src/client.c
wwoytenko: Done

test/test_auth.py
wwoytenko: Done

src/client.c
wwoytenko: Done

test/utils.py
```diff
@@ -337,17 +337,17 @@ async def asleep_coroutine(self, duration=3, times=1, sequentially=False, **kwar
             for _ in range(times):
                 await self.asql(f"select pg_sleep({duration})", **kwargs)
 
-    def test(self, **kwargs):
+    def test(self, query="select 1", **kwargs):
```
JelteF, line 340: Instead of allowing to pass a query to `test` I think it's better to use the `sql` method directly in the places where you want to use a different query. The intent of the `test` functions is really just to test if the connection is not broken.

test/test_auth.py
```diff
@@ -201,3 +202,149 @@ def test_scram_both(bouncer):
...
+
+            #     Check the pgbouncer does not crash when explicitly pgbouncer database
+            #     (admin DB) was set in auth_dbname in the databases definition section
+            with pytest.raises(psycopg.OperationalError, match="bouncer config error"):
+                bouncer.test(user="stats", password="stats", dbname="pgbouncer_test")
+
+            #     Check the pgbouncer does not crash when explicitly pgbouncer database
+            #     (admin DB) was set in auth_dbname in the databases definition section
+            with pytest.raises(psycopg.OperationalError, match="bouncer config error"):
+                bouncer.test(user="stats", password="stats", dbname="pgbouncer_test")
```
JelteF, line 249: It seems this currently tests the same as the the test above. I think you mean to use a different database name here, so we would trigger the wildcard case.

```suggestion
                bouncer.test(user="stats", password="stats", dbname="some_other_db")
```

test/test_auth.py
JelteF: I think @goksgie meant you should probably do this change. Because now it's unclear if the config apply is failing due to the `auth_dbname=pgbouncer` in the `[databases]` section or the one in the `[pgbouncer]` section. Since the `test_auth_dbname_usage` already tests `auth_dbname=pgbouncer` in the `[databases]` section, it seems good to remove here. 

```suggestion
        * = host={bouncer.host} port={bouncer.port}
```

JelteF (changes requested):
Overall this is very close to merging. Some final feedback. Once that is adressed I think this is good.

test/test_auth.py
wwoytenko: Agree! I provided the `p4` DB so that it now triggers the wildcard case

test/utils.py
wwoytenko: Thank you for pointing it out.
I have just reverted the changes in utils.py in the `QueryRunner` class and implemented tests using `bouncer.sql`

test/test_auth.py
wwoytenko: Thank you so much for your clarification! That's right. I have pushed the changes related to that.

wwoytenko:
Total notes about the last changes:
- Fixed doubled test cases with their description
- Added a few checks for the `test_auth_dbname_works_fine` test
- Reverted changes in QueryBuilder in utils.py

JelteF (approved):

PR #821, opened 2023-03-29
Add intervalstyle parameter

### LINKED ISSUE #470: Allow IntervalStyle parameter (like DateStyle)
IntervalStyle is a similar parameter to DateStyle (controls output format for a data type, passed in the same way to PostgreSQL connections). Currently the [Julia PostgreSQL client library](https://github.com/invenia/LibPQ.jl/blob/master/src/connections.jl) only supports parsing one IntervalStyle format, and it is not the default one, so it would be nice to have support for connecting with this option.

sj26:
It's common for client connection libraries to set `intervalstyle` when using interval types, the same as `datestyle`. The parameter can become desychronised in transaction pooling modes. So teach pgbouncer to pay attention to the parameter, the same as `datestyle`, to be sure it is consistent.

We were [bitten by this recently](https://github.com/pgbouncer/pgbouncer/issues/470#issuecomment-1480435000), and it looks like a few other folks have been as well. Using the interval type without this is prone to error. We're working around it using a connect query for the moment, but would love to see this tracked the same as `datestyle` for consistency, and to avoid future travellers experiencing the same bug.

Fixes #470

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -202,10 +202,11 @@ Default: 0
 
 ### ignore_startup_parameters
 
-By default, PgBouncer allows only parameters it can keep track of in startup
-packets: `client_encoding`, `datestyle`, `timezone` and `standard_conforming_strings`.
-All others parameters will raise an error.  To allow others parameters, they can be
-specified here, so that PgBouncer knows that they are handled by the admin and it can ignore them.
+By default, PgBouncer allows only parameters it can keep track of in startup packets:
+`client_encoding`, `datestyle`, `intervalstyle`, `timezone` and
+`standard_conforming_strings`.  All others parameters will raise an error.  To allow
+others parameters, they can be specified here, so that PgBouncer knows that they are
+handled by the admin and it can ignore them.
 
 If you need to specify multiple values, use a comma-separated list (e.g.
 `options,extra_float_digits`)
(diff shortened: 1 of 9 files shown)


petere:
One thing I never figured out: If we add settings like this and add them to `SHOW FDS`, how does that work if we do an online restart as part of a pgbouncer version upgrade? Does that work? How does it behave? At least the new setting would probably need to be the last column for this to have any chance of working.

JelteF:
Is online restart across versions really something that we (want to) support? It's definitely not tested. 

PS. Honestly the online restart feature seems rather exotic. I personally never used it, because it doesn't support SSL. So I'm wondering how many people actually use it (and if any do, in what situations they use it). Because it seems like it's only useful in production, and in production I hope most people use SSL.

sj26:
Thanks for taking a look!

If it's helpful as a data point, we use a pool of pgbouncer services and client connection lifetimes coupled with service draining to do online restarts in production.

It would be possible to do online restarts across versions (without ssl) if the values passed across sockets were tagged with keys. But that seems out of scope of this PR.

Is this change something you'd like to incorporate into pgbouncer? Is there anything else I can do here to help land this change?

sj26:
Oh yeah, if there's a way to generalise this then that's heaps better, and that solves our problem.  I was only proposing this way because it was an easy extrapolation.

Our application using rails will set intervalstyle, not as a startup parameter but as part of its preamble with a new connection, so if that is tracked by pgbouncer in the varcache beyond startup as well then that should work. We could probably also include it as a startup parameter in our connection string.

sj26:
Nice, I am glad to see #867 merge to solve this problem in a more flexible way :pray:

PR #833, opened 2023-04-21
Add a dedicated cancel_wait_timeout

JelteF:
Cancellations and regular queries are quite different. So it makes sense
to have different (default) timeouts for cancellation requests than for
queries. While I already think the default query_wait_timeout of 120
seconds we have is excessively long, it doesn't make sense at all to
have such a long timeout by default for cancellation requests that are
not sent. So this introduced a new `cancel_wait_timeout` which is set to
10 seconds by default. If the cancellation request cannot be forwarded
within that time it's probably better to drop it.

One reason why it's important to relatively quickly close cancellation
requests when they are not being handled is because libpq only allows
the client to send a cancel in a blocking way. So there's no possibility
to bail out after a certain amount of seconds on the client side, if
PgBouncer keeps the connection to the client open. Both me and Marco
noticed this when testing my PgBouncer peering PR (#666), when an entry
in the `[peers]` list pointed to an unreachable host. That meant that
pgbouncer would keep the cancel connection open, and even though the
query had finished the `psql` would not display the result because it
was stuck waiting for a result to the cancel request.

To avoid race conditions of cancellations that already forwarded, but
not yet answered cancellation requests are not timed out. See #717 for
details on such race conditions.

### DIFF
--- a/etc/pgbouncer.ini
+++ b/etc/pgbouncer.ini
@@ -266,6 +266,12 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; failure. (default: 120)
 ;query_wait_timeout = 120
 
+;; Dangerous.  Client connection is closed if the cancellation request
+;; is not assigned to a server in this time.  Should be used to limit
+;; the time a client application blocks on a queued cancel request in
+;; case of a database or network failure. (default: 120)
+;cancel_wait_timeout = 10
+
 ;; Dangerous.  Client connection is closed if no activity in this
 ;; time.  Should be used to survive network problems. (default: 0)
 ;client_idle_timeout = 0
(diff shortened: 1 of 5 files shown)


etc/pgbouncer.ini
```diff
@@ -266,6 +266,12 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; failure. (default: 120)
 ;query_wait_timeout = 120
 
+;; Dangerous.  Client connection is closed if the cancellation request
```
petere, line 269: I wonder if the "Dangerous" is really appropriate here, or was it just copied from the `query_` settings?

petere (approved):

etc/pgbouncer.ini
JelteF: Disabling it is definitely dangerous, but yeah it was mostly copied.

PR #837, opened 2023-04-25
Improve warning when connection to unix socket fails

### LINKED ISSUE #835: Debian and Unix Sockets
I can't seem to get PGBouncer to use **/tmp** directory as the **unix_socket_dir**.  It seems to always want **/var/run/postgresql**.  I already have another PGBouncer instance using **/var/run/postgresql** and I want to test a new version of PGBouncer on the same machine using a different unix socket directory, but it won't accept **/tmp**.  It always gives this error:
```
psql: error: could not connect to server: could not connect to server: Connection refused
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.6432"?

```
Using the latest, v 1.80 on Debian Linux distro.

JelteF:
PgBouncer is hard to debug when it's trying to connect to Postgres
server over a UNIX socket but that connection fails. You would get an
unhelpful error message like this:
```
WARNING sbuf_connect failed: No such file or directory
```

This change starts including the path in the error message:
```
WARNING sbuf_connect failed to connect to UNIX socket /aaaaa/.s.PGSQL.5432: No such file or directory
```

It also includes the address+port for IPv4/IPv6 connection attempts.

Fixes #835

### DIFF
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -150,6 +150,7 @@ bool sbuf_connect(SBuf *sbuf, const struct sockaddr *sa, socklen_t sa_len, time_
 {
 	int res, sock;
 	struct timeval timeout;
+	char buf[128];
 	bool is_unix = sa->sa_family == AF_UNIX;
 
 	Assert(iobuf_empty(sbuf->io) && sbuf->sock == 0);
@@ -189,7 +190,8 @@ bool sbuf_connect(SBuf *sbuf, const struct sockaddr *sa, socklen_t sa_len, time_
 	}
 
 failed:
-	log_warning("sbuf_connect failed: %s", strerror(errno));
+	log_warning("sbuf_connect failed to connect to %s: %s",
+			sa2str(sa, buf, sizeof(buf)), strerror(errno));
 
 	if (sock >= 0)
 		safe_close(sock);


eulerto:
Why don't you provide messages for IPv4 and IPv6? AFAICS, this `log_warning` call does not provide a PgSocket as first parameter so address information (`postgres/euler@unix:9915`) provided by `log_socket_prefix` is not added after the prefix (`2023-02-27 16:27:24.299 -03 [33171] DEBUG`). It is not possible to provide PgSocket to sbuf API (it is in a different layer), hence, you should construct a string that contains host + port or Unix socket path for sbuf. Perhaps, add this address information before the message (to be similar to what `logging_prefix_cb` does).

JelteF:
I wasn't able to trigger the WARNING for IPv4/IPv6 sockets. They would error during reading instead of erroring at connect. Those errors during reading would become a LOG with the log_socket_prefix. So those are already pretty helpful.

eulerto:
I didn't try to reproduce it but the [following code](https://github.com/pgbouncer/pgbouncer/blob/master/src/sbuf.c#L161) can trigger it. Looking at `man socket`, I expect some errors (such as EACCESS, EMFILE, ENFILE, ENOMEM) in certain situations.

```
    sock = socket(sa->sa_family, SOCK_STREAM, 0);
    if (sock < 0) { 
        /* probably fd limit */
        goto failed;
    }
```

JelteF:
Thanks for the pointer. By manually setting my file limit very low I was able to trigger the WARNING for IPv4/IPv6 sockets too. So now the patch includes a better error for those as well.

JelteF:
Makes total sense. Pushed that now. I had a bit of a brainfart I think because I misremembered the fix for this issue: https://github.com/libusual/libusual/commit/946941916ba8ea5db15498b20579ec4336be02e8

eulerto (approved):
LGTM

PR #449, opened 2019-12-20
Add application_name to SHOW CLIENTS/SERVERS/SOCKETS output

stevenwinfield:
Implements the feature request at https://github.com/pgbouncer/pgbouncer/issues/167

PR #839, opened 2023-04-26
WIP: v1.19.0

petere:


eulerto:
I think we should mention "Remove AppVeyor" (02824649975fb7c892dc544d87743a05e5cf7cc6) in the **Cleanups** section. I also think the PR links at the end of the release notes are not required because GitHub already adds the links for you in each item. If people aren't checking the [NEWS.md](https://github.com/pgbouncer/pgbouncer/blob/master/NEWS.md) in the GitHub interface, we should add the link directly in each item instead of adding them at the end. It is easy to connect the number with the item and numbers will be referred only once.

JelteF:
> I also think the PR links at the end of the release notes are not required because GitHub already adds the links for you in each item. If people aren't checking the [NEWS.md](https://github.com/pgbouncer/pgbouncer/blob/master/NEWS.md?rgh-link-date=2023-05-01T17%3A48%3A18Z) in the GitHub interface, we should add the link directly in each item instead of adding them at the end. It is easy to connect the number with the item and numbers will be referred only once.

In actual markdown files issue numbers don't automatically turn into links. I double checked this by removing the link from the #449 entry in the and looking at it in preview. You can see in this picture it is plain black instead link blue:
image

The list of issue numbers below in the source is actually not visible in the rendered output, but instead turns the issue numbers in the items into links. This is a type of reference style links that is supported by all markdown renderers that I ever used. This gist shows all the different styles of creating links: https://gist.github.com/emedinaa/28ed71b450243aba48accd634679f805

It is much less commonly known than this type:
```
[I'm an inline-style link](https://www.somewebsite.com)
```
But it's much more readable in source form in my opinion, because the long link doesn't interfere with all the text. Especially the nested parenthesis in this case look quite ugly.

eulerto:
I didn't know about the reference links. 👀 Sorry about the noise. Since the references are not displayed, it is fine.

JelteF:
> I didn't know about the reference links. 👀 Sorry about the noise.

No problem, I didn't know about them for a long time either. But now that I do I pretty much always use them.

PR #841, opened 2023-04-26
Retry adding an editorconfig file

JelteF:
This allows many editors to be configured with correct indentation
styles for the different filetypes that we have. Github also parses this
file and uses it for its web-based editor as well as for displaying
files.

It's a slightly modified version of reverted commit a718ad8. This new
version doesn't specify tab_width and indent_size for C files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+# top-most EditorConfig file
+root = true
+
+# basic rules for all files
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+trim_trailing_whitespace = true
+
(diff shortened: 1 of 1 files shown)


petere:
Was this changed from last time? I think the last one was wrong.

JelteF:
Yes:

> It's a slightly modified version of reverted commit https://github.com/pgbouncer/pgbouncer/commit/a718ad830a72e1ebc71b29a65307bceda7e37792. This new
version doesn't specify tab_width and indent_size for C files.

Please let me know if it still has issues for you. But I expect it should be fine now.

PR #846, opened 2023-05-22
Fix: FATAL in function disconnect_client(): bad client state: 0

### LINKED ISSUE #801: FATAL @src/objects.c:1052 in function disconnect_client(): bad client state: 0
Hi! 

we've experiencing an issue with pgbouncer.

`FATAL @src/objects.c:1052 in function disconnect_client(): bad client state: 0` 

after this message in the log it stopped working. pgbouncer version is 1.18.0

JelteF:
During normal operation, cancel clients get closed because their linked
server finished sending the cancel request. But this is not always the
case. It's possible for the client to disconnect itself, which would
cause the client object to be freed, before the linked server is. In
those cases the client was not unlinked from the server, thus when the
server was then closed it would try to close the already freed client
again. This would then result in a fatal error.

In passing this adds a test for cancel_wait_timeout. With the help of
that test and some hacky code changes (including ones in postgres) I was
able to reliably reproduce this issue.

Fixes #801


eulerto:
Did you test with `enable-cassert`? (CI covers it but since you mentioned you have to modify Postgres to reliably reproduce it, I'm curious if it passes your tests.)

Shouldn't you `disconnect_server()` with `send_term` = true when state is `CL_ACTIVE_CANCEL`?

Add a comment explaining why you're cleaning the link (extract the words from the commit message). It is faster to read the code than to find the exact change reference in the commit logs.

JelteF:
> Did you test with `enable-cassert`? (CI covers it but since you mentioned you have to modify Postgres to reliably reproduce it, I'm curious if it passes your tests.)

I don't think I had cassert enabled. I enabled it now and reran my manual tests. All is good.

> Shouldn't you `disconnect_server()` with `send_term` = true when state is `CL_ACTIVE_CANCEL`?

No, send_term is only part of the query protocol. The cancelation protocol would not understand it. It only expects 32 bits for the PID and 32bits for the secret, and then a connection close.

> Add a comment explaining why you're cleaning the link (extract the words from the commit message). It is faster to read the code than to find the exact change reference in the commit logs.

Done

eulerto:
> Good call. I don't think I had cassert enabled. I enabled it now and reran my manual tests. All is good.

Ok.

> No, send_term is only part of the query protocol. The cancelation protocol would not understand it. It only expects 32 bits for the PID and 32bits for the secret, and then a connection close.

Hmm. Ok.

eulerto (approved):
LGTM

PR #852, opened 2023-05-30
Add files required to run python based tests to release tarball

### LINKED ISSUE #851: pytest uses strict mode for asyncio without pyproject.toml in the release tarball
If the pyproject.toml shouldn't be in the release tarball, then I think for distro package maintainers' sake the pytest command should append `--asyncio-mode=auto` flag in the Makefile.

JelteF:
This is useful for distro package maintainers, as well as anyone else
that wants to run tests on the published sources.

Fixes #851

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,7 @@ dist_man_MANS = doc/pgbouncer.1 doc/pgbouncer.5
 
 # files in tgz
 EXTRA_DIST = AUTHORS COPYRIGHT Makefile config.mak.in config.sub config.guess \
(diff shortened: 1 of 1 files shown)


petere (approved):

narmatha-kk:
Hii is it possible to re-route read-only queries to replica using pytest?

JelteF:
I'm not sure how you would see that work. pytest is only for pgbouncer its testing framework, and currently there are no replicas in use at all in any of our tests.

Issue #482, opened 2020-04-29
not support the startup parameter search_path and extra_float_digits


aogb:
when i use JDBC connect to PGBOUNCER with currentSchema=...... in connection string, an exception with not support the startup parameter search_path and extra_float_digits occurs.

Please assist!!!

eulerto:
Add it to [ignore_startup_parameters](https://www.pgbouncer.org/config.html#ignore_startup_parameters) in your `pgbouncer.ini`.

ziegenberg:
Would it be feasible to add `extra_float_digits` to the list of parameters PgBouncer can keep track of in startup packets?

petere:
What makes `extra_float_digits` so special that we need to go out of our way to track it?

Issue #452, opened 2020-01-07
Add a warning about tools such as pg_dump w.r.t. transaction pooling.


pope1ni:
It would be useful to have a clearly visible warning in the documentation that certain standard tools, e.g. `pg_dump`, will cause session breakage if used with PgBouncer and transaction pooling mode.

I was experiencing the classic issue where the `search_path` was changed and was receiving errors about missing relations, but it didn't occur to me that `pg_dump` would be at fault. (We've been using `barman` for backups for a long time and more recently `pg_dump` was used again for some very limited data-only dumps.)

I eventually came across the following information which made me aware of the problem:
- https://github.com/pgbouncer/pgbouncer/issues/246#issuecomment-383582053
- https://gitlab.com/gitlab-org/gitlab/issues/23211

It's all to easy to overlook the port setting in something like this and the error doesn't quickly point the finger at the right culprit.

JelteF:
If it's a big problem for you, you can install Citus and configure `track_extra_parameters` to include `search_path`. Creating a small Postgres extension that does the same trick also shouldn't be hard. Other than that we need some change in Postgres to make this configurable. One of such options is this thread: https://www.postgresql.org/message-id/flat/0a6f5e6b-65b3-4272-260d-a17ce8f5b3a4%40eisentraut.org#bd6d06db7db9e4ded8100e4160050b17

PR #867, opened 2023-06-12
Make varcache (the pg parameters cached per connection) dynamically configurable.

### LINKED ISSUE #470: Allow IntervalStyle parameter (like DateStyle)

emelsimsek:
This change enables caching other postgres parameters in addition to the static list `{"client_encoding", "DateStyle", "TimeZone", "standard_conforming_strings", "application_name"}` that is already supported.

  **Testing:**
  
  Add  ` track_extra_parameters `  in *.ini file to enable caching search_path per connection.
  
  ```
  [pgbouncer]
    ...
    track_extra_parameters = search_path, IntervalStyle
  ```
 
Related to #452 and #482
Fixes #470
Fixes #821


include/varcache.h
```diff
@@ -1,20 +1,24 @@
 #include <usual/strpool.h>
 
+#define MAX_NUM_CACHE_VARS 100
```
JelteF, line 3: Let's remove this and make the size dependent on the variables that are configured in the ini file.

include/varcache.h
```diff
@@ -1,20 +1,24 @@
...
+
 enum VarCacheIdx {
 	VDateStyle = 0,
 	VClientEncoding,
 	VTimeZone,
 	VStdStr,
 	VAppName,
+	VSearchPath,
```
JelteF, line 11: I think this isn't needed anymore.

JelteF: Actually this whole enum can now be removed I think.

src/main.c
```diff
@@ -238,6 +239,7 @@ CF_ABS("auth_query", CF_STR, cf_auth_query, 0, "SELECT usename, passwd FROM pg_s
 CF_ABS("auth_type", CF_LOOKUP(auth_type_map), cf_auth_type, 0, "md5"),
 CF_ABS("auth_user", CF_STR, cf_auth_user, 0, NULL),
 CF_ABS("autodb_idle_timeout", CF_TIME_USEC, cf_autodb_idle_timeout, 0, "3600"),
+CF_ABS("cache_vars", CF_STR, cf_cache_vars, 0, NULL),
```
JelteF, line 242: We should have a test for this new config variable.

src/client.c
```diff
@@ -584,17 +585,20 @@ static bool decide_startup_pool(PgSocket *client, PktHdr *pkt)
 		if (!ok)
 			break;
 
+		if (!pg_quote_literal(qbuf, val, sizeof(qbuf)))
+			return 0;
```
JelteF, line 589: let's only do this when we actually use qbuf (i.e. application_name and varcache ones).

JelteF: Actually only the varcache ones. The appname one we want unquoted.

JelteF: Reason why we should only do this for varcache_set ones, is because we want to support values larger than 128 bytes for others.

src/varcache.c
```diff
@@ -45,9 +40,60 @@ static inline struct PStr *get_value(VarCache *cache, const struct var_lookup *l
...
+{
+	char *s, *p, *q;
+	struct var_lookup *lookup = NULL;
+
+	if ((s = strdup(cf_cache_vars)) == NULL)
+		return;
+
+	q = s;
+	while ((p = strsep(&q, ",")) != NULL) {
+		while (*p == ' ' || *p == '\t')
```
JelteF, line 53: Let's use the `parse_word_list` helper instead if possible.

src/main.c
JelteF, line 242: All tests in CI are failing because the option is not in the docs.
```
etc/optscan.sh
cache_vars is missing in doc/config.md
cache_vars is missing in etc/pgbouncer.ini
```

Lets add it. Also lets name the option `extra_tracked_startup_parameters`.

JelteF: For the test we can check that we indeed correctly track any of the settings mentioned in this section of the PG docs 

> At present there is a hard-wired set of parameters for which ParameterStatus will be generated: they are server_version, server_encoding, client_encoding, application_name, default_transaction_read_only, in_hot_standby, is_superuser, session_authorization, DateStyle, IntervalStyle, TimeZone, integer_datetimes, and standard_conforming_strings. (server_encoding, TimeZone, and integer_datetimes were not reported by releases before 8.0; standard_conforming_strings was not reported by releases before 8.1; IntervalStyle was not reported by releases before 8.4; application_name was not reported by releases before 9.0; default_transaction_read_only and in_hot_standby were not reported by releases before 14.) Note that server_version, server_encoding and integer_datetimes are pseudo-parameters that cannot change after startup. This set might change in the future, or even become configurable. Accordingly, a frontend should simply ignore ParameterStatus for parameters that it does not understand or care about.

Except for one of the ones that we always track, but e.g IntervalStyle would be a good one, because people have had issues where that was not tracked.

src/client.c
```diff
@@ -569,13 +569,29 @@ static void set_appname(PgSocket *client, const char *app_name)
 	}
 }
 
+static bool varcache_set_quoted(PgSocket *client, const char *key, const char *value)
```
JelteF, line 572: nit: lets move this function to `varcache.c`

eulerto:
This is the 2nd patch that is proposing to include [uthash](https://github.com/troydhanson/uthash). I'm wondering if, for maintenance reason, it would be interesting to include this code into the tree. It is another task to do before releasing a new PgBouncer version. It is also hard to track bugs in this code. Let's not be responsible to this code unless it is no longer maintained. My suggestion is to create another submodule (like [libusual](https://github.com/libusual/libusual)).

Should `uthash` be added to `libusual`? Couldn't the hash code from `libusual` be used?

Regarding the `track_startup_parameters`, it is described as:

> By default, PgBouncer tracks `client_encoding`, `datestyle`, `timezone`, `standard_conforming_strings`
> and `application_name` parameters per client. To allow other parameters to be tracked, they can be
> specified here, so that PgBouncer knows that they should be maintained in the client variable cache
> and restored in the server whenever the client becomes active.

It means there is an list that is part of tracked parameters but these parameters aren't part of the default value. These parameters are part of an implementation detail but your PR is exposing it but keeping the original part hidden. If I was proposing such feature I wouldn't use 2 lists. Instead, use only a dynamic list and the default value is the current list. I would also add a check to avoid someone from removing the original list (I don't have a strong opinion if we should allow removing some of these original parameters from the list.)

test/test_misc.py
```diff
@@ -32,6 +32,30 @@ def test_fast_close(bouncer):
...
+    cur1.execute("SHOW intervalstyle")
+    cur2.execute("SHOW intervalstyle")
+
+    result1 = cur1.fetchone();
+    assert result1[0] == "sql_standard"
+
+    result2 = cur2.fetchone();
+    assert result2[0] == "postgres"
+
+   finally:
```
JelteF, line 56: This is invalid python code. Let's use the `with` keyword and the `cur` method to simplify it and make it working:
```
def test_track_startup_parameters(bouncer):
    bouncer.admin(f"set pool_mode=transaction")
    bouncer.admin(f"set track_startup_parameters = search_path, IntervalStyle")

    with bouncer.cur(dbname="p1") as cur1:
        with bouncer.cur(dbname="p1") as cur2:
            cur1.execute("SET intervalstyle = sql_standard");
            cur2.execute("SET intervalstyle = postgres");

            cur1.execute("SHOW intervalstyle")
            cur2.execute("SHOW intervalstyle")

            result1 = cur1.fetchone();
            assert result1[0] == "sql_standard"

            result2 = cur2.fetchone();
            assert result2[0] == "postgres"

```

src/main.c
```diff
@@ -238,6 +239,7 @@ CF_ABS("auth_query", CF_STR, cf_auth_query, 0, "SELECT usename, passwd FROM pg_s
 CF_ABS("auth_type", CF_LOOKUP(auth_type_map), cf_auth_type, 0, "md5"),
 CF_ABS("auth_user", CF_STR, cf_auth_user, 0, NULL),
 CF_ABS("autodb_idle_timeout", CF_TIME_USEC, cf_autodb_idle_timeout, 0, "3600"),
+CF_ABS("track_startup_parameters", CF_STR, cf_track_startup_parameters, 0, NULL),
```
JelteF, line 242: nit: now that we changed the name, this is not the right place in this alphabetically ordered list anymore.

JelteF, line 242: If this setting is not part of the config pgbouncer now segfaults (which is why all tests in CI are failing). Changing default to `""` seems to help.
```suggestion
CF_ABS("track_startup_parameters", CF_STR, cf_track_startup_parameters, 0, ""),
```

src/varcache.c
```diff
@@ -79,12 +132,37 @@ bool varcache_set(VarCache *cache, const char *key, const char *value)
...
+	char qbuf[400];
+	const char *tmp;
+
+	tmp = value;
+	/* Only quote search_path. Quoting other parameters such as client_encoding
+	 * does not work well some clients like psycopg.*/
+	if (strcmp("search_path", key) == 0) {
+
+		if (!pg_quote_literal(qbuf, value, sizeof(qbuf))) {
+			slog_warning(client, "could not quote parameter: %s=%s", key, value);
```
marcocitus, line 146: is quoting search_path correct?

These are 2 different things:
```sql
SET search_path TO public, private;
SET search_path TO "public, private";
```

The second one refers to a schema created using `CREATE SCHEMA "public, private"`

JelteF: Emel is adding more tests for the quoting now. I also think the new code is incorrect in some ways. One thing to note as that `search_path` is special compared to the previous items in varcache because it's a GUC_LIST. So these two lines are equivalent:
```sql
SET search_path TO public, private;
SET search_path TO "public", "private";
```

emelsimsek:
> Regarding the `track_startup_parameters`, it is described as:
> 
> > By default, PgBouncer tracks `client_encoding`, `datestyle`, `timezone`, `standard_conforming_strings`
> > and `application_name` parameters per client. To allow other parameters to be tracked, they can be
> > specified here, so that PgBouncer knows that they should be maintained in the client variable cache
> > and restored in the server whenever the client becomes active.
> 
> It means there is an list that is part of tracked parameters but these parameters aren't part of the default value. These parameters are part of an implementation detail but your PR is exposing it but keeping the original part hidden. If I was proposing such feature I wouldn't use 2 lists. Instead, use only a dynamic list and the default value is the current list. I would also add a check to avoid someone from removing the original list (I don't have a strong opinion if we should allow removing some of these original parameters from the list.)

This is actually in line with `ignore_startup_parameters` in the way it has an immutable internal default list which can be extended via configuration. The configuration can only add to the default list.

What do you think about the name `track_startup_parameters`? Not every postgres parameter can be part of connection string: 
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING:~:text=for%20different%20hosts.-,34.1.2.%C2%A0Parameter%20Key%20Words,-The%20currently%20recognized

In fact IntervalStyle is not a startup parameter. 

Should we rename  `track_startup_parameters` to something else, e.g.  `track_postgres_parameters`?

src/varcache.c
```diff
@@ -99,12 +180,17 @@ static int apply_var(PktBuf *pkt, const char *key,
 	if (strcasecmp(cval->str, sval->str) == 0)
 		return 0;
 
-	/* the string may have been taken from startup pkt */
-	if (!pg_quote_literal(qbuf, cval->str, sizeof(qbuf)))
+	if (strcasecmp("search_path", key) == 0) {
```
JelteF, line 183: Needs a comment why search_path is special.

doc/config.md
```diff
@@ -200,6 +200,18 @@ achieving uniform load.
...
 
+### track_startup_parameters
+
+By default, PgBouncer tracks `client_encoding`, `datestyle`, `timezone`, `standard_conforming_strings`
+and `application_name` parameters per client. To allow other parameters to be tracked, they can be
+specified here, so that PgBouncer knows that they should be maintained in the client variable cache
+and restored in the server whenever the client becomes active.
+
+If you need to specify multiple values, use a comma-separated list (e.g.
+`search_path, IntervalStyle`)
```
JelteF, line 211:
```suggestion
If you need to specify multiple values, use a comma-separated list (e.g.
`default_transaction_readonly, IntervalStyle`). 

Note: Most parameters cannot be tracked this way. The only parameters that can be tracked are ones that Postgres reports to the client. Postgres has [an official list of parameters that it reports to the client](https://www.postgresql.org/docs/15/protocol-flow.html#PROTOCOL-ASYNC). Postgres extensions can change this list though, they can add parameters themselves that they also report, and they can start reporting already existing paremeters that Postgres does not report. Notably Citus 12.0+ causes Postgres to also report `search_path`.
```

etc/pgbouncer.ini
```diff
@@ -161,6 +161,11 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; is off, server_reset_query is used only for session-pooling.
 ;server_reset_query_always = 0
 
+;; Comma-separated list of parameters to track per client. The Postgres 
+;; parameters listed here will be cached per client by pgbouncer and 
+;; restored in server everytime the client runs a query. 
+;track_startup_parameters = search_path
```
JelteF, line 167: Let's use things that work on regular postgres in the docs.
```suggestion
;track_startup_parameters = IntervalStyle
```

src/varcache.c
JelteF: Nvm, this function is dead code now. So it can definitely be removed.

src/varcache.c
JelteF: Actually it turns out that this needs even more special quoting...

The correct way of quoting a GUC which has the GUC_LIST_QUOTE flag is by doing what pgdump does:
https://github.com/postgres/postgres/blob/555b929bbe7bd2f228e804d2c5bf4a246a319358/src/bin/pg_dump/dumputils.c#L847-L880

Otherwise things break with long search paths or an empty list:
```
> set search_path = aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
NOTICE:  42622: identifier "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" will be truncated to "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
and 
```
> set search_path = "";
ERROR:  42601: zero-length delimited identifier at or near """"
LINE 1: set search_path = "";
```

Probably best to copy this quoting logic from pgdump to a separate file.

src/varcache.c
```diff
@@ -79,13 +133,40 @@ bool varcache_set(VarCache *cache, const char *key, const char *value)
...
+}
+
 static int apply_var(PktBuf *pkt, const char *key,
 		     const struct PStr *cval,
 		     const struct PStr *sval)
 {
 	char buf[128];
 	char qbuf[128];
 	unsigned len;
+	const char *tmp;
```
JelteF, line 169: Let's increase those buffers a little bit.

src/varcache.c
```diff
@@ -99,12 +180,17 @@ static int apply_var(PktBuf *pkt, const char *key,
...
+	}
+	else if (pg_quote_literal(qbuf, cval->str, sizeof(qbuf))){
+		tmp = qbuf;
+	}
+	else
 		return 0;
 
 	/* add SET statement to packet */
-	len = snprintf(buf, sizeof(buf), "SET %s=%s;", key, qbuf);
+	len = snprintf(buf, sizeof(buf), "SET %s=%s;", key, tmp);
```
JelteF, line 193: If it doesn't fit, let's use a allocated buffer instead of logging a warning that the value is too long (line 199)

src/varcache.c
JelteF: As discussed, let's only consider the empty string case to keep the code simple. Long identifiers being truncated is not important, schema identifiers are already bounded to the truncated length anyway.

src/varcache.c
```diff
@@ -127,11 +213,13 @@ bool varcache_apply(PgSocket *server, PgSocket *client, bool *changes_p)
 	/* grab query position inside pkt */
 	sql_ofs = pktbuf_written(pkt);
 
-	for (lk = lookup; lk->name; lk++) {
+	HASH_ITER(hh, lookup_map, lk, tmp) {
+
```
JelteF, line 217: remove empty line
```suggestion
```

src/varcache.c
```diff
@@ -99,26 +167,48 @@ static int apply_var(PktBuf *pkt, const char *key,
 	if (strcasecmp(cval->str, sval->str) == 0)
 		return 0;
 
-	/* the string may have been taken from startup pkt */
-	if (!pg_quote_literal(qbuf, cval->str, sizeof(qbuf)))
+	/* parameters that are marked GUC_LIST_QUOTE are returned already fully quoted
+	 * re-quoting them using pg_quote_literal will result in malformed values. */
+	if (variable_is_guc_list_quote(key)) {
+		/* zero length elements of the form "" should be specially handled.*/
+		if (strlen(cval->str) == 2 && strcmp(cval->str,"\"\"") == 0) {
```
JelteF, line 174: No need to check strlen beforehand
```suggestion
		if (strcmp(cval->str,"\"\"") == 0) {
```

include/uthash.h
```diff
@@ -0,0 +1,1140 @@
+/*
```
JelteF, line 1: Can you make this a git submodule instead of including the file itself?

JelteF (approved):
IMHO this is now good to merge. Apart from the two new comments that I left (strlen check and uthash git submodule).

@eulerto do you still want to give this a final look or are you good with it too.

src/main.c
```diff
@@ -314,6 +315,7 @@ CF_ABS("tcp_keepidle", CF_INT, cf_tcp_keepidle, 0, "0"),
 CF_ABS("tcp_keepintvl", CF_INT, cf_tcp_keepintvl, 0, "0"),
 CF_ABS("tcp_socket_buffer", CF_INT, cf_tcp_socket_buffer, 0, "0"),
 CF_ABS("tcp_user_timeout", CF_INT, cf_tcp_user_timeout, 0, "0"),
+CF_ABS("track_extra_parameters", CF_STR, cf_track_extra_parameters, CF_NO_RELOAD, ""),
```
JelteF, line 318: Let's set the default to IntervalStyle, since people reported problems with it not being tracked. 
```suggestion
CF_ABS("track_extra_parameters", CF_STR, cf_track_extra_parameters, CF_NO_RELOAD, "IntervalStyle"),
```

Makefile
```diff
@@ -59,7 +59,9 @@ pgbouncer_SOURCES = \
 	include/common/unicode_norm.h \
 	include/common/unicode_norm_table.h
 
+UTHASH_HEADER = uthash/src
 pgbouncer_CPPFLAGS = -Iinclude $(CARES_CFLAGS) $(LIBEVENT_CFLAGS) $(TLS_CPPFLAGS)
+pgbouncer_CPPFLAGS += -I$(UTHASH_HEADER)
```
JelteF, line 64: You also need to add the uthash license and header file to EXTRA_DIST. Otherwise they won't be part of the resulting tar.gz file that's created by `make dist`. That's also the reason why some CI jobs are failing to build. Because they build from the tarball instead of the git sources.

JelteF (approved):

ReganRyanNZ:
Does this mean I can run `pg_dump` with pgbouncer now? In the past there were issues with search_path being set and things breaking. So now that it's cached per connection...?

JelteF:
If you have the Citus extension installed and have set `track_extra_parameters = search_path, IntervalStyle`, then yes. Otherwise you'll need to create your own extension that marks search_path as GUC_REPORT or wait for some of my relevant postgres patches to be merged and released.

PR #876, opened 2023-06-30
Support replication connections through PgBouncer

### LINKED ISSUE #382: Getting error while running the pg_basebackup through PGBOUNCER
We are using PGBOUNCER(connection pool mechanisam). PGBOUNCER uses port 5433.

Postgres database port number is 6433. By using port 5433 PGBOUNCER is connecting to postgres port 6433 database.

Now PGBOUNCER is establishing the connections properly but when I try to run the pg_basebackup through port 5433(PGBOUNCER port) we are receiving below error. Please guide me.
 
 
 /opt/postgres/9.2/bin/pg_basebackup -p 5433 -U postgres -P -v -x --format=tar --gzip --compress=1 --pgdata=- -D /opt/rao

pg_basebackup: could not connect to server: ERROR:  Unsupported startup parameter: replication

JelteF:
In session pooling mode PgBouncer is pretty much a transparent proxy,
i.e. the client does normally not even need to know that PgBouncer is in
the middle. This allows things like load balancing and failovers without
the client needing to know about this at all. But as soon as replication
connections are needed, this was not possible anymore, because PgBouncer
would reject those instead of proxying them to the right server.

This PR fixes that by also proxying replication connections. They are
handled pretty differently from normal connections though. A client and
server replication connection will form a strong pair, as soon as one is
closed the other is closed too. So, there's no caching of the server
replication connections, like is done for regular connections. Reusing
replication connections comes with a ton of gotchas. Postgres will
throw errors in many cases when trying to do so. So simply not
doing it seems like a good tradeoff for ease of implementation.
Especially because replication connections are pretty much always
very long lived. So re-using them gains pretty much no performance
benefits.

Fixes #382

Depends on #945

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -1436,7 +1436,7 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
 (see <https://www.postgresql.org/docs/current/auth-pg-hba-conf.html>).
 
 * Supported record types: `local`, `host`, `hostssl`, `hostnossl`.
-* Database field: Supports `all`, `sameuser`, `@file`, multiple names.  Not supported: `replication`, `samerole`, `samegroup`.
+* Database field: Supports `all`, `replication`, `sameuser`, `@file`, multiple names.  Not supported: `samerole`, `samegroup`.
 * User name field: Supports `all`, `@file`, multiple names.  Not supported: `+groupname`.
 * Address field: Supports IPv4, IPv6.  Not supported: DNS names, domain prefixes.
 * Auth-method field: Only methods supported by PgBouncer's `auth_type`
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -140,6 +140,7 @@ typedef struct PktBuf PktBuf;
 typedef struct ScramState ScramState;
 typedef struct PgPreparedStatement PgPreparedStatement;
 typedef enum ResponseAction ResponseAction;
+typedef enum ReplicationType ReplicationType;
 
 extern int cf_sbuf_len;
 
@@ -610,6 +611,14 @@ typedef struct OutstandingRequest {
 	uint64_t server_ps_query_id;
 } OutstandingRequest;
 
+enum ReplicationType {
+	REPLICATION_NONE = 0,
+	REPLICATION_LOGICAL,
+	REPLICATION_PHYSICAL,
+};
+
+extern const char *replication_type_parameters[3];
+
 /*
  * A client or server connection.
  *
@@ -653,6 +662,9 @@ struct PgSocket {
 	/* server: received an ErrorResponse, waiting for ReadyForQuery to clear
 	 * the outstanding requests until the next Sync */
 	bool query_failed : 1;
+
+	ReplicationType replication;	/* If this is a replication connection */
+	char *startup_options;	/* only tracked for replication connections */
 
 	usec_t connect_time;	/* when connection was made */
 	usec_t request_time;	/* last activity time */
--- a/include/common/postgres_compat.h
+++ b/include/common/postgres_compat.h
@@ -7,6 +7,7 @@
 /* from c.h */
 
 #include <string.h>
+#include <usual/ctype.h>
 
 #define int8 int8_t
 #define uint8 uint8_t
--- a/src/admin.c
+++ b/src/admin.c
@@ -623,13 +623,13 @@ static bool admin_show_users(PgSocket *admin, const char *arg)
 	return true;
 }
 
-#define SKF_STD "sssssisiTTiiississi"
-#define SKF_DBG "sssssisiTTiiississiiiiiiii"
+#define SKF_STD "ssssssisiTTiiississi"
+#define SKF_DBG "ssssssisiTTiiississiiiiiiii"
 
 static void socket_header(PktBuf *buf, bool debug)
 {
 	pktbuf_write_RowDescription(buf, debug ? SKF_DBG : SKF_STD,
-				    "type", "user", "database", "state",
+				    "type", "user", "database", "replication", "state",
 				    "addr", "port", "local_addr", "local_port",
 				    "connect_time", "request_time",
 				    "wait", "wait_us", "close_needed",
@@ -660,6 +660,7 @@ static void socket_row(PktBuf *buf, PgSocket *sk, const char *state, bool debug)
 	const struct PStr *application_name = v->var_list[VAppName];
 	usec_t now = get_cached_time();
 	usec_t wait_time = sk->query_start ? now - sk->query_start : 0;
+	char *replication;
 
 	if (io) {
 		pkt_avail = iobuf_amount_parse(sk->sbuf.io);
@@ -692,10 +693,18 @@ static void socket_row(PktBuf *buf, PgSocket *sk, const char *state, bool debug)
 	else
 		prepared_statement_count = HASH_COUNT(sk->client_prepared_statements);
 
+	if (sk->replication == REPLICATION_NONE)
+		replication = "none";
+	else if (sk->replication == REPLICATION_LOGICAL)
+		replication = "logical";
+	else
+		replication = "physical";
+
 	pktbuf_write_DataRow(buf, debug ? SKF_DBG : SKF_STD,
 			     is_server_socket(sk) ? "S" : "C",
 			     sk->login_user_credentials ? sk->login_user_credentials->name : "(nouser)",
 			     sk->pool && !sk->pool->db->peer_id ? sk->pool->db->name : "(nodb)",
+			     replication,
 			     state, r_addr, pga_port(&sk->remote_addr),
 			     l_addr, pga_port(&sk->local_addr),
 			     sk->connect_time,
@@ -898,7 +907,7 @@ static bool admin_show_pools(PgSocket *admin, const char *arg)
 		pool = container_of(item, PgPool, head);
 		waiter = first_socket(&pool->waiting_client_list);
 		max_wait = (waiter && waiter->query_start) ? now - waiter->query_start : 0;
-		pool_mode = pool_pool_mode(pool);
+		pool_mode = probably_wrong_pool_pool_mode(pool);
 		pktbuf_write_DataRow(buf, "ssiiiiiiiiiiiiis",
 				     pool->db->name, pool->user_credentials->name,
 				     statlist_count(&pool->active_client_list),
--- a/src/hba.c
+++ b/src/hba.c
@@ -381,8 +381,8 @@ static bool parse_names(struct HBAName *hname, struct TokParser *tp, bool is_db,
 				return false;
 			}
 			if (eat_kw(tp, "replication")) {
-				log_warning("replication is not supported");
-				return false;
+				hname->flags |= NAME_REPLICATION;
+				goto eat_comma;
 			}
 		}
 
@@ -924,7 +924,7 @@ static bool match_inet6(const struct HBARule *rule, PgAddr *addr)
 	       (src[2] & mask[2]) == base[2] && (src[3] & mask[3]) == base[3];
 }
 
-struct HBARule * hba_eval(struct HBA *hba, PgAddr *addr, bool is_tls, const char *dbname, const char *username)
+struct HBARule * hba_eval(struct HBA *hba, PgAddr *addr, bool is_tls, ReplicationType replication, const char *dbname, const char *username)
 {
 	struct List *el;
 	struct HBARule *rule;
@@ -958,8 +958,14 @@ struct HBARule * hba_eval(struct HBA *hba, PgAddr *addr, bool is_tls, const char
 		}
 
 		/* match db & user */
-		if (!name_match(&rule->db_name, dbname, dbnamelen, username))
-			continue;
+		if (replication == REPLICATION_PHYSICAL) {
+			if (!(rule->db_name.flags & NAME_REPLICATION)) {
+				continue;
+			}
+		} else {
+			if (!name_match(&rule->db_name, dbname, dbnamelen, username))
+				continue;
+		}
 		if (!name_match(&rule->user_name, username, unamelen, dbname))
 			continue;
 
--- a/src/janitor.c
+++ b/src/janitor.c
@@ -187,7 +187,14 @@ static void per_loop_activate(PgPool *pool)
 	sv_used = statlist_count(&pool->used_server_list);
 	statlist_for_each_safe(item, &pool->waiting_client_list, tmp) {
 		client = container_of(item, PgSocket, head);
-		if (!statlist_empty(&pool->idle_server_list)) {
+		if (client->replication) {
+			/*
+			 * For replication connections we always launch
+			 * a new connection, but we continue with the loop,
+			 * because there might be normal clients waiting too.
+			 */
+			launch_new_connection(pool, /* evict_if_needed= */ true);
+		} else if (!statlist_empty(&pool->idle_server_list)) {
 			/* db not fully initialized after reboot */
 			if (client->wait_for_welcome && !pool->welcome_msg_ready) {
 				launch_new_connection(pool, /* evict_if_needed= */ true);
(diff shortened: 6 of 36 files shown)


Melkij:
>  So re-using them gains pretty much no performance benefits.

Moreover, replication connections cannot be reliably reused at all. Attempting to do so will lead to various mysteries like
```
elog(ERROR, "no record found"); /* shouldn't happen */
```
in wal sender code. Wal sender implicitly assumes that it works with one wal receiver.
(Based on my experience with a single cloud service where admins tried to use connection pooling for replication too)

JelteF:
> Moreover, replication connections cannot be reliably reused at all.

Thanks. I updated the wording in the PR to include that.

src/admin.c
```diff
@@ -882,12 +890,15 @@ static bool admin_show_pools(PgSocket *admin, const char *arg)
 	statlist_for_each(item, &pool_list) {
 		pool = container_of(item, PgPool, head);
 		waiter = first_socket(&pool->waiting_client_list);
+		/* XXX: should maxwait include replication connections? */
```
JelteF, line 893: Conceptually it would make sense to include replication connections too, but I also don't think it's super important.

doc/config.md
```diff
@@ -1249,7 +1249,7 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
 (see <https://www.postgresql.org/docs/current/auth-pg-hba-conf.html>).
 
 * Supported record types: `local`, `host`, `hostssl`, `hostnossl`.
-* Database field: Supports `all`, `sameuser`, `@file`, multiple names.  Not supported: `replication`, `samerole`, `samegroup`.
+* Database field: Supports `all`, `replication` `sameuser`, `@file`, multiple names.  Not supported: `samerole`, `samegroup`.
```
petere, line 1252: comma

include/bouncer.h
```diff
@@ -490,6 +507,12 @@ struct PgDatabase {
 	struct AATree user_tree;	/* users that have been queried on this database */
 };
 
+enum ReplicationType {
+	REPLICATION_NONE,
```
petere, line 511: Maybe set this one to `= 0`, so this can be used as a boolean.

include/bouncer.h
```diff
@@ -54,6 +54,8 @@ enum SocketState {
 	CL_LOGIN,		/* login_client_list */
 	CL_WAITING,		/* pool->waiting_client_list */
 	CL_WAITING_LOGIN,	/*   - but return to CL_LOGIN instead of CL_ACTIVE */
+	CL_WAITING_LOGICAL_REP,	/* pool->waiting_logical_replication_list */
+	CL_WAITING_PHYSICAL_REP,	/* pool->waiting_physical_replication_list */
```
petere, line 58: Why do we need separate lists for logical and physical replication?

JelteF: I guess you're right and we probably don't need that (after changing a few things).

JelteF:
I fixed all the things I wanted to in this PR. So I think it's ready for a more thorough review now.

.editorconfig
```diff
@@ -8,7 +8,7 @@ insert_final_newline = true
 charset = utf-8
 trim_trailing_whitespace = true
 
-[*.{c,h}]
+[*.{c,h,eval,rules}]
```
petere, line 11: This had me confused. What is *.eval? Maybe list the full file names here. And make it a separate stanza from the C-code related things.

include/common/postgres_compat.h
```diff
@@ -30,6 +32,8 @@
 #define pg_sha256_update(ctx, data, len) sha256_update(ctx, data, len)
 #define pg_sha256_final(ctx, dst) sha256_final(ctx, dst)
 
+extern bool parse_bool(const char *value, bool *result);
+extern bool parse_bool_with_len(const char *value, size_t len, bool *result);
```
petere, line 36: Isn't this redundant with `include/common/builtins.h`?

include/bouncer.h
```diff
@@ -287,6 +289,13 @@ struct PgPool {
 	 */
 	struct StatList waiting_client_list;
 
+	/*
+	 * Clients that are waiting for a physical/logical replication connection to be
+	 * available. These requests are waiting for a new server connection to be
+	 * opened, before the request can be forwarded.
+	 */
+	struct StatList waiting_replication_client_list;
```
petere, line 297: Why is there a separate list for this? Why aren't these just part of the normal waiting_client_list? Maybe there is a reason, but it could be explained here. I mean, there is no separate list of active replication clients, so why a separate list of waiting replication clients?

test/hba_test.rules
```diff
@@ -5,10 +5,6 @@
 # hostssl    DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
 # hostnossl  DATABASE  USER  ADDRESS  METHOD  [OPTIONS]
 
-# ws
- 	 	 	
-			# z
-
```
petere, line 11: Wasn't this here for some parsing test?

JelteF: I couldn't gather that from the commit message that added the file 8 years ago. The lines seemed like nonsense to me and they were causing diff when I saved because they had trailing whitespace that was removed by the editorconfig settings.

I added them back now and disabled the trailing whitespace removal for this file.

include/bouncer.h
JelteF: Good question. I added the reason to the comment now (also to the one for cancel connections since it's the same reason).

test/test_replication.py
```diff
@@ -0,0 +1,250 @@
...
+        assert process.returncode == 1
+    else:
+        assert process.returncode == 0
+
+    children = list(wal_dump_dir.iterdir())
+    assert len(children) > 0
+
+
+@pytest.mark.skipif(
+    "PG_MAJOR_VERSION < 10", reason="pg_basebackup on PG9.6 fails with a weird error"
```
eulerto, line 203: I'm curious about the "weird error". Is it a random failure? Or it is because of a behavior change (it streams WAL by default)?

eulerto (commented):
This is _not_ a complete review. I want to test some cases to see if I can break it. One suggestion is to add a test to exercise the `waiting_replication_client_list`.

test/test_replication.py
JelteF: I took a bit closer look and I was able to fix it by the flag `-X stream`.

test/test_replication.py
```diff
@@ -0,0 +1,250 @@
...
+    # loop. Because it will receive its own changes, because we're checking for
+    # updates on the same table that we're inserting into. After this short sleep there should be at least 3 rows:
+    # 1. the original one that was already in the table
+    # 2. one from the table sync worker
+    # 3. many (but at least one) from the logical replication infinite loop
+    # Since we don't know exactly how many will be in there, we just check that
+    # its three or more.
+    assert (
+        bouncer.sql_value("SELECT count(*) FROM test_logical_rep_subscriber.table") >= 3
+    )
```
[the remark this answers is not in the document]
JelteF: Thanks for the suggestion to use different databases for the source and target. I agree that makes the test much easier to understand.

src/server.c
```diff
@@ -479,12 +480,12 @@ static bool handle_connect(PgSocket *server)
 	} else {
 		/* proceed with login */
 		if (server_connect_sslmode > SSLMODE_DISABLED && !is_unix) {
-			slog_noise(server, "P: SSL request");
+			slog_debug(server, "P: SSL request");
```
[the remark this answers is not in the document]
JelteF: Reverted this change.

src/server.c
```diff
@@ -619,7 +620,9 @@ bool server_proto(SBuf *sbuf, SBufEvent evtype, struct MBuf *data)
 			break;
 		}
 
-		if (pool_pool_mode(pool) != POOL_SESSION || server->state == SV_TESTED || server->resetting) {
+		if (server->state == SV_TESTED || server->resetting || (
+			    pool_pool_mode(pool) != POOL_SESSION && !server->replication
+			    )) {
```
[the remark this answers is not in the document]
JelteF: Removed this change. I think I added it for debugging purposes and it's not actually necessary.

include/bouncer.h
```diff
@@ -301,10 +303,30 @@ struct PgPool {
...
 
+	/*
+	 * Clients that are waiting for a physical/logical replication connection to be
+	 * available. These requests are waiting for a new server connection to be
+	 * opened, before the request can be forwarded.
+	 *
+	 * This is a separate list from waiting_client_list, because we need to trigger
+	 * opening a completely new server connection for them. The clients in
+	 * waiting_client_list can also use a currently idle server connection.
+	 *
```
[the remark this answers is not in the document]
JelteF: I rephrased it, I hope it's clearer now.

include/bouncer.h
JelteF: I actually removed the state completely now. It turned out not to be necessary (given some other changes).

include/bouncer.h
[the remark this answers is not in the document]
JelteF: Actually... I deleted the CL_WAITING_REPLICATION state completely now. I did still update the description of the wait_cancel_req_list to hopefully make it clearer.

eulerto:
Isn't the cancel key change a requirement to this feature? I suggest you to create a separate commit for it.

eulerto:
> Isn't the cancel key change a requirement to this feature? I suggest you to create a separate commit for it.

Ops... I realized that you did it. Another suggestion is to explain in the commit message that it is a requirement for this feature.

JelteF:
> Ops... I realized that you did it. Another suggestion is to explain in the commit message that it is a requirement for this feature.

I added a commit message and moved it before the actual replication change, but while doing that I realized I thought the change was special enough to warrant its own PR, so I made a new one: https://github.com/pgbouncer/pgbouncer/pull/945

.editorconfig
petere, line 1: I'm not sure these changes are necessary. Can we just leave it alone?

include/common/postgres_compat.h
```diff
@@ -30,6 +32,5 @@
 #define pg_sha256_update(ctx, data, len) sha256_update(ctx, data, len)
 #define pg_sha256_final(ctx, dst) sha256_final(ctx, dst)
 
-
```
petere, line 33: Let's keep that.

include/common/postgres_compat.h
```diff
@@ -15,6 +16,7 @@
 
 #define lengthof(array) (sizeof (array) / sizeof ((array)[0]))
 #define pg_hton32(x) htobe32(x)
+#define pg_strncasecmp strncasecmp
```
petere, line 19: These aren't necessarily the same thing. `pg_strncasecmp()` only case-folds ASCII characters. Maybe that's ok for current use? But we should be careful about documenting this.

src/hba.c
```diff
@@ -31,8 +31,9 @@ enum RuleType {
 	RULE_HOSTNOSSL,
 };
 
-#define NAME_ALL        1
-#define NAME_SAMEUSER   2
+#define NAME_ALL                1
+#define NAME_SAMEUSER           2
+#define NAME_REPLICATION        4
```
petere, line 36: We should comment here that this is a bit mask. It wasn't clear from the first two values, but with the third one it becomes relevant.

src/proto.c
```diff
@@ -584,15 +593,69 @@ bool answer_authreq(PgSocket *server, PktHdr *pkt)
...
+		 * because it allows us to skip running the SET logic
+		 * completely, which normally requires waiting on multiple
+		 * server responses. This SET logic is normally executed in the
+		 * codepath where we link the client to the server
+		 * (find_server), but because we link the client here already
+		 * we don't run that code for replication connections. Adding
+		 * the varcache parameters to the startup message allows us to
+		 * skip the dance that involves sending Query packets and
+		 * waiting for responses.
+		 */
```
petere, line 630: Could we also apply this to normal connections, to get the same optimization benefit?

src/server.c
```diff
@@ -122,7 +122,18 @@ static bool handle_server_startup(PgSocket *server, PktHdr *pkt)
...
+		/*
+		 * If we haven't been able to connect to the server since the
+		 * startup (or call to tag_pool_dirty), then we drop all
+		 * clients that are currently trying to log in because they
+		 * will almost certainly hit the same error.
+		 *
+		 * However, we don't do this if it's a replication connection,
+		 * because those can fail due to a variety of missing
+		 * permissions, while normal connections would still be able
+		 * connect and query the database just fine.
```
petere, line 134: I don't understand this distinction. Can you give examples?

include/bouncer.h
```diff
@@ -305,6 +306,11 @@ struct PgPool {
 	 * Clients that sent cancel request, to cancel another client its query.
 	 * These requests are waiting for a new server connection to be opened,
 	 * before the request can be forwarded.
+	 *
+	 * This is a separate list from waiting_client_list, because we want to
```
petere, line 310: Is this related to this patch?

src/server.c
```diff
@@ -192,6 +203,18 @@ static bool handle_server_startup(PgSocket *server, PktHdr *pkt)
...
+ * forces session pooling if the server is a replication connection, because
+ * replication connections require session pooling to work correctly.
+ */
+int server_pool_mode(PgSocket *server)
+{
+	if (server->replication)
+		return POOL_SESSION;
+	return pool_pool_mode(server->pool);
+}
+
```
petere, line 217: I'm worried that if we have both `server_pool_mode()` and still the old `pool_pool_mode()`, it will be difficult to enforce that the right one is called in each situation.

I wonder if we could alleviate that by allocating a `PgPool` structure for each replication connection, but only allow one connection through it. Then we can keep using `pool_pool_mode()`, and maybe also have fewer changes elsewhere.  (For example, this patch doesn't appear to have anything for statistics or `SHOW` commands. I wonder if this still works sensibly for connections that are not associated with pools. And of course, we do want some sort of monitoring support in any case.)

petere (dismissed):

src/proto.c
JelteF: Then we would need to link normal connections here already too. We don't do that currently. I'd expect that to cause some problems if we did that. It might be possible, but I don't think it fits well in this PR to do so.

.editorconfig
JelteF: I extracted this and moved it to: https://github.com/pgbouncer/pgbouncer/pull/1017

include/bouncer.h
JelteF: I extracted this small change and moved it to: https://github.com/pgbouncer/pgbouncer/pull/1016

src/server.c
JelteF: It's a reasonable worry, because I could definitely see that be an issue. To avoid the issue I added a warning to the function comment of `pool_pool_mode()` and renamed it to `probably_wrong_pool_pool_mode()`, because it should basically never be used directly (the only place where it should be used is in `SHOW pools`).

I did that instead of your proposal, because I do want to keep the current general design. The reason for that is that this is more in line with the future plan I have of allowing a client to choose if they want transaction or session pooling: https://github.com/pgbouncer/pgbouncer/issues/997 Basically for that near-future plan the pooling mode would be a property of the connection, not of the pool. Just like it is a property of the connection now for replication connections. 

Regarding statistics, I think it would make sense to add a replication and pool_mode column to `show servers/clients`. But the existing reporting also works for replication connections.

src/janitor.c
```diff
@@ -534,13 +541,13 @@ static void pool_server_maint(PgPool *pool)
 	check_unused_servers(pool, &pool->idle_server_list, 1);
 
 	/* disconnect close_needed active servers if server_fast_close is set */
```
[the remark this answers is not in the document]
JelteF: done

src/server.c
JelteF: FYI I added a replication column to `SHOW CLIENTS/SERVERS`.  I think that should be enough info for debugging any issues.

PR #878, opened 2023-07-03
Add support for options startup parameter

### LINKED ISSUE #875: Support options startup parameter
You can specify any postgres parameter in the `options` startup packet #867. But PgBouncer doesn't support it. It would be great if it could parse those parameters and set them on the connection:
```
PGOPTIONS='-c search_path=abc' psql -p 6432
psql: error: connection to server at "localhost" (127.0.0.1), port 6432 failed: FATAL:  unsupported startup parameter: options
```

This would mostly become relevant after merging #867 for transaction mode. But in session mode this would already be useful now.

JelteF:
It's allowed to specify any Postgres command line parameter in the `options`
startup packet #867. This is commonly used to specify GUC values on connection
startup, using the `-c key=value` command line flag. I believe the main reason
this is used instead of specifying the desired GUC as a startup parameter
directly is because libpq does not allow specifying arbitrary startup
parameters.

This PR adds support for configuring GUCs using the `-c` flag in the options
startup packet. While in theory people can specify any flag they want. In
practice the `-c` flag is the only one that I know is commonly used. Probably
because it is the only one that's documented in the Postgres docs for
`PGOPTIONS`. Any of the other command line flags are simply shorthands for
specific GUCs, and so even when those would be used they can be easily rewritten
using `-c`. So to keep parsing of the options startup parameter simple, this
only supports the `-c` flag.

Fixes #875

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -548,6 +548,84 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
 	return true;
 }
 
+static inline const char *skip_ws(const char *position)
+{
+	while (*position && isspace(*position))
+		position++;
+	return position;
+}
+
+_MUSTCHECK
+static bool get_escaped_string(struct MBuf *buf, const char **position_ptr)
+{
+	const char *position = *position_ptr;
+	const char *unwritten_start = position;
+	while (*position)
+	{
+		if (*position == '\\') {
+			if (!mbuf_write(buf, unwritten_start, position-unwritten_start))
+				return false;
+			position++;
+			unwritten_start = position;
+			if (!*position)
+				break;
+		} else if (isspace(*position)) {
+			break;
+		}
+		position++;
+	}
+	if (!mbuf_write(buf, unwritten_start, position-unwritten_start))
+		return false;
+	if (!mbuf_write_byte(buf, '\0'))
+		return false;
+	*position_ptr = position;
+	return true;
+}
+
+static bool set_startup_options(PgSocket *client, const char *options)
+{
+	char arg_buf[400];
+	struct MBuf arg;
+	const char *position = options;
+	mbuf_init_fixed_writer(&arg, arg_buf, sizeof(arg_buf));
+	slog_error(client, "received options: %s", options);
+	while (*position) {
+		const char *key_string, *value_string;
+		char *equals;
+		mbuf_rewind_writer(&arg);
+		position = skip_ws(position);
+		if (strncmp("-c", position, 2) != 0)
+			goto fail;
+		position += 2;
+		position = skip_ws(position);
+
+		if (!get_escaped_string(&arg, &position)) {
+			disconnect_client(client, true, "unsupported options startup parameter: parameter too long");
+			return false;
+		}
+
+		equals = strchr((char *) arg.data, '=');
+		if (!equals)
+			goto fail;
+		*equals = '\0';
+
+		key_string = (const char *) arg.data;
+		value_string = (const char *) equals + 1;
+		if (varcache_set(&client->vars, key_string, value_string)) {
+			slog_debug(client, "got var from options: %s=%s", key_string, value_string);
+		} else if (strlist_contains(cf_ignore_startup_params, key_string)) {
+			slog_debug(client, "ignoring startup parameter from options: %s=%s", key_string, value_string);
+		} else {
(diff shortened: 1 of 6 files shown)


src/client.c
```diff
@@ -548,6 +548,106 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
...
+
+		key_string = (const char *) arg.data;
+		value_string = (const char *) equals + 1;
+		if (varcache_set(&client->vars, key_string, value_string)) {
+			slog_debug(client, "got var from options: %s=%s", key_string, value_string);
+		} else if (strlist_contains(cf_ignore_startup_params, key_string)) {
+			slog_debug(client, "ignoring startup parameter from options: %s=%s", key_string, value_string);
+		} else {
+			slog_warning(client, "unsupported startup parameter in options: %s=%s", key_string, value_string);
+			disconnect_client(client, true, "unsupported startup parameter: %s", key_string);
```
emelsimsek, line 642: Should this be changed to

`goto fail;`
or
`return false;`

emelsimsek:
Looks good overall.

test/test_misc.py
```diff
@@ -132,3 +132,61 @@ async def test_host_list(bouncer):
...
+    ):
+        bouncer.test(options="-c timezone")
+
+    too_long_param = "a" * 1000
+
+    with pytest.raises(
+        psycopg.OperationalError,
+        match="unsupported options startup parameter: parameter too long",
+    ):
+        bouncer.test(options="-c timezone=" + too_long_param)
```
emelsimsek, line 192: How about adding a test case for a variable that is neither tracked nor ignored in the config:

-c extra_float_digits=3

to exhaustively test the failure paths.

src/client.c
```diff
@@ -548,6 +548,107 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
...
+
+		if (!read_escaped_token(&position, &arg)) {
+			disconnect_client(client, true, "unsupported options startup parameter: parameter too long");
+			return false;
+		}
+
+		equals = strchr((char *) arg.data, '=');
+		if (!equals)
+			goto fail;
+		*equals = '\0';
```
eulerto, line 632: It is a matter of style but I will suggest anyway.

`*equals++ = '\0';`

and remove `+ 1` 2 lines below.

src/client.c
```diff
@@ -548,6 +548,107 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
 	return true;
 }
 
+/*
+ * skip_ws returns a pointer to the first non whitespace character
+ * in the given string
```
eulerto, line 553: s/string/string./

src/client.c
```diff
@@ -548,6 +548,107 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
...
+ *
+ * It only supports the following type of postgres command line argument:
+ * -c config=value
+ *
+ * The reason that we don't support all arguments is to keep the parsing simple
+ * an this is by far the argument that's most commonly used in practice in the
+ * options startup paramater. Also all other postgres command line arguments
+ * can be rewritten to this form:
+ */
+static bool set_startup_options(PgSocket *client, const char *options)
```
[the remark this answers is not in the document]
JelteF: Sadly the parsing works slightly differently for the `options` string and the connections string. The options string uses `\` to escape whitespace characters, while the connection string quotes values in single quotes.

src/client.c
```diff
@@ -548,6 +548,107 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt) {
...
+		return false;
+	if (!mbuf_write_byte(unescaped_token, '\0'))
+		return false;
+	*escaped_string_ptr = position;
+	return true;
+}
+
+/*
+ * set_startup_options takes the value of to the "options" startup parameter
+ * and uses it to set the parameters that are embeded in this value.
```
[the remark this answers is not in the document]
JelteF: FYI for next time: You can do such suggestions using the Github interface, which allows easily applying them
```suggestion
 * and uses it to set the parameters that are embedded in this value.
```

src/client.c
JelteF: I'll leave this as is, I think it's clearer this way. Mostly because the variable is called `equals`, thus logically pointing to the equals character. If it was the `position` variable I'd have agreed with the suggestion.

PR #879, opened 2023-07-03
Update client varcache with canonical values from server

### LINKED ISSUE #776: Queries SET DateStyle='ISO'; appeared in every transaction after migration to pg 14
Hi, I'm not sure if I'm opening discussion in correct place, but I will really appreciate any help, advice or guidance. 

We are using pgbouncer in transaction pooling mode in our setup and it's looking like following : app ---> pgbouncer ---> Postgres. 
Postgres and pgbouncer are running on same server on Rocky OS. 

Applications are using PostgreSQL JDBC Driver 42.2.24 or 42.3.3 mainly. 
Pgbouncer package is used 1.16.1-10.rhel8.
And Postgres was migrated from 12 version to 14 version. 

After Migration to Postgres 14 we are starting to observe queries: SET DateStyle='ISO'; on Postgres side. 
So, because of this situation we are getting additional Query calls to our Postgres 14 cluster from pgbouncer side. 

Can you please tell us if it is intended to be so? On Postgres 12 we didn't monitor such Queries. 
So, after migration to Postgres 14 version nothing was changed on pgbouncer configuration side.  Non default parameters that we are using are: 

```
[databases]
* = host=127.0.0.1


[pgbouncer]
logfile = /var/log/postgresql/pgbouncer.log
pidfile = /var/run/postgresql/pgbouncer.pid
listen_addr = *
listen_port = 6432
unix_socket_dir = /var/run/postgresql
auth_type = md5
admin_users = pgbouncer
stats_users = pgbouncer,okagent
pool_mode = transaction
server_connect_timeout = 300
max_client_conn = 10000
default_pool_size = 10
query_wait_timeout = 300
log_connections = 0
log_disconnections = 0
```

DateStyle parameter on Postgres side after migration was not changed as well and it is: 

```
$ psql
psql (14.5)
Type "help" for help.

13:50:42 [postgres] # show datestyle;
 DateStyle
-----------
 ISO, MDY
(1 row)
```

I was trying to use datestyle parameter in pgbouncer.ini file explicitly
```
[databases]
* = host=127.0.0.1 datestyle='ISO, MDY'
```

but this didn't changed anything.

Also I was trying to upgrade pgbouncer to 1.17.0-10.rhel8, this didn't worked either. Query SET DateStyle='ISO'; remains. 

With verbose=2 parameter in pgbouncer.ini we are able to see records like these in the pgbouncer.log: 
```
2022-11-09 16:24:05.811 MSK [2997433] DEBUG S-0x563a4975b8d0: dbuser/dbname@127.0.0.1:5432 varcache_apply: SET DateStyle='ISO';
```

On database side logs it's looking like so: 
```
2022-11-09 14:53:01.199 UTC,"dbuser","dbname",3182464,"127.0.0.1:19074",636bb713.308f80,748,"SET",2022-11-09 14:20:03 UTC,14/0,0,LOG,00000,"duration: 0.090 ms  statement: SET DateStyle='ISO';",,,,,,,,,"PostgreSQL JDBC Driver","client backend",,-7713842022241034850
```
Just in case this is also matter pg_stat_statements extension is used. 
After postgres upgrade from version 12 to 14 it's updated from 1.6 to 1.9.

JelteF:
Since PG14, Postgres doesn't send ParameterStatus packets anymore when a
setting has not actually changed. This could cause the client varcache
and server varcache to become out of sync. Which then in turn resulted
in unnecessary `SET` commands to be sent to the server.

Fixes #776

### DIFF
--- a/include/varcache.h
+++ b/include/varcache.h
@@ -23,3 +23,4 @@ void varcache_fill_unset(VarCache *src, PgSocket *dst);
 void varcache_clean(VarCache *cache);
 void varcache_add_params(PktBuf *pkt, VarCache *vars);
 void varcache_deinit(void);
+void varcache_set_canonical(PgSocket *server, PgSocket *client);
--- a/src/server.c
+++ b/src/server.c
@@ -613,6 +613,18 @@ bool server_proto(SBuf *sbuf, SBufEvent evtype, struct MBuf *data)
 			PgSocket *client = server->link;
 			Assert(client);
 
+			/*
+			 * It's possible that the client vars and server vars have
+			 * different string representations, but still Postgres did not
+			 * send a ParamaterStatus packet. This happens when the server
+			 * variable is the canonical version of the client variable, i.e.
+			 * they mean the same just written slightly different. To make sure
+			 * that the canonical version is also stored in the client, we now
+			 * copy the server variables over to the client variables.
+			 * See issue #776 for an example of this.
+			 */
+			varcache_set_canonical(server, client);
+
 			server->setting_vars = false;
 			sbuf_continue(&client->sbuf);
 			break;
--- a/src/varcache.c
+++ b/src/varcache.c
@@ -233,6 +233,24 @@ bool varcache_apply(PgSocket *server, PgSocket *client, bool *changes_p)
 	return pktbuf_send_immediate(pkt, server);
 }
 
+void varcache_set_canonical(PgSocket *server, PgSocket *client)
+{
+	struct PStr *server_val, *client_val;
+	const struct var_lookup *lk, *tmp;
+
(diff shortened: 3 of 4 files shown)


JelteF:
I did some basic benchmarks with against PG15 using:
```
PGOPTIONS='-c datestyle=ISO' pgbench -p 6432 -T 20 -P 1 --select-only 'host=localhost port=6432'
```
And the TPS is as follows on my machine:

- master: 1970
- this PR: 2803 (42% faster)

If I introduce 10ms of artificial latency between PgBouncer and Postgres then it goes from:
- master: 45
- this PR: 90 (100% faster)

emelsimsek:
I tested this fix with PG 16.0.  It looks good from functional perspective.

PR #881, opened 2023-07-06
Add cf_dns_close_previous option

lukmajercak:
_**This is just a proposal, needs full implementation**_

## What
Currently, when a host is resolved, pgbouncer checks all server connections for this host and tags connections using IPs not found in the new resolved list as dirty (`close_needed`). I'm proposing to add a configuration option that would disable this behavior. This would be helpful in avoiding connection churn in scenarios where, for example, the host might return a list of different IP addresses (think weighted routing in Route53) every time.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -628,6 +628,7 @@ extern int cf_disable_pqexec;
 extern usec_t cf_dns_max_ttl;
 extern usec_t cf_dns_nxdomain_ttl;
 extern usec_t cf_dns_zone_check_period;
+extern int cf_dns_close_previous;
 extern char *cf_resolv_conf;
 
 extern int cf_auth_type;
--- a/src/dnslookup.c
+++ b/src/dnslookup.c
@@ -1401,7 +1401,7 @@ static void got_result_gai(int result, struct addrinfo *res, void *arg)
 		req->result = res;
 		req->current = res;
 
-		if (req->oldres)
+		if (req->oldres && cf_dns_close_previous)
 			check_req_result_changes(req);
 
 		/* show all results */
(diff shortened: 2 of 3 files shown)


doc/config.md
```diff
@@ -608,6 +608,12 @@ Works only with UDNS and c-ares backends (`configure` option `--with-udns` or `-
 
 Default: 0.0 (disabled)
 
+### dns_close_previous
```
JelteF, line 611: bikeshed: I think `dns_close_previous` is not a great name for this option. I think `removed` or `missing` is a better suffix.
```suggestion
### dns_close_removed
```

JelteF (commented):
I think this is a reasonable feature to want, and the amount of code needed for it is minimal. If there's an easy way to add tests for this that would be nice, but I guess not because it relies on changes being made to the DNS.

Needs a rebase though

Avinodh:
@JelteF / @lukmajercak  - Running into a related issue where pgbouncer connecting to a route53 endpoint behind which there are multiple Readers causes `min_pool_size` to not be honored. 

This is because every-time the Route53 returns a different IP to pgbouncer, pgbouncer sets `close_needed` on every other server connection in the current pool. This causes any `idle` connections to be churned, thereby preventing pgbouncer from establishing up to `min_pool_size` connections. 

Do we have any other workaround for this at the moment?

lukmajercak:
> @lukmajercak - I patched and tested this fix out, and looks like it works for our use case. Thank you!
> 
> Unless you were still planning to merge this, I can send out a similar PR to have this merged!

Sorry for the late response. Feel free to send out the other PR

PR #886, opened 2023-07-10
Make varcache case insensitive.

emelsimsek:
Override the HASH_FUNCTION and HASH_CMP with case insensitive ones for varcache.

This will enable 
   1. The clients to send case insensitive parameters in startup options making it compatible with how postgres handles them.
   2. track_extra_parameters values to be case insensitive.

Considered using a list instead, but the use of a hash seemed more advantageous.
   - It simplifies creating the initial lookup table handling the duplicate definitions etc.
   - It is faster when updating the parameter values.

### DIFF
--- a/include/varcache.h
+++ b/include/varcache.h
@@ -1,50 +1,5 @@
 #include <usual/strpool.h>
 
-#undef HASH_FUNCTION
-#define HASH_FUNCTION(keyptr,keylen,hashv) HASH_JEN_LOWERCASE(keyptr, keylen, hashv)
-
-#define HASH_JEN_LOWERCASE(key,keylen,hashv) \
-do {                                                                             \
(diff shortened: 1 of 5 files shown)


include/varcache.h
```diff
@@ -1,5 +1,50 @@
 #include <usual/strpool.h>
 
+#undef HASH_FUNCTION
+#define HASH_FUNCTION(keyptr,keylen,hashv) HASH_JEN_LOWERCASE(keyptr, keylen, hashv)
+
+#define HASH_JEN_LOWERCASE(key,keylen,hashv) \
```
JelteF, line 6: Let's put this define in a separate file: `include/common/uthash_lowercase.h`

That way we can include the original license. 

Also let's add a comment explaining that this is a copy of HASH_JEN, but with an added `tolower` call for every place where you get a character.

JelteF (approved):

PR #891, opened 2023-07-13
Correct editorconfig rules for C files

petere:
Let's discuss the indentation rules for C files again. I disagree with the existing text in .editorconfig that basically everybody can do what they want, which seems kind of insane. ;-) The actual rules for pgbouncer source code has always been tab width 8 and indent by tab, so I suggest we put those in here. Note also that the whitespace rules in .gitattributes are set for this.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,15 +10,12 @@ trim_trailing_whitespace = true
 
 [*.{c,h}]
 indent_style = tab
-# We specifically don't set indent_size and tab_with here so people can choose
-# what they prefer themselves
+indent_size = 8
 
 [*.py]
 indent_style = space
 indent_size = 4
-tab_width = 4
 
 [*.{yml,sh}]
 indent_style = space
 indent_size = 2
-tab_width = 2


.editorconfig
```diff
@@ -10,8 +10,8 @@ trim_trailing_whitespace = true
 
 [*.{c,h}]
 indent_style = tab
-# We specifically don't set indent_size and tab_with here so people can choose
-# what they prefer themselves
+indent_size = tab
+tab_width = 8
```
JelteF, line 14: I took a bit of time to actually read the editorconfig docs on the difference between `indent_size` and `tab_width`. And it turns out you only need to set `indent_size` here:

> tab_width: a whole number defining the number of columns used to represent a tab character. This defaults to the value of indent_size and doesn't usually need to be specified.

```suggestion
indent_size = 8
```

petere: Done, and also removed the other tab_width settings that don't seem necessary.

petere:
> (the gitattributes don't seem to specify a width of 8 either btw, I'm not sure if they can)

8 is the default.

> So overall, I think this change makes sense for now. But in the near future I would like to set up some automatic codeformatting. And preferably one that indents using tabs, and aligns using spaces (instead of the current approach of aligning with both spaces and tabs). That way people would be able to set their own tabwidth according to their preference.

I would welcome some automatic formatting support. I think we could consider using pgindent. It seems much easier to use now.

PR #892, opened 2023-07-14
Add autoformatting to the PgBouncer source code

JelteF:
This uses uncrustify with a config that matches the code style of the repo as
close as I was able to configure it.

Codestyle is always an easy thing to disagree on, but I think overall this
improves the styling. It also keeps the fairly small.

I also considered using pgindent, but that requires a typedefs.list which is a
pain to create. The current codestyle of PgBouncer also doesn't match the one
from Postgres at all, so it would cause many uninteresting changes.


src/sbuf.c
```diff
@@ -359,8 +359,8 @@ void sbuf_prepare_fetch(SBuf *sbuf, unsigned amount)
 }
 
 /*************************
- * Internal functions
- *************************/
+* Internal functions
+*************************/
```
JelteF, line 363: this seems unintended

include/util.h
```diff
@@ -25,17 +25,17 @@
...
 int log_socket_prefix(enum LogLevel lev, void *ctx, char *dst, unsigned int dstlen);
 
-#define slog_error(sk, args...) log_generic(LG_ERROR, sk, ## args)
-#define slog_warning(sk, args...) log_generic(LG_WARNING, sk, ## args)
-#define slog_info(sk, args...) log_generic(LG_INFO, sk, ## args)
-#define slog_debug(sk, args...) do { \
+#define slog_error(sk, args ...) log_generic(LG_ERROR, sk, ## args)
+#define slog_warning(sk, args ...) log_generic(LG_WARNING, sk, ## args)
+#define slog_info(sk, args ...) log_generic(LG_INFO, sk, ## args)
+#define slog_debug(sk, args ...) do { \
```
eulerto, line 31: Does it worth it? IIRC C99 changed this function-like macros.

src/dnslookup.c
```diff
@@ -606,7 +606,7 @@ struct SOA {
 	uint32_t dnssoa_minttl;
 };
 
-typedef void query_soa_fn(struct dns_ctx *ctx, struct SOA *result, void *data);
+typedef void query_soa_fn (struct dns_ctx *ctx, struct SOA *result, void *data);
```
eulerto, line 609: It doesn't seem consistent because there is no space between function name and arguments.

src/pam.c
```diff
@@ -317,7 +317,7 @@ static int pam_conversation(int msgc,
 
 	rc = PAM_SUCCESS;
 
-	for (i=0; i<msgc; i++) {
+	for (i=0; i < msgc; i++) {
```
eulerto, line 320: Isn't it replace `i=0` with `i = 0`?

eulerto (commented):
It is not a complete review because it is easy to bikeshed on topics like this.

include/bouncer.h
```diff
@@ -366,18 +372,18 @@ struct PgPool {
...
 	VarCache orig_vars;		/* default params from server */
 
 	usec_t last_lifetime_disconnect;/* last time when server_lifetime was applied */
 
 	/* if last connect to server failed, there should be delay before next */
 	usec_t last_connect_time;
-	bool last_connect_failed:1;
-	bool last_login_failed:1;
+	bool last_connect_failed : 1;
+	bool last_login_failed : 1;
```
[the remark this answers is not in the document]
JelteF: Yeah, I also would have preferred to keep this as is, simply for consistency. But I also couldn't find an option to keep the previous styling. IMHO this new formatting is also acceptable though.

include/bouncer.h
```diff
@@ -66,13 +66,13 @@ enum SocketState {
 	SV_ACTIVE,		/* pool->active_server_list */
 	SV_ACTIVE_CANCEL,	/* pool->active_cancel_server_list */
 	SV_USED,		/* pool->used_server_list */
-	SV_TESTED		/* pool->tested_server_list */
+	SV_TESTED,	/* pool->tested_server_list */
```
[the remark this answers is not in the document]
JelteF: I personally stopped caring about pretty much all aspects of codestyling. As long as it's consistent and automatically enforced I'm happy. 

The only exception to that is trailing/dangling commas. I think those should be used whenever the language allows them to be (when items are on multiple lines). They keep future diffs clean, and allows reordering of lines without problems. This blog outlines my viewpoint on this pretty well:  https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8

src/janitor.c
```diff
@@ -351,7 +350,7 @@ void per_loop_maint(void)
 		} else {
 			active_count += statlist_count(&login_client_list);
 		}
-		/* fallthrough */
+	/* fallthrough */
```
JelteF, line 353: This seems suboptimal. Let me see if I can fix that.

petere: Keeping the old indenting would be better, but I'm not too bothered by it either way.

src/main.c
```diff
@@ -78,7 +78,7 @@ struct HBA *parsed_hba;
 
 int cf_daemon;
 int cf_pause_mode = P_NONE;
-int cf_shutdown; /* 1 - wait for queries to finish, 2 - shutdown immediately */
+int cf_shutdown;/* 1 - wait for queries to finish, 2 - shutdown immediately */
```
petere, line 81: strange

src/objects.c
```diff
@@ -2021,19 +2018,19 @@ void for_each_server(PgPool *pool, void (*func)(PgSocket *sk))
 	struct List *item;
 
 	statlist_for_each(item, &pool->idle_server_list)
-		func(container_of(item, PgSocket, head));
+	func(container_of(item, PgSocket, head));
```
petere, line 2021: This looks confusing after the reformatting. Maybe putting braces around the loop body would help?

include/bouncer.h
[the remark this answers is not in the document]
petere: I do like this change, but maybe it could be put into a separate commit? The first commit should just normalize the existing formatting, then we can evolve it.

.gitmodules
```diff
@@ -4,3 +4,6 @@
 [submodule "uthash"]
 	path = uthash
 	url = https://github.com/troydhanson/uthash.git
+[submodule "uncrustify"]
```
petere, line 7: I would rather have fewer submodules than more. But apart from that, pulling this in this way would include it in the distributed source tarballs, which isn't good. Also, it's GPL. So we should really keep it at arm's length. Better just install it as a normal build dependency.

Makefile
```diff
@@ -177,3 +177,21 @@ htmls:
 
 doc/pgbouncer.1 doc/pgbouncer.5:
 	$(MAKE) -C doc $(@F)
+
+indent-check: uncrustify/build/uncrustify
+	flake8
```
petere, line 182: Seems like flake8 isn't really related to indenting.

uncrustify.cfg
```diff
@@ -0,0 +1,83 @@
...
+sp_inside_for = remove
+sp_after_sparen = force
+sp_sparen_brace = force
+sp_do_brace_open = force
+sp_while_paren_open = force
+sp_before_semi_for = remove
+sp_after_semi_for = force
+sp_inside_square = remove
+sp_after_comma = force
+sp_after_cast = ignore  # XXX: We're not very consistent with this, do we want to be?
```
petere, line 52: I would like the setting "add" here (or "force", what's the difference?). But it's not how the current code is formatted, mostly. I suggest leaving it as is, with the comment, and we can address it as a follow-up.

.flake8
```diff
@@ -1,3 +1,8 @@
 [flake8]
 extend-ignore = E203, E265, F401, F541
 max-line-length = 150
+exclude =
+    .git,
+    __pycache__,
+    lib,
```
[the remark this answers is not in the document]
JelteF: The uncrustify repo had a python file that failed the tests. So yes.

.cirrus.yml
```diff
@@ -260,14 +260,14 @@ task:
       type: text/plain
 
 task:
-  name: Python check
+  name: Formatting check
```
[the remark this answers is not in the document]
JelteF: Did you mean Python and C separate? and/or did you mean separating `flake8` (which does not format automatically) from `isort` & `black`?

pyproject.toml
```diff
@@ -19,9 +19,21 @@ timeout = 30
 # versions is a small price to pay to not have to worry about that.
 asyncio_mode = 'auto'
 
+# Make test discovery quicker
+norecursedirs = [
+    '*.egg',
```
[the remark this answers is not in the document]
JelteF: uncrustify directory had a python file that caused some issues with this I think

uncrustify.cfg
JelteF: Sounds good. 

> "add" here (or "force", what's the difference?)

FYI: `force` means make it 1 space. `add` means add 1 space when there are none, but if there is already spacing leave as is So if you have 3 spaces, then force makes it 1, but add leaves as 3.

.gitmodules
JelteF: The main problem with installing a system uncrustify version through `apt`/`yum`/`brew` is that it is important that everyone uses the exact same version. A newer version might have a formatting bug fixed and result in slightly different formatting. So a submodule seemed like a reasonable way to ensure a consistent version. (for black this is less of an issue, because we can just say we should use the latest version, which is very easy to install using `pip`)

>  But apart from that, pulling this in this way would include it in the distributed source tarballs, which isn't good. 

You mean the ones created by `make dist`? Since it's not in the `EXTRA_DIST` list they won't be included there. Also github doesn't seem to include submodule contents in the source tarballs it generates.

> Also, it's GPL. So we should really keep it at arm's length. Better just install it as a normal build dependency.

Since we're not linking against it (only executing the built binary) I personally don't see much of an issue with it. 

To resolve concerns though, I now pushed a version where we get the tarball from github and compile the `uncrustify` binary in a temporary directory.

src/objects.c
JelteF: done

include/bouncer.h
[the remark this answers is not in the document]
JelteF: Removed for now. Once this initial one is merged I'll create a few PRs dedicated to some possible styling changes (this will be the first one).

src/main.c
JelteF: Well spotted. This was a weird thing. uncrustify is trying to align the comment on a tabstop turns out is right after the `;`.  I added a tab manually to give it some spacing. For some reason it wouldn't do it automatically. I think in a follow up PR I might propose changing to `indent_cmt_with_tabs = false`, which did add a single space between the semicolon and the comment. But for this PR that would definitely change too many places.

JelteF:
@petere I think I addressed all feedback.

src/main.c
petere: ok

.gitmodules
petere: We can just say, we use uncrustify version x.y.z, and everyone just has to get that one.

I hate git submodules and I don't want more git submodules. :)

pyproject.toml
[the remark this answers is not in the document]
petere: Yeah, I don't want uncrustify's stuff in pgbouncer's tree. We have enough crap in there, we don't need to import someone else's crap too and then keep track of that as versions change.

src/main.c
petere: seems ok after your change

uncrustify.cfg
petere: Ok, let's look at that later.

pyproject.toml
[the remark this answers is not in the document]
JelteF: Yeah fair enough, I had already updated the PR to have a make target to compile it in a temporary directory. But I forgot to remove these changes. I did that now. And I created a separate PR to update the excludes for the already existing submodules: https://github.com/pgbouncer/pgbouncer/pull/924

.flake8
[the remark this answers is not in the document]
petere: This might not be applicable anymore after it's no longer a submodule?

.gitmodules
JelteF: Did you see I updated the the PR to remove the submodule and have a make target to fetch and compile the uncrustify version from sources? Is that okay? Or would you rather I remove the install steps from the `Makefile` completely and instead move them to a `CONTRIBUTING.md` and have people run them manually?

petere (approved):
Ok by me.

Please check if the changes in `.flake8` and `pyproject.toml` are still required.

.flake8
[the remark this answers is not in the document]
JelteF: removed from here now now indeed. And I moved the still relevant changes to https://github.com/pgbouncer/pgbouncer/pull/924

petere:
> Please check if the changes in `.flake8` and `pyproject.toml` are still required.

It seems this was done. :)

JelteF:
> Ok by me.

Just to be clear, you're happy with the current method of installing uncrustify? (i.e. Makefile that fetches and compiles tarball in a temp directory, and puts the final binary in the root of the repo)

PR #893, opened 2023-07-17
Draft release 1.20.0

JelteF:


eulerto:
Should we keep the "Deprecations" section at the top?

JelteF:
> Should we keep the "Deprecations" section at the top?

It's definitly a preference thing. I usually prefer to put a Breaking Changes section at the top, because that's usually the main thing I want to find out when reading a changelog: is everything I use still working? Only after that I care about new features that I might want to start using, or fixed bugs that impact me.

So I while deprecations are not a breaking change yet. They'll probably be a breaking change in the future. So I think I'd prefer to keep it at the top.

eulerto:
Makes sense. I don't know when/if we will remove it but people will know (even if you rapidly check the first sentences) that its use is discouraged.

PR #902, opened 2023-07-24
Redesign shutdown logic to support rolling restarts using so_reuseport

### LINKED ISSUE #770: Deployment on Heroku: add options to handle SIGTERM

JelteF:
In 1.20 we announced deprecation of online restarts using the `-R` flag and recommended for people to use `so_reuseport` instead. But the safe shutdown logic that we executed when receiving a `SIGINT` signal didn't actually allow for zero-downtime rolling restarts.

This PR changes the behaviour of our shutdown signals to allow for rolling restarts. The changes that this PR makes are:

1. SIGTERM does not do "immediate shutdown" anymore, but instead triggers a newly introduced "super safe shutdown". This "super safe shutdown" waits for all clients to disconnect before shutting down the server. This can be used for rolling restarts as described in the newly added documentation. A second SIGTERM will cause an "immediate shutdown".
2. SIGINT mostly keeps doing what it was doing previously: waiting until all servers are released before shutting down PgBouncer. But it incorporates a few improvements from the new SIGTERM logic:
     1. It also stops listening for new connections while PgBouncer is waiting for servers to be released. This does have the downside that SIGUSR2 cannot be used to cancel a safe shutdown anymore, but that was quite a weird/niche feature anyway.
     2. Clients that have released their server and then do another query will be disconnected, instead of putting them in pause mode until either `query_wait_timeout` triggers or PgBouncer shuts down. This way they can reconnect quicker to another PgBouncer, because they were never going to get a new server anyway.
     3.  A that a second SIGINT will trigger an "immediate shutdown"
3. SIGQUIT handling is introduced and now causes an immediate shutdown.

Since this changes the existing behaviour of SIGTERM and SIGINT, this can be considered a breaking change. But the amount of breakage should be minimal since most of the time people would not want an immediate shutdown, and the new SIGINT behaviour is closer to the behaviour most people would expect.

Related to #894
Related to #900

Since this changes SIGTERM to not do a fast shutdown anymore this also indirectly addresses #806 and #770.

Fixes #806 
Fixes #770

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -760,6 +760,37 @@ Resume work from previous **KILL**, **PAUSE**, or **SUSPEND** command.
 
 The PgBouncer process will exit.
 
+#### SHUTDOWN WAIT_FOR_SERVERS
+
+Stop accepting new connections and shutdown after all servers are released.
+This is basically the same as issuing **PAUSE** and **SHUTDOWN**, except that
+this also stops accepting new connections while waiting for the **PAUSE** as
+well as eagerly disconnecting clients that are waiting to receive a server
+connection.
+
+#### SHUTDOWN WAIT_FOR_CLIENTS
+
+Stop accepting new connections and shutdown the process once all existing
+clients have disconnected. This command can be used to do zero-downtime rolling
+restart of two PgBouncer processes using the following procedure:
+
+1. Have two or more PgBouncer processes running on the same port using
+   `so_reuseport` ([configuring peering](/config.html#section-peers) is
+   recommended, but not required). To achieve zero downtime when
+   restarting we'll restart these processes one-by-one, thus leaving the
+   others running to accept connections while one is being restarted.
+2. Pick a process to restart first, let's call it A.
+3. Run `SHUTDOWN WAIT_FOR_CLIENTS` (or send `SIGTERM`) to process A.
+4. Cause all clients to reconnect. Possibly by waiting some time until the
+   client side pooler causes reconnects due to its `server_idle_timeout`
+   (or similar config). Or if no client side pooler is used, possibly by
+   restarting the clients. Once all clients have reconnected. Process A
+   will exit automatically, because no clients are connected to it anymore.
+5. Start process A again.
+6. Repeat step 3, 4 and 5 for each of the remaining processes, one-by-one
+   until you restarted all processes.
+
+
 #### RELOAD
 
 The PgBouncer process will reload its configuration files and update
@@ -800,10 +831,20 @@ passed to the PostgreSQL backend like any other SQL command.)
 SIGHUP
 :   Reload config. Same as issuing the command **RELOAD** on the console.
 
-SIGINT
-:   Safe shutdown. Same as issuing **PAUSE** and **SHUTDOWN** on the console.
-
 SIGTERM
+:   Super safe shutdown. Wait for all existing clients to disconnect, but don't
+    accept new connections. This is the same as issuing
+    **SHUTDOWN WAIT_FOR_CLIENTS** on the console. If this signal is received while
+    there is already a shutdown in progress, then an "immediate shutdown" is
+    triggered instead of a "super safe shutdown". In PgBouncer versions earlier
+    than 1.23.0, this signal would cause an "immediate shutdown".
+
+SIGINT
+:   Safe shutdown. Same as issuing **SHUTDOWN WAIT_FOR_SERVERS** on the console.
+    If this signal is received while there is already a shutdown in progress,
+    then an "immediate shutdown" is triggered instead of a "safe shutdown".
+
+SIGQUIT
 :   Immediate shutdown. Same as issuing **SHUTDOWN** on the console.
 
 SIGUSR1
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -90,9 +90,26 @@ enum PauseMode {
 };
 
 enum ShutDownMode {
+	/* just stay running */
 	SHUTDOWN_NONE = 0,
-	/* wait for all servers to become idle before stopping the process */
+	/*
+	 * Wait for all servers to become idle before stopping the process. New
+	 * client connection attempts are denied while waiting for the servers
+	 * to be released. Already connected clients that go to CL_WAITING
+	 * state are disconnected eagerly.
+	 */
 	SHUTDOWN_WAIT_FOR_SERVERS,
+	/*
+	 * Wait for all clients to disconnect before stopping the process.
+	 * While waiting for this we stop listening on the socket so no new
+	 * clients can connect. Still connected clients will continue to be
+	 * handed connections from the pool until they disconnect.
+	 *
+	 * This allows for a rolling restart in combination with so_reuseport.
+	 *
+	 * This is an even more graceful shutdown than SHUTDOWN_WAIT_FOR_SERVERS.
+	 */
+	SHUTDOWN_WAIT_FOR_CLIENTS,
 	/* close all connections immediately and stop the process */
 	SHUTDOWN_IMMEDIATE,
 };
--- a/src/admin.c
+++ b/src/admin.c
@@ -1123,8 +1123,15 @@ static bool admin_cmd_reload(PgSocket *admin, const char *arg)
 /* Command: SHUTDOWN */
 static bool admin_cmd_shutdown(PgSocket *admin, const char *arg)
 {
-	if (arg && *arg)
-		return syntax_error(admin);
+	enum ShutDownMode mode = SHUTDOWN_IMMEDIATE;
+	if (arg && *arg) {
+		if (strcasecmp(arg, "WAIT_FOR_CLIENTS") == 0)
+			mode = SHUTDOWN_WAIT_FOR_CLIENTS;
+		else if (strcasecmp(arg, "WAIT_FOR_SERVERS") == 0)
+			mode = SHUTDOWN_WAIT_FOR_SERVERS;
+		else
+			return syntax_error(admin);
+	}
 
 	if (!admin->admin_user)
 		return admin_error(admin, "admin access needed");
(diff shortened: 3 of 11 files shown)


src/main.c
```diff
@@ -481,8 +481,14 @@ static void handle_sigint(evutil_socket_t sock, short flags, void *arg)
 		die("takeover was in progress, going down immediately");
 	if (cf_pause_mode == P_SUSPEND)
 		die("suspend was in progress, going down immediately");
-	cf_pause_mode = P_PAUSE;
-	cf_shutdown = 1;
+	if (cf_so_reuseport) {
```
JelteF, line 484: Do we want to piggy-back on SIGINT? Or should we use a new signal for this behaviour.

doc/usage.md
```diff
@@ -771,7 +792,10 @@ SIGHUP
 :   Reload config. Same as issuing the command **RELOAD** on the console.
 
 SIGINT
-:   Safe shutdown. Same as issuing **PAUSE** and **SHUTDOWN** on the console.
+:   Safe shutdown. If `so_reuseport` is set to 0, then this is the same as
+    issuing **PAUSE** and **SHUTDOWN** on the console. If `so_reuseport` is set
+    to 1 then this is the same as issuing **SHUTDOWN WAIT_FOR_CLIENTS** on
+    the console.
```
petere, line 799: Hmm, we should use a separate signal for this. Otherwise, it's too much magic.

I think the best way would be:
- Keep SIGINT as is.
- Move the current SIGTERM behavior to SIGQUIT.
- This new behavior goes to SIGTERM.

It's a compatibility break, but it has some advantages. It restores the intended progression that SIGTERM is "softer" than SIGINT. And it creates some consistency with the postgres/pg_ctl shutdown modes.

We could even have a setting like signal_compat that keeps the old signal assignments. Not sure if it's needed, but it would be possible pretty easily.

include/bouncer.h
```diff
@@ -76,9 +76,28 @@ enum PauseMode {
...
 	SHUTDOWN_WAIT_FOR_SERVERS,
+	/*
+	 * Wait for all clients to disconnect before stopping the process. Stop
+	 * listening on the socket so no new clients can connect. Still
+	 * connected clients will continue to be handed connections from the
+	 * pool until they disconnect.
+	 *
+	 * This allows for a rolling restart in combination with so_reuseport.
+	 */
+	SHUTDOWN_WAIT_FOR_CLIENTS,
```
petere, line 100: I suppose "wait for clients" kind of includes "wait for servers"? Not sure if that is worth writing down, just thinking out loud.

src/pooler.c
```diff
@@ -57,6 +57,47 @@ static struct timeval err_timeout = {5, 0};
...
+	if (ls->fd > 0) {
+		safe_close(ls->fd);
+		ls->fd = 0;
+	}
+	if (pga_is_unix(&ls->addr) && cf_unix_socket_dir[0] != '@') {
+		char buf[sizeof(struct sockaddr_un) + 20];
+		snprintf(buf, sizeof(buf), "%s/.s.PGSQL.%d", cf_unix_socket_dir, cf_listen_port);
+		unlink(buf);
+	}
+	statlist_remove(&sock_list, &ls->node);
```
petere, line 74: This is a bit weird. Refactoring this function out from `cleanup_sockets()` makes sense, but then this should only work on a single socket in isolation, not manipulate a higher-layer list.

src/pooler.c
```diff
@@ -57,6 +57,47 @@ static struct timeval err_timeout = {5, 0};
...
+
+/*
+ * cleanup_tcp_sockets stops accepting new connections on the configured TCP
+ * sockets. This is used to support zero-downtime restarts with with
+ * so_reuseport=1. By not accepting new connections it's possible to
+ * effectively drain this process of all its connections by making sure the
+ * clients all reconnect (at their leisure). Then once they've all reconnected
+ * to another PgBouncer process, we can stop this one.
+ */
+void cleanup_tcp_sockets(void)
```
petere, line 86: Why only TCP sockets? I would expect all socket types to be treated equally here.

Also, do we need to do anything with libevent here, similar to cleanup_sockets? I guess if we close the socket, then no event will fire anyway, but this looks a bit odd, because it could potentially create different states where sometimes sockets are closed but still registered in libevent.

Maybe this function should just be cleanup_sockets?

src/main.c
petere: Prefer new signal, see above.

src/admin.c
```diff
@@ -1127,8 +1141,14 @@ static void full_resume(void)
 	if (tmp_mode == P_SUSPEND)
 		resume_all();
 
-	/* avoid surprise later if cf_shutdown stays set */
-	if (cf_shutdown) {
+	/*
+	 * Avoid surprises later if cf_shutdown stays set to
```
petere, line 1145: Maybe if we're rewriting this comment anyway, we can use a more precise language than "avoid surprises"? I for one don't know what surprise we are trying to avoid.

src/admin.c
```diff
@@ -1113,11 +1118,20 @@ static bool admin_cmd_shutdown(PgSocket *admin, const char *arg)
...
+		log_info("SHUTDOWN command issued");
+		event_base_loopbreak(pgb_event_base);
+		/*
+		 * By not running admin_ready the connection is kept open
+		 * until the process is actually shut down.
+		 */
+		return true;
+	} else {
+		log_info("SHUTDOWN WAIT_FOR_CLIENTS command issued");
+		cleanup_tcp_sockets();
```
petere, line 1132: Ah, so in the signal handler you are using `cleanup_sockets()`, but here you use `cleanup_tcp_sockets()`, because you're still connected. See my comments there, we should make this more coherent. I think the behavior we want here is to clean up all connections except *this one*, independent of the socket type.

petere (changes requested):
This all looks like a very good direction to me. We need work out how to handle the signals.

doc/usage.md
JelteF: While I understand the desire to be in line with how Postgres uses signals, I don't think breaking compatibility in such a big way is worth it. Changing SIGTERM would start making a lot of Docker and SystemD stop commands suddenly change their behaviour. I don't think that's something we should be doing.

The reason I suggested changing SIGINT, is because if `so_reuseport` because there are multiple processes listening is that the current SIGINT behaviour is actively harmful. Because the PgBouncer process is still accepting new connections only to then send an error on them, while one of the other listening procesess might still be accepting connections.

JelteF: I just checked:
- https://hub.docker.com/r/bitnami/pgbouncer/ uses SIGTERM (afaict)
- pgdg apt systemd service file uses SIGINT
- pgdg rpm systemd service file uses SIGINT

So changing behaviour of any of them seems pretty likely to cause problems in production for some users. I think assigning the new behaviour to SIGQUIT is the least disruptive.

It would be possible to add some config compat flag, to change the meaning of the signals to match postgres its intent. I think that's only likely to cause more confusion, because now you have to know what the config is to know how to stop pgbouncer in the way that you want.

petere: > I just checked:
> 
> * https://hub.docker.com/r/bitnami/pgbouncer/ uses SIGTERM (afaict)

I didn't find the source code for this, but if it's true, then I think it's possible that it's a mistake, possibly exactly because pgbouncer does SIGTERM and SIGINT backwards.

> * pgdg apt systemd service file uses SIGINT
> * pgdg rpm systemd service file uses SIGINT
> 
> So changing behaviour of any of them seems pretty likely to cause problems in production for some users.

My proposal would leave SIGINT unchanged, and almost everybody should be using SIGINT normally.

I'm not saying I'm 100% comfortable, but it doesn't seem too bad.

doc/usage.md
```diff
@@ -730,6 +730,27 @@ Resume work from previous **KILL**, **PAUSE**, or **SUSPEND** command.
...
+       client side pooler causes reconnects due to its `server_idle_timeout`
+       (or similar config). Or if no client side pooler is used, possibly by
+       restarting the clients. Once all clients have reconnected. Process A
+       will exit automatically, because no clients are connected to it anymore.
+    4. Start process A again
+    5. Repeat step 2, 3 and 4 for process B.
+
+If you have more than two processes, repeat the steps 2, 3 and 4 for each of
+them.
+
```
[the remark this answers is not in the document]
JelteF: I improved the description of the steps.

doc/usage.md
JelteF: Okay I implemented this proposal now, and if we put a warning in the release notes I think we should be fine.

The only issue I have is that Windows does not have SIGQUIT, so now you cannot do an immediate shutdown in windows anymore. Which maybe is fine.

src/admin.c
JelteF: Okay, so I implemented this. One thing I'm wondering is if we should do also close all sockets when issuing `SHUTDOWN WAIT_FOR_SERVERS` (or a SIGINT). It seems quite weird that we're allowing new connections, even though they will be held in waiting state until all the servers are done with their current task. It seems nicer to stop listening for new connections.

doc/usage.md
JelteF: Okay, I think I found (and implemented) a rather elegant solution to the windows problem: A second SIGTERM signal will trigger an immediate shutdown.

JelteF: Hmm, turns out windows doesn't actually send SIGTERM requests. So this was a non-issue. Still, it seems quite nice to have this "immediate" shutdown happen on repeated signals. I think I'd actually go so far as also wanting to implement this for SIGINT. Having double Ctrl+C do an immediate shutdown seems really nice for interactive usages.

include/bouncer.h
JelteF: > I suppose "wait for clients" kind of includes "wait for servers"?

This wasn't the case before, but with my most recent changes it does. I updated the comment to reflect this.

src/admin.c
JelteF: It was meant to avoid a future pause to cause the server to be shutdown when sending RESUME during a safe shutdown. I removed this code (and thus also the comment), because resuming from a shutdown is impossible now that we're closing the listening sockets at the start of the shutdown.

doc/usage.md
JelteF: I implement "double SIGINT" == "immediate shutdown"

murugiaha:
@JelteF any idea which version will this feature be available in?

eulerto:
@murugiaha In the next one. There is no release plans yet.

murugiaha:
> @murugiaha In the next one. There is no release plans yet.

Thanks for the quick response. Much appreciated.

JelteF:
My expectation for the next release would be somewhere this summer, but no promises.

PR #908, opened 2023-07-27
Support "options" in ignore_startup_parameters again

### LINKED ISSUE #907: Unsupported startup parameter in options: statement_timeout
Hi, 
I found that changes were performed about options parsing but did not appear in the changelog.
Is it normal behavior that now I should specify all parameters that I want to use in options in ignore_startup_parameters?

For example:
`
PGOPTIONS='-c idle_in_transaction_session_timeout=22' psql -p 6432 -h localhost postgres
PGOPTIONS='-c session_timeout=22' psql -p 6432 -h localhost postgres
`

In pgbouncer 1.19.1 it worked, in 1.20.0 it started to work after changing the config to look like this:
`
ignore_startup_parameters = extra_float_digits, options , statement_timeout, idle_in_transaction_session_timeout
`

JelteF:
By supporting parsing the `options` startup parameter, we accidentally
stopped supporting using it in `ignore_startup_parameters`.

Users have put `options` in `ignore_startup_parameters` and so upgrading
to 1.20.0 might break their existing connection strings, if the actual
options that are contained in the `options` startup parameter are
unsupported.

This changes the parsing of `options` to allow any unsupported parameter
if `options` itself is in `ignore_startup_parameters`.

NOTE: There's other failure paths in the `options` startup parameter
parsing that could still trigger now even though `options` is in
`ignore_startup_parameters`:

- too long parameter
- non `-c` parameter
- options would have been unparsable by PostgreSQL too

But honestly those seem unlikely enough that I think it's better to
error for them than to ignore those errors.

Fixes #907

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -217,6 +217,13 @@ Postgres extensions can change this list though, they can add parameters themsel
 and they can start reporting already existing paremeters that Postgres does not report.
 Notably Citus 12.0+ causes Postgres to also report `search_path`.
 
+The postgres protocol allows specifying parameters settings, both direcly as a
+parameter in the startup packet, or inside the [`options` startup
+packet][options-startup]. Parameters specified using both of these methods are
+supported by `track_extra_parameters`. However, it's not possible to include
+`options` itself in `track_extra_parameters`, only the parameters contained in
+`options`.
+
 Default: IntervalStyle
 
 ### ignore_startup_parameters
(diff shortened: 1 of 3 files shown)


emelsimsek:
Changes look good in restoring the compatibility with previous versions.

As a separate note, `ignore_startup_parameters` documentation is confusing to me. What does it mean that **ignored parameters are handled by admin**? I expect them not being cached but passed to postgres as they are. 

ignore_startup_parameters
By default, PgBouncer allows only parameters it can keep track of in startup packets: client_encoding, datestyle, timezone and standard_conforming_strings. All others parameters will raise an error. T**o allow others parameters, they can be specified here, so that PgBouncer knows that they are handled by the admin and it can ignore them.**

If you need to specify multiple values, use a comma-separated list (e.g. options,extra_float_digits)

emelsimsek (approved):

eulerto:
@JelteF The patch looks good to me. However, I have 2 suggestions:

* add a comment in the `set_startup_options` function describing the behavior of `options` value in `ignore_startup_parameters`.
* we should better document the `options` parameter. The commit dff974ff does not add any documentation. I'm not sure exactly where to add it. Maybe in the `ignore_startup_parameters` as a new paragraph.

JelteF:
Added the docs. Merging this now.

torstenfohrer:
Hi, i trying to get pgbouncer work with some software that  send following

image

And nothing set in ignore_startup_parameters=options,client_encoding,standard_conforming_strings will get around it.

Looking at code you try to find first -c option=value and then look if options is set in ignore_startup_parameters

PR #913, opened 2023-08-05
Add support for waiting clients count

franciscoreinolds:
Hello 👋🏽 

In this PR, I'm trying to add support to a new logging variable that would track how many clients have had to wait in order to connect to the server.

I feel like this makes sense when taking into account that there already exist `{query,xact}_{count,time}` / `wait_time` variables.

I tried looking into how these were calculated and follow the same approach for this one, but I'm of course open to feedback.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -249,6 +249,7 @@ int pga_cmp_addr(const PgAddr *a, const PgAddr *b);
 struct PgStats {
 	uint64_t xact_count;
 	uint64_t query_count;
+	uint64_t wait_count;	/*total amount of clients that had to wait*/
 	uint64_t server_bytes;
 	uint64_t client_bytes;
 	usec_t xact_time;	/* total transaction time in us */
--- a/src/objects.c
+++ b/src/objects.c
@@ -206,6 +206,8 @@ void change_client_state(PgSocket *client, SocketState newstate)
 		break;
 	case CL_WAITING:
 	case CL_WAITING_LOGIN:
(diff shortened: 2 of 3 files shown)


franciscoreinolds:
Hiya @JelteF 

I managed to fix most checks but there are 2 windows ones failing due to `connection_count` assertions failing. Since I didn't touch any code regarding that I find it weird, have you seen such behaviour before?

eulerto:
IMO this counter doesn't help. If I obtain 1,000 _wait_count_ that waits for 80ms and 50,000 _wait_count_ that waits for 1ms, it seems the later indicates a bad scenario but indeed the former is worse. This counter without the time information is useless. That's why there is **only** _wait_time_ available. Let's not add another counter just because other metrics already provide a counter.

franciscoreinolds:
Well, feel free to close then but I don't really understand the argument to be honest.

> If I obtain 1,000 wait_count that waits for 80ms and 50,000 wait_count that waits for 1ms, it seems the later indicates a bad scenario but indeed the former is worse. This counter without the time information is useless.

But we *do* have the time, so what's the problem?

> That's why there is **only** *wait_time* available.

What's the problem with giving people information about what's going on in their system? Just document what it does and let people decide what (if anything) they want to do with said information. Not giving people the information because one's afraid that they might misinterpret it seems like a strange concept to me.

Regardless, this is just my 2cents, if you feel like this adds no value to the project then feel free to close.

JelteF:
Closing this because #727 has now been merged

PR #921, opened 2023-08-14
[auth_dbname] introduce auto-registration option in case auto-database is configured

goksgie:
When the database section has an auto-database, '*', the expected behavior for login experience to have client's database registered when not found. However, this experience would be blocked if we configure `auth_dbname` to fall to auto-database. Hence, the main goal of this PR is to ensure that `auth_dbname` can still function on cases where it is not configured, but auto-database is provided.

To achieve that, this PR makes already existing function, find_or_register_database, public and uses it from various places to simplify code.

Edit:

Quoting the targeted use case of this change from @JelteF 's comment, which makes the change more presentable:

```ini
[databases]
* = host=localhost

[pgbouncer]
auth_dbname=postgres
```

Where, we will use `auth_dbname=postgres` with auto-registered database's connection string.

### DIFF
--- a/test/test_auth.py
+++ b/test/test_auth.py
@@ -77,9 +77,7 @@ def test_unregistered_auto_dbname_with_auto_database(bouncer):
         original = f.read()
     with bouncer.ini_path.open("w") as f:
         # uncomment the auto-database line
-        new = re.sub(
-            r"^;\* = ", "* = ", original, flags=re.MULTILINE
-        )
+        new = re.sub(r"^;\* = ", "* = ", original, flags=re.MULTILINE)
         print(new)
         f.write(new)
     # configure the auth_dbname to a database that is not configured
@@ -98,7 +96,8 @@ def test_unregistered_auto_dbname_with_auto_database(bouncer):
     # database. Hence, we can conclude that we were able to look up the password using
     # auth_dbname
     with pytest.raises(
-        psycopg.OperationalError, match="database \"this_database_doesnt_exist\" does not exist"
+        psycopg.OperationalError,
+        match='database "this_database_doesnt_exist" does not exist',
     ):
         bouncer.test(dbname="this_database_doesnt_exist", user="muser1", password="foo")
 
--- a/test/test_auth.py
+++ b/test/test_auth.py
@@ -72,9 +72,19 @@ def test_auth_dbname_with_auto_database(bouncer):
     bouncer.test(dbname="postgres", user="someuser", password="anypasswd")
 
 
-def test_unregistered_auto_dbname_with_auto_database(bouncer):
+def test_unconfigured_auth_database_with_auto_database(bouncer):
+    """
+    Tests the scenario where the authentication database does not
+    have a connection string configured under [databases] section.
+    However, there is an auto-datatabase, '*', configured. The expectation
+    is to use the wild card connection string for the auth_database.
+    """
     with bouncer.ini_path.open() as f:
         original = f.read()
+        assert (
+            re.search(r"^unconfigured_auth_database", original, flags=re.MULTILINE)
+            == None
+        )
     with bouncer.ini_path.open("w") as f:
         # uncomment the auto-database line
         new = re.sub(r"^;\* = ", "* = ", original, flags=re.MULTILINE)
@@ -82,7 +92,7 @@ def test_unregistered_auto_dbname_with_auto_database(bouncer):
         f.write(new)
     # configure the auth_dbname to a database that is not configured
     # expected behavior is to fallback to auto-database and auto-register.
-    bouncer.admin("set auth_dbname=postgres")
+    bouncer.admin("set auth_dbname=unconfigured_auth_database")
     bouncer.admin("reload")
     bouncer.admin("set auth_user='pswcheck'")
     bouncer.admin(f"set auth_type='md5'")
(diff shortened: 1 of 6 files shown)


eulerto:
We [discussed this topic](https://github.com/pgbouncer/pgbouncer/pull/764#discussion_r1039650643) when the `auth_dbname` feature was reviewed. This idea was rejected and removed from the patch.

I cannot imagine strong arguments to allow such option. You generally create some objects (table, view, function) to centralized the authentication into one specific database. This proposal is against this idea; indeed, it is similar to the pre-auth_dbname (create objects in multiple databases). What happen if you create a new database and forget to create the authentication objects?

I'm also concerned about security. If you have a credentials table in database A and in database B but you shouldn't use that table from database B. Depending on your setup, you might allow access to it.

goksgie:
> I cannot imagine strong arguments to allow such option. You generally create some objects (table, view, function) to centralized the authentication into one specific database. This proposal is against this idea; indeed, it is similar to the pre-auth_dbname (create objects in multiple databases). What happen if you create a new database and forget to create the authentication objects?
> 
> I'm also concerned about security. If you have a credentials table in database A and in database B but you shouldn't use that table from database B. Depending on your setup, you might allow access to it.

Users are still required to configure their authentication databases manually, whether it is through setting "global" or per database. This change lifts the requirement where auth_dbname needs to be an entry under `[databases]` section.

In terms of security, I do not think this PR poses any other risk than what already exists. Because:

1-> We enforce you to configure `auth_dbname`, whether globally or per database. This can only be performed from admin console, or through configuration file. If attacker has access to any of them, they can do anything with the old implementation as well.
2-> If the authentication database is the same as the client's target database, and it is not an entry under `[databases]`, then we auto_register before calling the `prepare_auth_database` as per client routine. Hence, the use case already partially supported.

JelteF:
> We [discussed this topic](https://github.com/pgbouncer/pgbouncer/pull/764#discussion_r1039650643) when the `auth_dbname` feature was reviewed. This idea was rejected and removed from the patch.

@eulerto So to be clear, this idea in this PR is quite different from the one discussed in that comment (afaict from my memory and re-reading the comment). The type of config that we decided to disallow there was:
```ini
[databases]
mydb = dbname=mydb

[pgbouncer]
auth_dbname=postgres
```
The initial version of #764 would then fall back to using `mydb` as the `auth_db`. I still think this should fail hard. Because falling back to using a different database indeed seems like a security issue

The type of config that this PR intends to support instead is:
```ini
[databases]
* = host=localhost

[pgbouncer]
auth_dbname=postgres
```

And I think it makes sense to allow this type of config. We'd still connect to the postgres database. But instead of requiring a dedicated connection string to connect to it, we use the wildcard connection string.

test/test_auth.py
```diff
@@ -72,6 +72,36 @@ def test_auth_dbname_with_auto_database(bouncer):
...
+        new = re.sub(r"^;\* = ", "* = ", original, flags=re.MULTILINE)
+        print(new)
+        f.write(new)
+    # configure the auth_dbname to a database that is not configured
+    # expected behavior is to fallback to auto-database and auto-register.
+    bouncer.admin("set auth_dbname=postgres")
+    bouncer.admin("reload")
+    bouncer.admin("set auth_user='pswcheck'")
+    bouncer.admin(f"set auth_type='md5'")
+    # postgres is not defined in test.ini
```
JelteF, line 89: Would be good to ensure that this is the case (e.g by checking the contents of the ini file). This seems like something we might accidentally change in the future.

goksgie: Shall we create a database for this purpose, something like `unconfigured_auth_database` to leave `postgres` somewhat reserved?

goksgie: added new database and improved the validation with [57e51cabc1f24ac1276743093eb086306b94a34e ](https://github.com/pgbouncer/pgbouncer/pull/921/commits/57e51cabc1f24ac1276743093eb086306b94a34e)

test/test_auth.py
```diff
@@ -72,6 +72,46 @@ def test_auth_dbname_with_auto_database(bouncer):
...
+        print(new)
+        f.write(new)
+    # configure the auth_dbname to a database that is not configured
+    # expected behavior is to fallback to auto-database and auto-register.
+    bouncer.admin("set auth_dbname=unconfigured_auth_database")
+    bouncer.admin("reload")
+    bouncer.admin("set auth_user='pswcheck'")
+    bouncer.admin(f"set auth_type='md5'")
+    # postgres is not defined in test.ini
+    bouncer.test(dbname="authdb", user="someuser", password="anypasswd")
```
JelteF, line 100: This comment is outdated. But I also think that this line doesn't really test anything more than the line below.

test/test_auth.py
```diff
@@ -54,22 +54,42 @@ def test_auth_dbname_disabled(bouncer):
         bouncer.test(dbname="pauthz", user="someuser", password="anypasswd")
 
 
-def test_auth_dbname_with_auto_database(bouncer):
```
JelteF, line 57: I think this previously existing test was testing something slightly differently. The auto-db with an auth_dbname on the same line. So I think it makes sense to keep it.

eulerto:
> And I think it makes sense to allow this type of config. We'd still connect to the postgres database. But instead of requiring a dedicated connection string to connect to it, we can generate the connection string using the wildcard connection string.

I misunderstood the proposal because the description wasn't clear enough. I'm fine with this proposal.

JelteF (approved):

PR #926, opened 2023-08-16
refactor: Use an enum for shutdown values

JelteF:
In #902 I want to introduce a 4th shutdown value. So I started using an
enum for this to improve readibility. I think this refactor from using
integer values to enum values stands on its own. So to make reviewing
easier, and not have it be blocked on adding tests and some discussions
for #902 this PR includes only that refactor.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -77,8 +77,10 @@ enum PauseMode {
 
 enum ShutDownMode {
 	SHUTDOWN_NONE = 0,
-	SHUTDOWN_WAIT_FOR_SERVERS = 1,
-	SHUTDOWN_IMMEDIATE = 2,
+	/* wait for all servers to become idle before stopping the process */
+	SHUTDOWN_WAIT_FOR_SERVERS,
+	/* close all connections immediately and stop the process */
+	SHUTDOWN_IMMEDIATE,
 };
 
 enum SSLMode {
--- a/include/pooler.h
+++ b/include/pooler.h
@@ -22,7 +22,6 @@ void resume_pooler(void);
 void suspend_pooler(void);
 void per_loop_pooler_maint(void);
 void pooler_tune_accept(bool on);
-void cleanup_tcp_sockets(void);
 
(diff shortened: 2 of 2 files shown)


include/pooler.h
```diff
@@ -22,6 +22,7 @@ void resume_pooler(void);
 void suspend_pooler(void);
 void per_loop_pooler_maint(void);
 void pooler_tune_accept(bool on);
+void cleanup_tcp_sockets(void);
```
eulerto, line 25: It seems unrelated to this change.

include/bouncer.h
```diff
@@ -75,6 +75,12 @@ enum PauseMode {
 	P_SUSPEND = 2		/* wait for buffers to be empty */
 };
 
+enum ShutDownMode {
+	SHUTDOWN_NONE = 0,
+	SHUTDOWN_WAIT_FOR_SERVERS = 1,
+	SHUTDOWN_IMMEDIATE = 2,
+};
```
eulerto, line 82: Do you need the numbers?

eulerto (approved):

include/bouncer.h
JelteF: The 0 one is necessary, so we can do `if (cf_shutdown) { ... }`. So I left that one and removed the others, because those were indeed not needed.

include/pooler.h
JelteF: Thanks for noticing, that one should be part of #902 only.

include/bouncer.h
eulerto: There is no harm specifying the `= 0` but the spec says it starts with zero if not specified [1]. Since you mention this `if`:

```
if (cf_shutdown != SHUTDOWN_NONE) {
...
```

is my preference.

[1] https://en.cppreference.com/w/c/language/enum

JelteF: >There is no harm specifying the = 0 but the spec says it starts with zero if not specified [1]

Hmm, interesting I didn't know that. I got in the habit of at least providing the 0 one because @petere suggested that in a few PRs.

PR #938, opened 2023-08-29
Remove udns

eulerto:
libudns is not maintained for a long time (+9 years) [1]. Since there are other DNS resolvers that have been maintained and there isn't a modern operating system that enables it, remove libudns support.

[1] https://www.corpit.ru/mjt/udns.html

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -16,7 +16,7 @@ task:
     - container:
         image: ubuntu:20.04
       env:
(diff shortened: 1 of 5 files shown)


JelteF (approved):
Looks good. There's clearly no reason to use it. It's strictly worse than c-ares, even c-ares <=1.10, since udns doesn't support IPv6 at all.

PR #845, opened 2023-05-14
Support of prepared statements

knizhnik:
This PR is actually successor of #757 which seems to be stuck.
It contains the following fixes/improvements:
1. Fix bug transformation of bind request
2. Fix bug in calculating size of prepare packet
3. Prevent possible collisions in prepared statements mapping caused by using query hash
4. Replace prepared statement eviction algorithm (use LRU instead of sorting by bind_count)
5. Reject prepare command with parameter types.
6. Rewrite cache of prepared statement to avoid occasional colliions

Co-Author: @dashorst

PR #939, opened 2023-08-30
Stop supporting Ubuntu 18.04, since it is EOL

JelteF:
Ubuntu 18.04 has been causing build issues in PR #845, because we
started using a unicode symbol in the docs. This fails the docs build.
Since Ubuntu 18.04 has been EOL since April, it seems better to drop
support than to add code to work around this.

PG9.6 is also EOL, but continuing to support it hasn't been a big issue
yet. Also it seems useful for PgBouncer to still be able to talk to EOL
Postgres instances, but running the newest version of PgBouncer itself
on EOL operating systems seems much less useful.

We still remove the whole matrix task though, because there's already
another task with PG9.6.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -49,10 +49,6 @@ task:
     - container:
         image: ubuntu:22.04
     - container:
-        image: ubuntu:18.04
-      env:
-        PGVERSION: 9.6
-    - container:
         image: debian:stable
       env:
         PGVERSION: 13


eulerto:
LGTM. However, I propose that we remove this combination at all because there is another instance that is covering Postgres 9.6.

eulerto (approved):

PR #943, opened 2023-09-01
Fix errors pointed out by static analyzer

Medvecrab:
A static analyzer found some errors -

1) Deref of Null - admin.c
>Return value of a function 'pktbuf_dynamic' is dereferenced at admin.c:245 without checking for NULL, but it is usually checked for this function (25/26).

2) Memory Leak - proto.c
>Dynamic memory, referenced by 'ibuf', is allocated at proto.c:474 by calling function 'malloc' and lost at proto.c:488.

3) Memory Leak - scram.c
>Dynamic memory, referenced by 'server_signature_base64', is allocated at scram.c:932 by calling function 'malloc' and lost at scram.c:938.

4) No VA End - proto.c : 686; 
>Function 'va_end' was not called before proto.c:686 inside function 'scan_text_result'. (Function 'va_start' was called at proto.c:624.)

5) No VA End - proto.c : 693
>Function 'va_end' was not called before proto.c:693 inside function 'scan_text_result'. (Function 'va_start' was called at proto.c:624.)

So here's proposed fixes

### DIFF
--- a/src/proto.c
+++ b/src/proto.c
@@ -485,7 +485,7 @@ static bool login_scram_sha_256_final(PgSocket *server, unsigned datalen, const 
 	if (!verify_server_signature(&server->scram_state, user, ServerSignature)) {
 		slog_error(server, "invalid server signature");
 		kill_pool_logins(server->pool, NULL, "server login failed: invalid server signature");
-		return false;
+		goto failed;
 	}
 
 	free(ibuf);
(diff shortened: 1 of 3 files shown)


src/proto.c
```diff
@@ -683,14 +683,17 @@ int scan_text_result(struct MBuf *pkt, const char *tupdesc, ...)
 				int newlen;
 				if (strncmp(val, "\\x", 2) != 0) {
 					log_warning("invalid bytea value");
+					va_end(ap);
```
JelteF, line 686: To avoid errors like this in the future, let's use the `goto failed` pattern for this function.

JelteF (approved):
Thank you, looks good overall. But left one small suggestion for an improvement.

Medvecrab:
Added suggested change, thanks for review!

JelteF (approved):

Medvecrab:
Force-pushed a formatting fix

PR #947, opened 2023-09-15
Maintain `min_pool_size` for pools with forced user (even if no client conns) 

fschmager:
Continuation/Restart of PR #437 (shoutout to @barqshasbite) to address issue #426.

The default behaviour of PGBouncer requires that clients be connected to a pool before it is initialized. There is no way to toggle this behaviour to ensure that a pool is initialized immediately by PGBouncer.

This means that `min_pool_size` will only be respected if clients are connected. If all clients disconnect, then the pool can drop below the `min_pool_size` setting to `0` (`server_lifetime` will eventually close all connections). This can cause performance problems when a client finally reconnects since all of the pool connections have to be recreated.

This PR changes things so that PgBouncer (creates and) maintains pools that have user connections and, this is new, for those pools that have force users. 

This is changing behavior as `min_pool_size` will now be enforced even if there are no client connection.


src/janitor.c
```diff
@@ -680,6 +680,25 @@ static void do_full_maint(evutil_socket_t sock, short flags, void *arg)
...
+	 * there are many users. Are all database/user combos even known ahead of time?
+	*/
+	statlist_for_each_safe(item, &database_list, tmp) {
+		db = container_of(item, PgDatabase, head);
+		if (database_min_pool_size_without_clients(db) &&
+			database_min_pool_size(db) > 0 &&
+			db->forced_user != NULL) {
+			get_pool(db, db->forced_user);
+		}
+	}
```
fschmager, line 700: I believe this is the crucial part. Creating pools to maintain the `min_pool_size`. Pools needs a database and a user. If we don't have a user we can't create a pool.
Are all user/database combinations known or are there ways to configure users externally? Do we want that? Is doing this just for forced users sufficient and clear (unsurprising) enough?

For my use case, enabling this for forced users is sufficient.

src/server.c
```diff
@@ -210,12 +210,35 @@ int pool_pool_size(PgPool *pool)
...
+	return database_min_pool_size_without_clients(pool->db);
+}
+
+/* returns true if the db requires clients to maintain min_pool_size or if it's set globally, false otherwise */
+bool database_min_pool_size_without_clients(PgDatabase *db)
+{
+	if (db->min_pool_size_requires_clients == 0)
+		return true;
+	else
+		return cf_min_pool_size_requires_clients == 0;
```
fschmager, line 241: here I'm reversing the logic: if the setting is `1` this method returns `true`, and `0` -> `true`.

This made me think of an alternative name for the setting. Should we call it `min_pool_size_without_clients`? The current `min_pool_size_requires_clients` has a nice ring because it implies some kind of behaviour.

test/test.ini
```diff
@@ -23,6 +23,9 @@ p62= port=6666 host=127.0.0.1 dbname=p6
 p7a= port=6666 host=127.0.0.1 dbname=p7
 p7b= port=6666 host=127.0.0.1 dbname=p7
 p7c= port=6666 host=127.0.0.1 dbname=p7
+p7x= port=6666 host=127.0.0.1 dbname=p7 min_pool_size=3 min_pool_size_requires_clients=0
+p7y= port=6666 host=127.0.0.1 dbname=p7 min_pool_size=3 min_pool_size_requires_clients=1 user=marko
+p7z= port=6666 host=127.0.0.1 dbname=p7 min_pool_size=3 min_pool_size_requires_clients=0 user=pswcheck
```
fschmager, line 28: The test relying on these works, but sadly this breaks tests that rely on the number of log `"new connection to server"` events

* `test/test_limits.py::test_min_pool_size_with_lower_max_db_connections` ([code](https://github.com/pgbouncer/pgbouncer/blob/bb018a6ac765b41d8d1c1129fe8c6ccc56018ec2/test/test_limits.py#L66-L68))
* `test/test_limits.py::test_min_pool_size_with_lower_max_user_connections` ([code](https://github.com/pgbouncer/pgbouncer/blob/bb018a6ac765b41d8d1c1129fe8c6ccc56018ec2/test/test_limits.py#L79-L81))
* `test/test_misc.py::test_host_list_dummy` ([code](https://github.com/pgbouncer/pgbouncer/blob/master/test/test_misc.py#L133-L134))

Suggestions on how to handle that?

Is there a way to define pool settings ad-hoc in a test (without having them in `test/test.ini`? That would avoid polluting the other tests.
Rewrite the other tests to be more specific about the connections? The log messages to contain the pool name and user name (e.g. `LOG S-0x562b6ab0c0a0: p0x/maxedout2@127.0.0.1:10211 new connection to server (from 127.0.0.1:58340)`)
Thoughts?

JelteF: there is the `run_with_config` context manager which allows you to specify a full config. Another option is to use the config line uncommenting trick that `test_auth_dbname_with_auto_database` uses.

src/janitor.c
JelteF: > I believe this is the crucial part. Creating pools to maintain the min_pool_size. Pools needs a database and a user. If we don't have a user we can't create a pool.

I agree.

> Are all user/database combinations known or are there ways to configure users externally? 

Using auth_query, users can be dynamically loaded. So no, not all combinations are known.

> Do we want that? Is doing this just for forced users sufficient and clear (unsurprising) enough?

I think preferably we'd have a way to specify what pools (i.e. database+user combinations) should always be maintained. Something like this (suggests for better config name welcome):
```ini
mydb = min_pool_size=3 ensure_pools_exist_for=postgres,jelte,fschmager
```

We could even have that option default to the forced user if the forced user is specified. Which I think would make sense as a default. So this would already automatically create the pool (and thus maintain the min_pool_size):
```ini
mydb = min_pool_size=4 user=fschmager
```

fschmager: I was being conservative to maintain back-compat. The "maintain pools for forced users"-feature should be an opt-in as not to change existing behavior.

A blunt global setting to opt in might just do the trick (name up for debate):
```ini
create_pools_for_forced_users = 1
```

I don't know enough about how the community is using pgbounce. I just know my own use case, and for me "global opt-in + forced users == create & maintain pool" is exactly what I need. Feels like the most straightforward and least surprising path.

If we want to make it more complex:
* Do we need an opt-out on a per-pool basis? 
* Do we need a way to maintain pools for non-forced users?

fschmager: > if you do set min_pool_size I think the only reasonable thing is to actually maintain that min_pool_size from startup

I agree, this was _exactly_ my assumption. I think we can improve the situation with documentation of `min_pool_size`. Adding the caveat that this only works for forced users, especially on the global `min_pool_size` end. The term "forced users" might not be widely know and might need a bit of word smithing.

I'd be more than happy to drop the setting flags.

> There might be some PgBouncer timeouts in play that we should ignore in that case, to actually retry opening a connection.

Let me think out loud to explain what I understand. 

no "forced pool" (for want of a better word) scenario:
PgBouncer starts up, client connects, but postgres isn't available yet. Connection timeout might kick in, error, no pool created, no pool needs to be maintained.

"forced pool" scenario:
PgBouncer starts up and tries to create backend connection for "forced pools". It will do that every 1/3sec. If postgres isn't available.
The operation fails and will get retried 1/3sec later.
Is what you are concerned about the connection-creation timeout being higher than 1/3sec and connection-creation attempts will pile up? Or is pool maintenance single threaded and waiting for a connection will just delay the next run of pool-maintenance? Or is there a problem around clients trying to get a backend connection while the maintenance is trying to create such a connection?

While we work through that to see if anything needs to change around timetouts, I'm going ahead with simplifying the code by removing the additional settings and adjusting the docs.

mcampo-dd (approved):

JelteF:
Generally it slows tests down more, because it's running with valgrind (and without optimizations).

JelteF:
Release cadance is irregular, but roughly every 2-4 months.

JelteF:
I think it's because with the config used in the `min_pool_size` test we have many forced user connections. By setting `min_pool_size` globally, it starts opening connections for all of those. I think it would be best to update the test, and not set `min_pool_size` globally at all, but only on a specific database. Just like you're doing for your new test.

test/test_limits.py
```diff
@@ -37,9 +38,32 @@ async def test_pool_size(pg, bouncer):
 
 @pytest.mark.asyncio
 async def test_min_pool_size(pg, bouncer):
-    bouncer.admin("set min_pool_size = 3")
```
fschmager, line 40: The test has been changed to not set the global `min_pool_size` anymore. Caused too many connection creations in a number of pools overloading the github test runner. 
Rather this test check various combinations with individual pools.

doc/config.md
```diff
@@ -137,6 +137,11 @@ Add more server connections to pool if below this number.
 Improves behavior when the normal load suddenly comes back after a period
 of total inactivity.  The value is effectively capped at the pool size.
 
+Only enforced for pools where at least one of the following is true:
+
+* the pool has a `user` set in the `[database]` section (aka forced user)
+* there are clients connected to the pool
```
JelteF, line 143:
```suggestion
* the entry in the `[database]` section for the pool its database has a value
  set for the `user` key (aka forced user)
* there is at least one client connected to the pool
```

doc/config.md
```diff
@@ -1080,6 +1085,11 @@ the `default_pool_size` is used.
 Set the minimum pool size for this database. If not set, the global `min_pool_size` is
 used.
 
+Only enforced if at least one of the following is true:
+
+* the pool has a `user` set (aka forced user)
+* there are clients connected to the pool
```
JelteF, line 1091:
```suggestion
* this entry in the `[database]` section has a value set for the `user` key
  (aka forced user)
* there is at least one client connected to the pool
```

src/janitor.c
```diff
@@ -680,6 +680,22 @@ static void do_full_maint(evutil_socket_t sock, short flags, void *arg)
...
 		return;
 
+	/*
+	 * Creating new pools to enable `min_pool_size` enforcement even if there are no active clients.
+	 *
+	 * If clients never connect there won't be a pool to maintain the min_pool_size on, which means we have to
+	 * proactively create a pool, so that it can be maintained.
+	 *
+	 * We are doing this for databases with forced users only, to reduce the risk of creating connections in
+	 * unexpected ways, where there are many users.
```
JelteF, line 690: wrap this comment at 80 characters:
```suggestion
	 * Creating new pools to enable `min_pool_size` enforcement even if
	 * there are no active clients.
	 *
	 * If clients never connect there won't be a pool to maintain the
	 * min_pool_size on, which means we have to proactively create a pool,
	 * so that it can be maintained.
	 *
	 * We are doing this for databases with forced users only, to reduce
	 * the risk of creating connections in unexpected ways, where there are
	 * many users.
```

test/test_limits.py
```diff
@@ -37,9 +38,32 @@ async def test_pool_size(pg, bouncer):
...
+    bouncer.admin("reload")
+
+    # having to wait a little to give janitor time to create connection to satisfy min_pool_size
+    await asyncio.sleep(2)
 
+    # ensure db without min_pool_size has no connections
+    # p0
+    assert pg.connection_count(dbname="p0", users=("bouncer",)) == 0
+    # ensure db with min_pool_size and forced user has required connections (the uncommented one)
+    # p0z
```
JelteF, line 60:
```suggestion
    # ensure db with min_pool_size and forced user (p0z) has the required
    # backend connections
```

test/test_limits.py
```diff
@@ -37,9 +38,32 @@ async def test_pool_size(pg, bouncer):
...
+    assert pg.connection_count(dbname="p0", users=("pswcheck",)) == 3
+
+    # ensure db with min_pool_size and no forced user has no backend connections
+    # p0x
```
JelteF, line 64:
```suggestion
    # ensure db with min_pool_size and no forced user (p0x) has no backend
    # connections
```

test/test_limits.py
```diff
@@ -37,9 +38,32 @@ async def test_pool_size(pg, bouncer):
...
+    assert pg.connection_count(dbname="p0", users=("postgres",)) == 0
+    # client connecting to p0x should trigger backend connection creation up to min_pool_size
     # It's a bit tricky to get the timing of this test to work
```
JelteF, line 67:
```suggestion
    # client connecting to p0x should trigger backend connection creation up to
    # min_pool_size.
    #
    # NOTE: It's a bit tricky to get the timing of this test to work
```

JelteF (approved):
Looks good to me. I left some final suggestions on comment/docs improvements.

PR #948, opened 2023-09-21
Improve security of OpenSSL settings

JelteF:
The default OpenSSL ciphers were extremely outdated. This PR brings them
up to date. The main code changes for this are in libusual (which this
PR updates):
https://github.com/libusual/libusual/pull/40

The changes in PgBouncer itself are mostly doc changes. The only code
change is changing the default of the cipher to actually be "default",
which now means the same as "fast" anyway.


eulerto (approved):
You forgot to change `etc/pgbouncer.ini`. Other than that, LGTM.

PR #949, opened 2023-09-22
PgBouncer md5 FIPS compatibility changes

ajit-at-edb:
Root cause:

MD5 encryption functions are internally implemented in libusual (./lib/usual/crypto/md5.c). MD5 Authentication functions can not be used on FIPS compliant host. Due to internal implementation of MD5 algorithm, MD5 authentication is working without error on FIPS enabled host. Ideally it should throw error.

 Solution:

Due to internal implementation, we are not using openssl APIs for MD5. In this fix all internal MD5 calls are replaced to openssl APIs after which MD5 authentication throws expected error on FIPS compliant host.

Testing:

Pytest  testsuite ran successfully.
MD5 authentication scenario on non-FIPS compliant host runs successfully with and without compiling with “--without-openssl ” flag.
MD5 authentication scenario on FIPS compliant host errors out as expected when compiled with openssl flag.
MD5 authentication scenario on FIPS compliant host runs successfully when compiled with " “--without-openssl ” flag.

### DIFF
--- a/include/util.h
+++ b/include/util.h
@@ -41,7 +41,7 @@ int log_socket_prefix(enum LogLevel lev, void *ctx, char *dst, unsigned int dstl
  * password tools
  */
 #define MD5_PASSWD_LEN  35
-bool pg_md5_encrypt(const char *part1, const char *part2, size_t p2len, char *dest);
+bool pg_md5_encrypt(const char *part1, const char *part2, size_t p2len, char *dest) _MUSTCHECK;
 void get_random_bytes(uint8_t *dest, int len);
 
 const char *bin2hex(const uint8_t *src, unsigned srclen, char *dst, unsigned dstlen);
--- a/include/util.h
+++ b/include/util.h
@@ -41,7 +41,7 @@ int log_socket_prefix(enum LogLevel lev, void *ctx, char *dst, unsigned int dstl
  * password tools
  */
 #define MD5_PASSWD_LEN  35
-void pg_md5_encrypt(const char *part1, const char *part2, size_t p2len, char *dest);
+bool pg_md5_encrypt(const char *part1, const char *part2, size_t p2len, char *dest);
 void get_random_bytes(uint8_t *dest, int len);
 
 const char *bin2hex(const uint8_t *src, unsigned srclen, char *dst, unsigned dstlen);
--- a/src/util.c
+++ b/src/util.c
@@ -148,9 +148,8 @@ failed:
 
 	memcpy(dest, "md5", 3);
 	hash2hex(hash, dest + 3);
-	result = true;
+	return true;
 #endif
-	return result;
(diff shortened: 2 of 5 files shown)


src/util.c
```diff
@@ -104,6 +109,32 @@ void pg_md5_encrypt(const char *part1,
...
+	if (EVP_DigestInit(&mdctx, EVP_md5()) <= 0)
+	{
+		log_error("MD5 authentication: %s", ERR_reason_error_string(ERR_get_error()));
+	}
+
+	else if (EVP_DigestUpdate(&mdctx, part1, strlen(part1)) <= 0 ||
+			EVP_DigestUpdate(&mdctx, part2, part2len) <= 0 ||
+			EVP_DigestFinal_ex(&mdctx, hash, 0) <= 0 )
+	{
+		log_error("MD5 authentication: %s", ERR_reason_error_string(ERR_get_error()));
```
JelteF, line 127: Now that we can have errors when "encrypting" with md5, we should make pg_md5_encrypt return true/false accordingly. So we can correctly handle an error. We wouldn't want an error to result in comparing to an empty hash or something.

JelteF (changes requested):
Compilation is failing on CI for many targets

include/util.h
```diff
@@ -41,7 +41,7 @@ int log_socket_prefix(enum LogLevel lev, void *ctx, char *dst, unsigned int dstl
  * password tools
  */
 #define MD5_PASSWD_LEN  35
-void pg_md5_encrypt(const char *part1, const char *part2, size_t p2len, char *dest);
+bool pg_md5_encrypt(const char *part1, const char *part2, size_t p2len, char *dest);
```
JelteF, line 44:
```suggestion
bool pg_md5_encrypt(const char *part1, const char *part2, size_t p2len, char *dest) _MUSTCHECK;
```

src/util.c
```diff
@@ -100,10 +105,34 @@ static void hash2hex(const uint8_t *hash, char *dst)
...
 		    const char *part2, size_t part2len,
 		    char *dest)
 {
+	bool result = false;
+#ifdef USUAL_LIBSSL_FOR_TLS
+	EVP_MD_CTX *mdctx;
+	uint8_t hash[MD5_DIGEST_LENGTH];
+
+	ERR_clear_error();
+	mdctx = EVP_MD_CTX_create();
```
JelteF, line 118: It might be nice to do this allocation only once for the lifetime of the program, instead of for every call to pg_md5_encrypt.

JelteF (changes requested):

src/util.c
JelteF: To be clear, I don't think this is strictly necessary for this PR.

JelteF (approved):

PR #972, opened 2023-10-27
Handle DEALLOCATE ALL and DISCARD ALL for prepared statements

### LINKED ISSUE #974: `DEALLOCATE ALL` not being handled with new support for named prepared statements leads to problems
Thanks for adding support for named prepared statements in 1.21.0, that was a long-time pain point for Clients that use named prepared statements by default (like Prisma Client) and made weird workarounds necessary in the past 🥳 

Prisma Client, for example, added a `DEALLOCATE ALL` statement before all its named prepared statements in a transaction to make sure the connection is free of them, so that any new named prepared statement that might have been used by another Client before, would not lead to an error (`prepared statement "s0" already exists`). Not a great solution, but at least it made it work ™️.

Unfortunately, with 1.21.0 and the new support for named prepared statements, this now leads to its own new problem, which is described here: https://github.com/prisma/prisma/issues/21635 Essentially, the `DEALLOCATE ALL` does its job on the underlying connection and removes all existing named prepared statements, but PgBouncer does not understand that and thinks _its own prepared statement_ with the `PGBOUNCER_x` name is still present, and tries to use it, leading to an error like `prepared statement "PGBOUNCER_3" does not exist`.

This unfortunately will make all Prisma Client users that have been using our workaround with versions before <1.21.0 have to change their setup (which in Prisma's case is via the connection string) not to explode when the underlying PgBouncer is upgrade to 1.21.0.

Optimally, PgBouncer would handle `DEALLOCATE ALL` and similar statements and also remove its internal "cache" of prepared statements for the connection. https://github.com/pgbouncer/pgbouncer/pull/972 does something in that direction if I understand correctly.

knizhnik:
This implements a hacky/clever approach of handling the `DEALLOCATE ALL` and `DISCARD ALL` SQL statements. Instead of parsing the queries themselves, which PgBouncer avoids doing completely, this parses the CommandComplete packet returned by the server.

Fixes https://github.com/prisma/prisma/issues/21635
Fixes #974

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -365,6 +365,20 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 			if (client)
 				client->expect_rfq_count = 0;
 		}
+		if (is_prepared_statements_enabled(server->pool)
+		    && (pkt->len == 1 + 4 + 15 || pkt->len == 1 + 4 + 12)) {	/* size of complete DEALLOCATE/DISCARD ALL */
+			const char *tag;
+			if (mbuf_get_string(&pkt->data, &tag)) {
+				if (strcmp(tag, "DEALLOCATE ALL") == 0 ||
+				    strcmp(tag, "DISCARD ALL") == 0) {
+					free_server_prepared_statements(server);
+					if (client)
+						free_client_prepared_statements(client);
+				}
+			} else {
+				return false;
+			}
+		}
 		pop_outstanding_request(server, "E", &ignore_packet);
 
 		break;
--- a/src/server.c
+++ b/src/server.c
@@ -365,6 +365,13 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 			if (client)
 				client->expect_rfq_count = 0;
 		}
+		/*
+		 * Clean up prepared statements if needed if the client sent a
+		 * DEALLOCATE ALL or a DISCARD ALL query. Not doing so would
+		 * confuse our prepared statement handling, because we would
+		 * expect certain queries to be prepared at the server that are
+		 * not.
+		 */
 		if (is_prepared_statements_enabled(server->pool)
 		    && (pkt->len == 1 + 4 + 15 || pkt->len == 1 + 4 + 12)) {	/* size of complete DEALLOCATE/DISCARD ALL */
 			const char *tag;
--- a/doc/config.md
+++ b/doc/config.md
@@ -326,12 +326,11 @@ prepared on the server yet, it automatically prepares that statement before
 forwarding the command that the client sent.
 
 Note: This tracking and rewriting of prepared statement commands does not work
-for SQL-level prepared statement commands such as `PREPARE`, `EXECUTE`,
-`DEALLOCATE`, `DEALLOCATE ALL` and `DISCARD ALL`. Running `DEALLOCATE ALL` and
-`DISCARD ALL` is especially problematic, since those commands appear to run
-successfully, but they mess up with the state of the server connection
-significantly without PgBouncer noticing. Which in turn will very likely break
-the execution of any further prepared statements on that server connection.
+for SQL-level prepared statement commands, so `PREPARE`, `EXECUTE` and
+`DEALLOCATE` are forwarded straight to Postgres. The exception to this rule are
+the `DEALLOCATE ALL` and `DISCARD ALL` commands, these do work as expected and
+will clear the prepared statements that PgBouncer tracked for the client that
+sends this command.
 
 The actual value of this setting controls the number of prepared statements
 kept active on a single server connection. When the setting is set to 0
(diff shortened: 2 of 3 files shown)


JelteF (changes requested):
I think supporting DEALLOCATE ALL is a worthy effort. It currently silently breaks the whole server. And because there is no protocol level equivalent, all clients that want send this need to send it as SQL.

When I saw the title of this PR I thought I was going to hate it, because I'd expect it to implement a very simple query parser. And I was happily surprised that's not what's going on. Instead it's using the command tag from the CommandComplete packet. And I actually think that idea is quite elegant, since the command tags are well defined by postgres and easy to parse (as opposed to SQL which can have comments or extra spaces or lowercase). So I quite like this general approach. I left some comments on changes that should be done to make this handle some more edge cases.

Also this needs some tests. And docs need to be updated that we now actually do support DEALLOCATE ALL

janpio:
I have just opened https://github.com/pgbouncer/pgbouncer/issues/974, which explains the problem and thus motivation behind such a change a bit more. Great to see there is already a PR aiming to fix this problem! 💚

src/server.c
```diff
@@ -365,6 +365,14 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 			if (client)
 				client->expect_rfq_count = 0;
 		}
+		if (is_prepared_statements_enabled(server->pool)) {
+			const char* tag;
+			if (mbuf_get_string(&pkt->data, &tag)) {
```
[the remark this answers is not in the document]
knizhnik: Done

src/server.c
```diff
@@ -365,6 +365,14 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
...
+				if (strcmp(tag, "DEALLOCATE ALL") == 0)	{
+					free_server_prepared_statements(server);
```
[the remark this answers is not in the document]
knizhnik: Done

src/server.c
```diff
@@ -365,6 +365,14 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 			if (client)
 				client->expect_rfq_count = 0;
 		}
+		if (is_prepared_statements_enabled(server->pool)) {
+			const char* tag;
+			if (mbuf_get_string(&pkt->data, &tag)) {
+				if (strcmp(tag, "DEALLOCATE ALL") == 0)	{
```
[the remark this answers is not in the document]
knizhnik: Done

src/server.c
```diff
@@ -365,6 +365,22 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
 			if (client)
 				client->expect_rfq_count = 0;
 		}
+		if (is_prepared_statements_enabled(server->pool)
+			&& (pkt->len == 1+4+15 || pkt->len == 1+4+12)) /* size of complete DEALLOCATE/DISCARD ALL */
+		{
+			const char* tag;
+			if (mbuf_get_string(&pkt->data, &tag)) {
+				if (strcmp(tag, "DEALLOCATE ALL") == 0 ||
+					strcmp(tag, "DISCARD ALL") == 0)
```
JelteF, line 374: We should have a test for both of these

lib
JelteF, line 1: You seem to have accidentally changed (rolled back?) some changes to the libusual submodule. Can you undo those?

JelteF (changes requested):
Code looks good. All it now needs is a test.

JelteF:
Also CI is complaining about the codeformatting. `make format` should fix that.

lib
knizhnik: Sorry, fixed

src/server.c
knizhnik: Added

JelteF (approved):

JelteF:
> How might this affect non-PostgreSQL uses of PgBouncer.

To be clear, you're talking about other databases that use the Postgres protocol, but don't speak Postgres its SQL dialect, right? If so, then tl;dr I think the impact would be negligable. 

The "CommandComplete parsing" here would only have any impact in practice if the server responded with `DISCARD ALL`/`DEALLOCATE ALL` to a query. While not a strict requirement [the protocol description of the CommandComplete packet][1] indicates that this really only should be done when a `DEALLOCATE ALL`/`DISCARD ALL` query was sent.
> The command tag. This is usually a single word that identifies which SQL command was completed. 

While in theory it's possible that some other non-PostgreSQL server implements a `DEALLOCATE ALL`/`DISCARD ALL` command with different semantics than Postgres, it seems rather unlikely that in practice such different semantics would not include clearing of the prepared statements. Especially because there's no protocol level message that can be used for the purpose of clearing all prepared statements. So in practice I don't think this would impact anyone badly. 

[1]: https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-COMMANDCOMPLETE

PR #979, opened 2023-11-02
Support configuring auth_query per database

### LINKED ISSUE #910: auth_query per database
In our case auth_query will contain encryption key which is different per database.

It would be helpful if auth_query would be configurable per database.

GuanqunYang193:
This adds support for allowing specifing a different auth_query for different databases. This is useful if the auth_query needs to contain an encryption key that can differ per database.

Fixes #910

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -1170,7 +1170,7 @@ Override of the global `auth_user` setting, if specified.
 
 ### auth_query
 
-Override of the global `auth_query` setting, if specified. The entire SQL statement needs to be enclosed in double quotes.
+Override of the global `auth_query` setting, if specified. The entire SQL statement needs to be enclosed in single quotes.
 
 ### pool_size
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -1169,6 +1169,10 @@ If no password is specified here, the password from the `auth_file` or
 
 Override of the global `auth_user` setting, if specified.
 
+### auth_query
+
+Override of the global `auth_auery` setting, if specified. The entire SQL statement needs to be enclosed in quotes.
+
 ### pool_size
 
 Set the maximum size of pools for this database.  If not set,
--- a/src/loader.c
+++ b/src/loader.c
@@ -134,22 +134,22 @@ static bool strings_equal(const char *str_left, const char *str_right)
 	return strcmp(str_left, str_right) == 0;
 }
 
-static bool set_auth_dbname(PgDatabase *db, const char *new_auth_dbname)
+static bool set_value(char **old_value, const char *new_value)
 {
-	if (strings_equal(db->auth_dbname, new_auth_dbname))
+	if (strings_equal(*old_value, new_value))
 		return true;
 
-	if (db->auth_dbname)
-		free(db->auth_dbname);
+	if (*old_value)
+		free(*old_value);
 
-	if (new_auth_dbname) {
-		db->auth_dbname = strdup(new_auth_dbname);
-		if (!db->auth_dbname) {
-			log_error("auth_dbname %s could not be set for database %s, out of memory", new_auth_dbname, db->name);
+	if (new_value) {
+		*old_value = strdup(new_value);
+		if (!(*old_value)) {
+			log_error("out of memory");
 			return false;
 		}
 	} else {
-		db->auth_dbname = NULL;
+		*old_value = NULL;
 	}
 
 	return true;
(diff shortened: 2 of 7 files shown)


JelteF:
Is your reason for wanting this the same as in issue #910? i.e. an encryption key that differs per database. Or do you have another usecase?

doc/config.md
```diff
@@ -1169,6 +1169,10 @@ If no password is specified here, the password from the `auth_file` or
 
 Override of the global `auth_user` setting, if specified.
 
+### auth_query
+
+Override of the global `auth_auery` setting, if specified. The entire SQL statement needs to be enclosed in quotes.
```
JelteF, line 1174:
```suggestion
Override of the global `auth_query` setting, if specified. The entire SQL statement needs to be enclosed in double quotes.
```

GuanqunYang193:
> Is your reason for wanting this the same as in issue #910? i.e. an encryption key that differs per database. Or do you have another usecase?

no just the same use case

test/test_auth.py
```diff
@@ -340,6 +340,70 @@ def test_auth_dbname_usage_global_setting(
...
+    """
+    Check the pgbouncer can use auth_query in database section to get password
+    """
+
+    config = f"""
+        [databases]
+        postgres = auth_query='SELECT usename, passwd FROM pg_shadow where usename = $1'\
+            host={bouncer.pg.host} port={bouncer.pg.port}
+        [pgbouncer]
+        auth_query = SELECT usename, passwd FROM pg_shadow where usename = $1
```
JelteF, line 356: This is the same as the one in the databases section. Let's change it to something useless like `SELECT 1`.

eulerto: Don't we want to test that it works? If so, I suggest that a functional test is added. Create a table, insert one row and test if the authentication works.

```
create table auth_query_table (u text, p text);
insert into auth_query_table VALUES('u0', 'p0');
```

I would also suggest that the used auth_query is exposed for debug purposes.

```
-   slog_noise(client, "doing auth_conn query");
+   slog_noise(client, "doing auth_conn query: %s", auth_query);
```

JelteF: > Don't we want to test that it works?

Your proposed test would work too afaict. But with the current test we test that setting auth_query in the database section has effect. i.e. if it didn't have any effect, then the global broken one would be used and you'd get an error like this in the logs:
```
BOUNCER_LOG
2023-11-03 16:45:42.916 CET [1845812] LOG S-0x5633b3d02dd0: postgres/pswcheck@127.0.0.1:10202 closing because: error response from auth_query (age=0s)
2023-11-03 16:45:42.916 CET [1845812] LOG C-0x5633b3cf5f50: postgres/(nouser)@127.0.0.1:53118 closing because: bouncer config error (age=0s)

PG_LOG
2023-11-03 16:45:42.916 CET [1845814] ERROR:  bind message supplies 1 parameters, but prepared statement "" requires 0
2023-11-03 16:45:42.916 CET [1845814] STATEMENT:  SELECT 1
``` 

So I think the current test is is acceptable at least.

eulerto: WFM.

doc/config.md
```diff
@@ -1168,6 +1168,10 @@ If no password is specified here, the password from the `auth_file` or
 
 Override of the global `auth_user` setting, if specified.
 
+### auth_query
+
+Override of the global `auth_query` setting, if specified. The entire SQL statement needs to be enclosed in double quotes.
```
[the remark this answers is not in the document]
GuanqunYang193: Yes, you are right, I used single quotes for testing but forgot it. Thanks for pointing it out. Indeed this is no doc about quotes format. I found this by reading the source code of cstr_get_pair.

src/loader.c
```diff
@@ -134,22 +134,25 @@ static bool strings_equal(const char *str_left, const char *str_right)
 	return strcmp(str_left, str_right) == 0;
 }
 
-static bool set_auth_dbname(PgDatabase *db, const char *new_auth_dbname)
+/*
+ * Free the old value and set the new value
+ */
+static bool set_value(char **old_value, const char *new_value)
```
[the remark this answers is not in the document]
GuanqunYang193: good suggestion

JelteF (approved):
LGTM now.

eulerto (approved):
LGTM.

PR #985, opened 2023-11-15
Fix connection to non-existing database - fail fast

lubennikovaav:
Check for `wait_for_user_conn` flag, when connecting to non-existing database to return error to the client immediately, instead of waiting in a new connection loop  until `query_wait_timeout`.

fixes https://github.com/pgbouncer/pgbouncer/issues/989


JelteF:
I created #998 to fix this in a different way. @lubennikovaav/@khanova please let me know what you think of that one and if it indeed fixes your issue too.

PR #996, opened 2023-12-18
Introduce auth_ident_file setting and enhance cert and peer auth methods with user name maps

### LINKED ISSUE #148: Support for username map= parameter and username maps
We have auto-scaling EC2 instances that get issued certificates with Common Names containing the EC2 instance ID. Our intent is to use pgbouncer in hostssl/cert mode, but support is needed for the username map parameter in HBA (and some kind of username mapping similar to postgresql's pg_ident.conf) to map the certificate CN to a known role. Ideally, the feature would mimic postgresql's implementation (i.e. with regexp support etc).

The documentation (https://pgbouncer.github.io/config.html#hba-file-format) says:

> username map (map=) parameter is not supported.

Keen to get this feature sooner rather than later. Can any of the existing maintainers help us out?

### LINKED ISSUE #767: Certificate based authentication issues in pgBouncer 1.17.0
we are facing an issue with enabling cert based authentication between clients and pgbouncer 1.17.0. Pgbouncer 1.17.0 and postgresql 14 are on the same host in different containers. There is TLS enabled between client and pgbouncer. The issue is that we want client to connect to pgbouncer with client key and certificate, the certificate will have a Common Name (lets say "abc") which wont be same as the user "joe" which is trying to login to pgbouncer, so here we want a mapping of the user "joe" to "abc". If we try to login as joe , Immediately we get error because pgbouncer will check the user ("joe") and Common Name ("abc") in cert and they doesn't match. What is the right way of solving this issue? otherwise how can different users login to pgBouncer?

On postgreSQL, this mapping can be done using pg_ident.conf but that support is not available in pgBouncer. What is the most secured way of achieving this authentication?

emelsimsek:
Postgres supports user name maps[1] for `cert`  and `peer` authentication methods. This change provides the same functionality with some limitations for pgbouncer. The limitations are listed in "Ident map file format" section of the docs.

1. https://www.postgresql.org/docs/current/auth-username-maps.html

Fixes #767
Fixes #148

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -449,13 +449,15 @@ pam
 
 ### auth_hba_file
 
-HBA configuration file to use when `auth_type` is `hba`.
+HBA configuration file to use when `auth_type` is `hba`. See
+section [HBA file format](#hba-file-format) below about details.
 
 Default: not set
 
 ### auth_ident_file
 
-Identity map file to use when `auth_type` is `hba` and a user map will be defined.
+Identity map file to use when `auth_type` is `hba` and a user map will be defined.  See
+section [Ident map file format](#ident-map-file-format) below about details.
 
 Default: not set
 
@@ -1419,8 +1421,20 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
 * Address field: Supports IPv4, IPv6.  Not supported: DNS names, domain prefixes.
 * Auth-method field: Only methods supported by PgBouncer's `auth_type`
   are supported, plus `peer` and `reject`, but except `any` and `pam`, which only work globally.
-  User name map (`map=`) parameter is not supported.
+* User name map (`map = `) parameter is only supported for `auth_type` `cert`.
 
+## Ident map file format
+
+The location of the ident map file is specified by the setting
+`auth_ident_file`. It is only loaded if `auth_type` is set to `hba`.
+
+The file format is a simplified variation of the PostgreSQL ident map file
+(see <https://www.postgresql.org/docs/current/auth-username-maps.html>).
+
+* Supported lines are only of the form `map-name system-username database-username`.
+* There is no support for including file/directory.
+* System-username field: Not supported: regular expressions.
+* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, multiple names.
 
 ## Examples
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -1421,7 +1421,7 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
 * Address field: Supports IPv4, IPv6.  Not supported: DNS names, domain prefixes.
 * Auth-method field: Only methods supported by PgBouncer's `auth_type`
   are supported, plus `peer` and `reject`, but except `any` and `pam`, which only work globally.
-* User name map (`map = `) parameter is only supported for `auth_type` `cert`.
+* User name map (`map=`) parameter is only supported for `auth_type` `cert`.
 
 ## Ident map file format
 
@@ -1434,7 +1434,7 @@ The file format is a simplified variation of the PostgreSQL ident map file
 * Supported lines are only of the form `map-name system-username database-username`.
 * There is no support for including file/directory.
 * System-username field: Not supported: regular expressions.
-* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, multiple names, regular expressions.
+* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, regular expressions.
 
 ## Examples
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -1434,7 +1434,7 @@ The file format is a simplified variation of the PostgreSQL ident map file
 * Supported lines are only of the form `map-name system-username database-username`.
 * There is no support for including file/directory.
 * System-username field: Not supported: regular expressions.
-* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, multiple names.
+* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, multiple names, regular expressions.
 
 ## Examples
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -453,7 +453,7 @@ HBA configuration file to use when `auth_type` is `hba`.
 
 Default: not set
 
-### auth_pgident_file
+### auth_ident_file
 
 Identity map file to use when `auth_type` is `hba` and a user map will be defined.
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -1421,7 +1421,7 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
 * Address field: Supports IPv4, IPv6.  Not supported: DNS names, domain prefixes.
 * Auth-method field: Only methods supported by PgBouncer's `auth_type`
   are supported, plus `peer` and `reject`, but except `any` and `pam`, which only work globally.
-* User name map (`map=`) parameter is only supported for `auth_type` `cert`.
+* User name map (`map=`) parameter is supported when `auth_type` is `cert` or `peer`. 
 
 ## Ident map file format
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -453,6 +453,12 @@ HBA configuration file to use when `auth_type` is `hba`.
 
 Default: not set
 
+### auth_pgident_file
+
+Identity map file to use when `auth_type` is `hba` and a user map will be defined.
+
+Default: not set
+
 ### auth_file
 
 The name of the file to load user names and passwords from.  See
--- a/include/hba.h
+++ b/include/hba.h
@@ -15,6 +15,9 @@
  * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
+
+#define NAME_ALL        1
+
 enum RuleType {
 	RULE_LOCAL,
 	RULE_HOST,
@@ -52,6 +55,7 @@ struct IDENTMap {
 	char *map_name;
 	char *system_user_name;
 	char *postgres_user_name;
+	unsigned int name_flags;
 };
 
 struct IDENT {
--- a/include/hba.h
+++ b/include/hba.h
@@ -15,9 +15,51 @@
  * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  */
+enum RuleType {
+	RULE_LOCAL,
+	RULE_HOST,
+	RULE_HOSTSSL,
+	RULE_HOSTNOSSL,
+};
 
-struct HBA;
+struct NameSlot {
+	size_t strlen;
+	char str[];
+};
+struct HBAName {
+	unsigned int flags;
+	struct StrSet *name_set;
+};
 
-struct HBA *hba_load_rules(const char *fn);
+struct HBARule {
+	struct List node;
+	enum RuleType rule_type;
+	int rule_method;
+	int rule_af;
+	uint8_t rule_addr[16];
+	uint8_t rule_mask[16];
+	struct HBAName db_name;
+	struct HBAName user_name;
+	struct IDENTMap *identmap;
+};
+
+struct HBA {
+	struct List rules;
+};
+
+struct IDENTMap {
+	struct List node;
+	char *map_name;
+	char *system_user_name;
+	char *postgres_user_name;
+};
+
+struct IDENT {
+	struct List maps;
+};
+
+struct IDENT *ident_load_map(const char *fn);
+void ident_free(struct IDENT *ident);
+struct HBA *hba_load_rules(const char *fn, struct IDENT *ident);
 void hba_free(struct HBA *hba);
-int hba_eval(struct HBA *hba, PgAddr *addr, bool is_tls, const char *dbname, const char *username);
+struct HBARule *hba_eval(struct HBA *hba, PgAddr *addr, bool is_tls, const char *dbname, const char *username);
--- a/include/hba.h
+++ b/include/hba.h
@@ -17,6 +17,7 @@
  */
 
 #define NAME_ALL        1
+#define NAME_SAMEUSER   2
 
 enum RuleType {
 	RULE_LOCAL,
--- a/include/hba.h
+++ b/include/hba.h
@@ -44,27 +44,32 @@ struct HBARule {
 	uint8_t rule_mask[16];
 	struct HBAName db_name;
 	struct HBAName user_name;
-	struct IDENTMap *identmap;
+	struct IdentMap *identmap;
 };
 
 struct HBA {
 	struct List rules;
 };
 
-struct IDENTMap {
+struct Mapping {
 	struct List node;
-	char *map_name;
 	char *system_user_name;
 	char *postgres_user_name;
 	unsigned int name_flags;
 };
 
-struct IDENT {
+struct IdentMap {
+	struct List node;
+	char *map_name;
+	struct List mappings;
+};
+
+struct Ident {
 	struct List maps;
 };
 
-struct IDENT *ident_load_map(const char *fn);
-void ident_free(struct IDENT *ident);
-struct HBA *hba_load_rules(const char *fn, struct IDENT *ident);
+struct Ident *ident_load_map(const char *fn);
+void ident_free(struct Ident *ident);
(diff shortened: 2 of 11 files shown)


doc/config.md
```diff
@@ -448,6 +448,12 @@ HBA configuration file to use when `auth_type` is `hba`.
 
 Default: not set
 
+### auth_pgident_file
```
JelteF, line 451: Let's call this auth_ident_file to use teh same convention as hba_file (that's not called auth_pghba_file either).
```suggestion
### auth_ident_file
```

test/test_auth.py
```diff
@@ -475,3 +475,37 @@ def test_auth_dbname_works_fine(
...
+    bouncer.write_ini(f"client_tls_ca_file = {root}")
+    bouncer.write_ini(f"client_tls_sslmode = require")
+    bouncer.write_ini(f"auth_type = hba")
+    bouncer.write_ini(
+        f"auth_query = SELECT usename, passwd FROM pg_shadow where usename = $1"
+    )
+    bouncer.write_ini(f"auth_user = pswcheck")
+    bouncer.write_ini(f"auth_file = {bouncer.auth_path}")
+    bouncer.write_ini(f"auth_hba_file = pgbouncer_hba.conf")
+    bouncer.write_ini(f"auth_pgident_file = pgident.conf")
```
JelteF, line 497: These two files are not committed, which is why this test is failing.

test/test_auth.py
```diff
@@ -475,3 +475,37 @@ def test_auth_dbname_works_fine(
         # The client connects to postgres DB that matches with autodb
         # pgbouncer must use postgres_authdb1, which is defined in [pgbouncer] section
         bouncer.test(user="stats", password="stats", dbname="postgres")
+
+
+@pytest.mark.skipif("WINDOWS", reason="Windows does not have SIGHUP")
+def test_client_hba_cert(bouncer, cert_dir):
```
JelteF, line 481: Some additional tests that should be added:
- [ ] Rejects connection with the wrong CN in the cert (e.g. try connecting with localhost cert)
- [ ] Rejects connection when CN matches, but mapname doesn't match.
- [ ] "all" allows connection with any user

doc/config.md
```diff
@@ -1404,8 +1412,20 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
...
+The location of the ident map file is specified by the setting
+`auth_ident_file`. It is only loaded if `auth_type` is set to `hba`.
+
+The file format is a simplified variation of the PostgreSQL ident map file
+(see <https://www.postgresql.org/docs/current/auth-username-maps.html>).
 
+* Supported lines are only of the form `map-name system-username database-username`.
+* There is no support for including file/directory.
+* System-username field: Not supported: regular expressions.
+* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, multiple names.
```
JelteF, line 1428:
```suggestion
* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, multiple names, regular expressions.
```

src/client.c
```diff
@@ -221,9 +221,22 @@ static bool login_via_cert(PgSocket *client)
 		goto fail;
 
 	log_debug("TLS cert login: %s", tls_peer_cert_subject(client->sbuf.tls));
+
 	if (!tls_peer_cert_contains_name(client->sbuf.tls, client->login_user->name)) {
-		slog_error(client, "TLS certificate name mismatch");
-		goto fail;
+		log_debug("TLS certificate name mismatch. Looking for a map.");
```
JelteF, line 226: This logic is incorrect in a small but very important way. W should only try a direct match if no map is defined. If a map is defined, it's very much possible that a direct match is actually not allowed to give access.

src/client.c
```diff
@@ -221,9 +221,22 @@ static bool login_via_cert(PgSocket *client)
...
+			goto fail;
+
+		if (rule->identmap == NULL)
+			goto fail;
+
+		if (!tls_peer_cert_contains_name(client->sbuf.tls, rule->identmap->system_user_name))
+			goto fail;
+
+		if (strcmp(client->login_user->name, rule->identmap->postgres_user_name)) {
+			if (strcmp(rule->identmap->postgres_user_name, "all"))
```
JelteF, line 237: There's two ways all can be formatted in the ident file, which means two different things:
- `all` (literal string), this is the special keyword that matches all users
- `"all"` (quoted), this means it only matches the user called `all` (assuming this user exists)

Postgres tracks this difference by setting a flag in case the special keyword is used.

src/main.c
```diff
@@ -444,7 +447,19 @@ void load_config(void)
...
 
 	if (cf_auth_type == AUTH_HBA) {
-		struct HBA *hba = hba_load_rules(cf_auth_hba_file);
+		struct IDENT *ident;
+		struct HBA *hba;
+
+		ident = ident_load_map(cf_auth_ident_file);
+
+		if (ident) {
+			if (parsed_ident)
```
JelteF, line 456: nit: this check is unnecessary, you can always free even if it is NULL (then free is a no-op)

src/client.c
JelteF: For HBA parsing we use the `NAME_ALL` flag in pgbouncer too actually.

src/hba.c
```diff
@@ -499,7 +467,143 @@ static bool bad_mask(struct HBARule *rule)
...
+		if (strcmp(rest, "=") == 0) {
+			next_token(tp);
+
+			if (!expect(tp, TOK_IDENT, &mapname))
+				return false;
+		} else if (rest[0] == '=') {
+			mapname = rest + 1;
+		} else {
+			return false;
+		}
```
JelteF, line 516: Afaict this logic is to handle spaces between the `=` sign an `map`. This is unnecessary, postgres also does not support this: https://github.com/postgres/postgres/blob/master/src/backend/libpq/hba.c#L1856-L1884

Also, as you can see in the `parse_hba_auth_opt` function, Postgres actually supports a bunch of these options in hba lines. Let's at least prepare our code to do the same (i.e. using `strchr` to get the option name). That way we can give more useful error messages, e.g. option clientcert is not supported.

doc/config.md
```diff
@@ -1408,8 +1416,20 @@ The file follows the format of the PostgreSQL `pg_hba.conf` file
...
+The location of the ident map file is specified by the setting
+`auth_ident_file`. It is only loaded if `auth_type` is set to `hba`.
+
+The file format is a simplified variation of the PostgreSQL ident map file
+(see <https://www.postgresql.org/docs/current/auth-username-maps.html>).
 
+* Supported lines are only of the form `map-name system-username database-username`.
+* There is no support for including file/directory.
+* System-username field: Not supported: regular expressions.
+* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, multiple names, regular expressions.
```
JelteF, line 1432: Multiple values are not supported by Postgres either in the ident file (so let's make this limitation implicit).

```suggestion
* Database-username field: Supports `all` or a single postgres user name. Not supported: `+groupname`, regular expressions.
```

src/hba.c
```diff
@@ -24,41 +24,8 @@
...
 
-enum RuleType {
-	RULE_LOCAL,
-	RULE_HOST,
-	RULE_HOSTSSL,
-	RULE_HOSTNOSSL,
-};
-
-#define NAME_ALL        1
 #define NAME_SAMEUSER   2
```
JelteF, line 27: Let's move this one to hba.h too, to keep these defines together.

onurctirtir:
SIGSEGV happens when trying to connect to a database that's present in ini file:

```c
#0  0x000055a024613e61 in finish_set_pool (client=client@entry=0x55a025a09a80, takeover=takeover@entry=false) at src/client.c:310
#1  0x000055a024615918 in handle_client_startup (client=client@entry=0x55a025a09a80, pkt=pkt@entry=0x7ffc669afb10) at src/client.c:928
#2  0x000055a024616549 in client_proto (sbuf=0x55a025a09c50, evtype=<optimized out>, data=0x7ffc669afb90) at src/client.c:1365
#3  0x000055a024628f68 in sbuf_call_proto (sbuf=sbuf@entry=0x55a025a09c50, event=event@entry=0) at src/sbuf.c:498
#4  0x000055a0246294fa in sbuf_process_pending (sbuf=sbuf@entry=0x55a025a09c50) at src/sbuf.c:802
#5  0x000055a024629e24 in sbuf_main_loop (skip_recv=<optimized out>, sbuf=0x55a025a09c50) at src/sbuf.c:1031
#6  sbuf_main_loop (sbuf=0x55a025a09c50, skip_recv=<optimized out>) at src/sbuf.c:966
#7  0x000055a024616230 in handle_auth_query_response (client=0x55a025a09a80, pkt=pkt@entry=0x7ffc669b05f0) at src/client.c:545
#8  0x000055a02462cf56 in handle_server_work (pkt=0x7ffc669b05f0, server=0x55a025a23a80) at src/server.c:458
#9  server_proto (sbuf=0x55a025a23c50, evtype=<optimized out>, data=<optimized out>) at src/server.c:702
#10 0x000055a024628f68 in sbuf_call_proto (sbuf=sbuf@entry=0x55a025a23c50, event=event@entry=0) at src/sbuf.c:498
#11 0x000055a0246294fa in sbuf_process_pending (sbuf=sbuf@entry=0x55a025a23c50) at src/sbuf.c:802
#12 0x000055a024629e24 in sbuf_main_loop (skip_recv=<optimized out>, sbuf=0x55a025a23c50) at src/sbuf.c:1031
#13 sbuf_main_loop (sbuf=0x55a025a23c50, skip_recv=<optimized out>) at src/sbuf.c:966
#14 0x00007ff08f61e13f in ?? () from /lib/x86_64-linux-gnu/libevent-2.1.so.7
#15 0x00007ff08f61e87f in event_base_loop () from /lib/x86_64-linux-gnu/libevent-2.1.so.7
#16 0x000055a02461ccba in main_loop_once () at src/main.c:794
#17 0x000055a02460ec45 in main (argc=<optimized out>, argv=<optimized out>) at src/main.c:1082
```

Assuming that `ident_map` is non-NULL seems problematic at src/client.c:310:
```c
log_warning("hba_eval returned Ident map %s %s %s", rule->identmap->map_name, rule->identmap->system_user_name, rule->identmap->postgres_user_name);
```

src/client.c
```diff
@@ -221,8 +221,16 @@ static bool login_via_cert(PgSocket *client)
 		goto fail;
 
 	log_debug("TLS cert login: %s", tls_peer_cert_subject(client->sbuf.tls));
-	if (!tls_peer_cert_contains_name(client->sbuf.tls, client->login_user->name)) {
-		slog_error(client, "TLS certificate name mismatch");
+
+	if (rule && rule->identmap) {
+		if (!tls_peer_cert_contains_name(client->sbuf.tls, rule->identmap->system_user_name))
```
JelteF, line 226: We should log errors here, failures in these kind of things are hard enough to debug already.

src/client.c
```diff
@@ -221,8 +221,16 @@ static bool login_via_cert(PgSocket *client)
 		goto fail;
 
 	log_debug("TLS cert login: %s", tls_peer_cert_subject(client->sbuf.tls));
-	if (!tls_peer_cert_contains_name(client->sbuf.tls, client->login_user->name)) {
-		slog_error(client, "TLS certificate name mismatch");
+
+	if (rule && rule->identmap) {
```
JelteF, line 225: Let's add similar logic to use the `map` when available to `login_as_unix_peer`, because postgres allow username mapping for that one too: https://www.postgresql.org/docs/current/auth-peer.html

(and afaict to support that we only similar 5 lines)

test/pgident.conf
```diff
@@ -0,0 +1,6 @@
+# PostgreSQL User Name Maps
+# =========================
+
+# MAPNAME       SYSTEM-USERNAME         PG-USERNAME
+test		pgbouncer.acme.org	someuser
```
JelteF, line 5: I think it would be good to test a case where there are multiple mappings for the same name, then the cert should be able to authenticate as both:
```suggestion
test		pgbouncer.acme.org	someuser
test		pgbouncer.acme.org	someuser2
```

test/test_auth.py
```diff
@@ -576,3 +577,118 @@ async def test_change_server_password_server_lifetime(bouncer, pg):
...
+    bouncer.psql_test(
+        host="localhost",
+        user="someuser",
+        sslmode="verify-full",
+        sslkey=client_key,
+        sslcert=client_cert,
+        sslrootcert=root,
+    )
+
+    bouncer.pg.sql("create user anotheruser with login;")
```
JelteF, line 622: sidenote: We should change this for `bouncer.pg.create_user` once #876 is merged

include/hba.h
```diff
@@ -16,8 +16,55 @@
...
+	struct HBAName db_name;
+	struct HBAName user_name;
+	struct IDENTMap *identmap;
+};
+
+struct HBA {
+	struct List rules;
+};
+
+struct IDENTMap {
```
JelteF, line 54: Let's use CamelCase for the struct name. I realize took the naming scheme from HBA, but that is an acronym (and even then I don't agree with this type of casing personally).
```suggestion
struct IdentMap {
```

src/hba.c
```diff
@@ -499,7 +481,131 @@ static bool bad_mask(struct HBARule *rule)
...
+		goto failed;
+	}
+
+	ident_map->map_name = strdup(map_name);
+
+	next_token(tp);
+
+	if (!expect(tp, TOK_IDENT, &system_user_name)) {
+		goto failed;
+	}
```
JelteF, line 575: This one should allow TOK_STRING too.

test/pgident.conf
```diff
@@ -0,0 +1,6 @@
...
+test2		bouncer			all
```
JelteF, line 6: We should have some tests for quoted names too in all of the columns.

src/hba.c
```diff
@@ -499,7 +467,143 @@ static bool bad_mask(struct HBARule *rule)
...
+
+	ident_map = calloc(sizeof *ident_map, 1);
+
+	if (!ident_map) {
+		log_warning("ident: no mem for parsing ident map");
+		return false;
+	}
+
+	if (!expect(tp, TOK_IDENT, &map_name)) {
```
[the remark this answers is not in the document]
JelteF: This one still requires TOK_IDENT for the map_name afaict.

JelteF (changes requested):

test/pgident.conf
emelsimsek: The current design does not support this.  The first encountered map is effective. But I see that postgres supports multiple mappings with the same name:
https://www.postgresql.org/docs/current/auth-username-maps.html#:~:text=%23%20MAPNAME%20%20%20%20%20%20%20SYSTEM%2DUSERNAME,omicron%20%20%20%20%20%20%20%20%20bryanh%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20guest1

Need to make data structure changes to support this.

src/client.c
emelsimsek: I have done this change but could not add tests for it.  I was able to manually test.

I tried adding a test via: 

 result = subprocess.run([f"**sudo -u asystemuser psql -p {bouncer.port} -h  {bouncer.admin_host} p0 -c \"SELECT 1;\"**"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
    print(result.returncode, result.stdout, result.stderr)

returns error 

..... Permission denied\npsql: error: connection to server on socket **"/tmp/pytest-of-emel/pytest-34/test_client_hba_cert0/bouncer/.s.PGSQL.10202**" failed: Permission denied\n\tIs the server running locally and accepting connections on that socket?\n'

emelsimsek: Maybe we should make this a seperate PR with proper tests added.

JelteF: I think probably the easiest way to create a test is to manually create an ident file, that contains the current system user. So something like this (untested code):
```python
cur_user = os.current_user()
with open("myident.conf", 'w') as f:
    f.write(f"mymap {cur_user} postgres")	

bouncer.write_ini(f"auth_ident_file = pgident.conf")
```

JelteF: > Maybe we should make this a seperate PR with proper tests added.

One big reason why I'd like it in too, is because it's generally much easier to do some manual testing with it.

emelsimsek: Thanks @JelteF. I pushed a simple test now.

test/test_auth.py
```diff
@@ -576,3 +578,219 @@ async def test_change_server_password_server_lifetime(bouncer, pg):
...
+        sslcert=client_cert,
+        sslrootcert=root,
+    )
+
+
+@pytest.mark.skipif("WINDOWS", reason="Windows does not have SIGHUP")
+def test_peer_auth_ident_map(bouncer):
+    cur_user = getpass.getuser()
+
+    with open("ident.conf", "w") as f:
```
JelteF, line 754: These files should be created in the config directory of pgbouncer. Otherwise they polute the current working directory when running the tests locally.

JelteF (approved):
LGTM now. But as already discussed in private let's wait until we merged and released #1025 as a patch release

PR #998, opened 2023-12-22
Improve connection fail fast logic

### LINKED ISSUE #989: When connecting to a new non-existing database pgbouncer doesn't report it
How to reproduce:

I have the latest version of the pgbouncer repo. And I build pgbouncer from sources on Ubuntu 22.

<details>
<summary>test.ini (I just replaced `[databases]` with the autodb: `* = port=6666 host=127.0.0.1 auth_user=postgres password=password`):</summary>

```
[databases]
* = port=6666 host=127.0.0.1 auth_user=postgres password=password

[users]
maxedout = max_user_connections=3
maxedout2 = max_user_connections=2

[pgbouncer]
logfile = test.log
pidfile = test.pid

listen_addr = 127.0.0.1
listen_port = 6667
unix_socket_dir = /tmp

auth_type = trust
auth_file = userlist.txt

pool_mode = statement

server_check_delay = 10
max_client_conn = 10
default_pool_size = 5

server_lifetime = 120
server_idle_timeout = 60

tcp_defer_accept = 0
tcp_keepalive = 0

; use a small tcp_socket_buffer so that we trigger EAGAIN more often in tests.
; Otherwise bugs caused by EAGAIN are hard to debug.
tcp_socket_buffer = 4096

;track_extra_parameters is case sensitive.
;bgbouncer will only track the parameters with GUC_REPORT flag set in Postgres
track_extra_parameters = search_path, intervalstyle

```
</details>

test:
```py
def test_pgbouncer_stuck(bouncer):
    bouncer.admin(f"set verbose=6")
    utils.run(["psql", f"port={bouncer.port} host={bouncer.host} dbname=fake"], shell=False, capture_output=True)
```

Expected behaviour: immediately return error to the client.

Actual behaviour: falls in the `launching new connection` loop until `query_wait_timeout`.

I suspect that the issue is with [this line ](https://github.com/pgbouncer/pgbouncer/blame/master/src/server.c#L76).

A bit of debugging showed me that the client has `wait_for_welcome` equal to false, however `wait_for_user_conn` is true.
I think that replacing it with:
```c
if (!client->wait_for_welcome && !client->wait_for_user_conn)
	continue;
```
Added suggested fix with the corresponding test in this PR: https://github.com/pgbouncer/pgbouncer/pull/985.

However I am not sure whether the described problem could be a symptom of larger bug.

JelteF:
For a very long time, we've had logic to quickly close all clients when
the a server could not be connected to.

This wasn't working correctly in all cases, one of those was when an
`auth_user` was set (and no `auth_db`). This was reported in #989.

While looking closer into that issue I realized it also wouldn't handle
cases where first connecting worked but then after a while it doesn't
anymore, because it relied only on `wait_for_welcome`. It's not strictly
necessary to kill all waiting connections when creating new server
connections is not possible anymore, as long as there are still working
server connections. But if those don't exist either, then we're
effectively in the same state as complete refresh (except that we have
cached the welcome message).

Fixes #989
Fixes #985

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -65,22 +65,37 @@ failed_store:
 	return false;
 }
 
-/* we cannot log in at all, notify clients */
+/*
+ * We cannot log in to the server at all. If we don't already have any usable
+ * server connections, we disconnect all other clients in the pool that are
+ * waiting for a server.
+ */
 void kill_pool_logins(PgPool *pool, const char *sqlstate, const char *msg)
 {
 	struct List *item, *tmp;
 	PgSocket *client;
+	/*
+	 * The check for welcome_msg_ready is necessary because that indicates
+	 * that the pool got tagged as dirty. It's possible that there's still
+	 * working server connections in that case, but as soon as they get
+	 * unassigned from their client they would be closed. So they don't
+	 * really count as usable anymore.
+	 */
+	if (pool_connected_server_count(pool) != 0 && pool->welcome_msg_ready)
+		return;
 
 	statlist_for_each_safe(item, &pool->waiting_client_list, tmp) {
 		client = container_of(item, PgSocket, head);
-		if (!client->wait_for_welcome)
-			continue;
-
 		disconnect_client_sqlstate(client, true, sqlstate, msg);
 	}
 }
 
-/* we cannot log in at all, notify clients with server error */
+/*
+ * We cannot log in to the server at all. If we don't already have any usable
+ * server connections, we disconnect all other clients in the pool that are
+ * also waiting for a server. We disconnect them with exactly the same error
+ * message and code as we received from the server.
+ */
 static void kill_pool_logins_server_error(PgPool *pool, PktHdr *errpkt)
 {
 	const char *level, *msg, *sqlstate;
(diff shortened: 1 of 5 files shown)


khanova:
Thanks a lot for looking into it!
This PR fixes our issue with pgbouncer.

src/server.c
```diff
@@ -133,10 +148,7 @@ static bool handle_server_startup(PgSocket *server, PktHdr *pkt)
 		break;
 
 	case 'E':		/* ErrorResponse */
-		if (!server->pool->welcome_msg_ready)
-			kill_pool_logins_server_error(server->pool, pkt);
-		else
-			log_server_error("S: login failed", pkt);
+		kill_pool_logins_server_error(server->pool, pkt);
```
emelsimsek, line 151: Why to remove the below log?
`log_server_error("S: login failed", pkt);`

emelsimsek:
Changes look fine to me.

src/server.c
JelteF: We now always call `kill_pool_logins_server_error` and that already logs the error using:
```c
log_warning("server login failed: %s %s", level, msg);
```

(it's a warning not an error this time, but that seems like a not very important change)

PR #1001, opened 2023-12-27
Fix false positive reported by valgrind when using hba

### LINKED ISSUE #1000: Valgrind reports possible leak with an hba.conf file which includes db, user names other than "all"
When `auth_type= hba `is used, CI runs report **possible memory leak**  when hba.conf while has DATABASE or USER name other than all.

I can reproduce the error report as below. 

### 1. Create the following files

cat test/test_leak.py
```python
import re
import subprocess
import time

import psycopg
import pytest

from .utils import LONG_PASSWORD, PG_SUPPORTS_SCRAM, TLS_SUPPORT, WINDOWS


@pytest.mark.skipif("WINDOWS", reason="Windows does not have SIGHUP")
@pytest.mark.skipif(not TLS_SUPPORT, reason="pgbouncer is built without TLS support")
def test_leak_report(bouncer, cert_dir):
    bouncer.write_ini(f"auth_type = hba")
    bouncer.write_ini(f"auth_hba_file = hba.conf")

    bouncer.admin("reload")

```
cat test/hba.conf
```
local   p0              all                              peer
```

### 2. Run test using valgrind

ENABLE_VALGRIND=yes pytest test/test_leak.py

.....

  ```
VALGRIND LOG /tmp/pytest-of-emel/pytest-39/test_leak_report0/bouncer/valgrind.36460.log
  
  ==36460== VALGRIND-ERROR-BEGIN
  ==36460== 1,104 bytes in 1 blocks are possibly lost in loss record 258 of 280
  ==36460==    at 0x4848899: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
  ==36460==    by 0x145BA4: cx_new_pool (cxextra.c:257)
  ==36460==    by 0x11EBA6: strset_new (hba.c:89)
  ==36460==    by 0x11EF66: parse_names (hba.c:437)
  ==36460==    by 0x11F4D3: parse_line (hba.c:531)
  ==36460==    by 0x11F4D3: hba_load_rules (hba.c:639)
  ==36460==    by 0x1245A7: load_config (main.c:447)
  ==36460==    by 0x119F51: admin_cmd_reload (admin.c:1094)
  ==36460==    by 0x119F51: admin_cmd_reload (admin.c:1085)
  ==36460==    by 0x119839: exec_cmd (admin.c:94)
  ==36460==    by 0x119839: admin_parse_query (admin.c:1544)
  ==36460==    by 0x11A62D: admin_handle_client (admin.c:1593)
  ==36460==    by 0x11BBAB: handle_client_work (client.c:1205)
  ==36460==    by 0x11D9FC: client_proto (client.c:1355)
  ==36460==    by 0x1315A8: sbuf_call_proto (sbuf.c:498)
  ==36460==    by 0x1315A8: sbuf_process_pending (sbuf.c:802)
  ==36460==
  ==36460== VALGRIND-ERROR-END
```

......

### PR DESCRIPTION
**Pull request by JelteF**
To make `valgrind` pass we have a cleanup function that we only use in
assert mode. In this function we free all globals before process exit.
However, we didn't free `parsed_hba`. This fixes that. In passing it
also slightly modifies how we free `parsed_hba` when reloading the
config.

Fixes #1000

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -446,10 +446,12 @@ void load_config(void)
 	if (cf_auth_type == AUTH_HBA) {
 		struct HBA *hba = hba_load_rules(cf_auth_hba_file);
 		if (hba) {
-			if (parsed_hba)
-				hba_free(parsed_hba);
+			hba_free(parsed_hba);
 			parsed_hba = hba;
 		}
+	} else {
+		hba_free(parsed_hba);
+		parsed_hba = NULL;
 	}
 
 	/* kill dbs */
@@ -852,6 +854,8 @@ static void cleanup(void)
 	}
 	adns_free_context(adns);
 	adns = NULL;
+	hba_free(parsed_hba);
+	parsed_hba = NULL;
 
 	admin_cleanup();
 	objects_cleanup();
--- a/test/test_auth.py
+++ b/test/test_auth.py
@@ -475,3 +475,23 @@ def test_auth_dbname_works_fine(
         # The client connects to postgres DB that matches with autodb
         # pgbouncer must use postgres_authdb1, which is defined in [pgbouncer] section
         bouncer.test(user="stats", password="stats", dbname="postgres")
+
+
+def test_hba_leak(bouncer):
+    """
+    Don't actually check if HBA auth works, but check that it doesn't leak
+    memory when using the feature.
+    """
+    bouncer.write_ini(f"auth_type = hba")
+    bouncer.write_ini(f"auth_hba_file = hba_test.rules")
+
+    bouncer.admin("reload")
+
+    bouncer.write_ini(f"auth_type = trust")
+
+    bouncer.admin("reload")
+
+    bouncer.write_ini(f"auth_type = hba")
+
+    bouncer.admin("reload")
+    bouncer.admin("reload")

### CODE REVIEW

File: test/test_auth.py [hunk 1]
```diff
@@ -475,3 +475,23 @@ def test_auth_dbname_works_fine(
...
         bouncer.test(user="stats", password="stats", dbname="postgres")
+
+
+def test_hba_leak(bouncer):
+    """
+    Don't actually check if HBA auth works, but check that it doesn't leak
+    memory when using the feature.
+    """
+    bouncer.write_ini(f"auth_type = hba")
+    bouncer.write_ini(f"auth_hba_file = hba.conf")
```
emelsimsek, line 486:
```suggestion
    bouncer.write_ini(f"auth_hba_file = hba_test.rules")
```

No new conf file is needed actually. Could test with the existing `hba_test.rules`.

src/main.c
```diff
@@ -446,10 +446,12 @@ void load_config(void)
...
 		if (hba) {
-			if (parsed_hba)
-				hba_free(parsed_hba);
+			hba_free(parsed_hba);
 			parsed_hba = hba;
 		}
+	} else {
+		hba_free(parsed_hba);
+		parsed_hba = NULL;
 	}
```
emelsimsek, line 455: It looks like the else block is redundant.

We could initialized  `parsed_hba`  to NULL at line 73:

`struct HBA *parsed_hba = NULL;`

But even that is not neede.

emelsimsek (approved):

src/main.c
JelteF: load_config is rerun whenever pgbouncer receives a SIGHUP or RELOAD command. So this block is executed when auth_type changes from HBA to something else. It seemed reasonable to me we free parsed_hba whenever that happens, instead of keeping the structure allocated.

test/test_auth.py
JelteF: done

src/main.c
emelsimsek: Ok. Makes sense.

PR #1004, opened 2024-01-03
docs: add auth_dbname in section databases

tcraxs:
Add documentation in section databases about `auth_dbname` option to override global setting.


JelteF:
Thanks :)

PR #1007, opened 2024-01-12
Add support for negotiating the protocol version with the client

JelteF:
PgBouncer currently only supports protocol version 3.0. However, it was
still accepting connections if the client asked for the 3.1 protocol
version. This is obviously a bug, because PgBouncer likely won't
support the changes introduced in 3.1 of the protocol without
any code changes.

This starts sending the NegotiateProtocolVersion message when the client
asks for a minor version that PgBouncer does not support. Also, instead
of erroring when it receives unsupported protocol extensions parameters
it will include these in this NegotiateProtocolVersion message.

For reference this is the Postgres commit introducing NegotiateProtocolVersion: https://github.com/postgres/postgres/commit/ae65f6066dc3d19a55f4fdcd3b30003c5ad8dbed

I tested this manually using this patchset that bumps the protocol version. https://www.postgresql.org/message-id/flat/CAGECzQScQ3N-Ykv2j4NDyDtrPPc3FpRoa%3DLZ-2Uj2ocA4zr%3D4Q%40mail.gmail.com#cd9e8407820d492e8f677ee6a67c21ce

### DIFF
--- a/src/proto.c
+++ b/src/proto.c
@@ -103,7 +103,7 @@ bool get_header(struct MBuf *data, PktHdr *pkt)
 		} else if (code == PKT_GSSENCREQ) {
 			type = PKT_GSSENCREQ;
 		} else if ((code >> 16) == 3 && (code & 0xFFFF) < 2) {
-			type = PKT_STARTUP;
+			type = PKT_STARTUP_V3;
 		} else if (code == PKT_STARTUP_V2) {
 			type = PKT_STARTUP_V2;
 		} else {
--- a/src/proto.c
+++ b/src/proto.c
@@ -102,8 +102,10 @@ bool get_header(struct MBuf *data, PktHdr *pkt)
 			type = PKT_SSLREQ;
 		} else if (code == PKT_GSSENCREQ) {
 			type = PKT_GSSENCREQ;
-		} else if ((code >> 16) == 3 && (code & 0xFFFF) < 2) {
+		} else if (code >= PKT_STARTUP_V3 && code < PKT_STARTUP_V3_UNSUPPORTED) {
 			type = PKT_STARTUP_V3;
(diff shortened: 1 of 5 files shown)


src/proto.c
```diff
@@ -102,8 +102,10 @@ bool get_header(struct MBuf *data, PktHdr *pkt)
 			type = PKT_SSLREQ;
 		} else if (code == PKT_GSSENCREQ) {
 			type = PKT_GSSENCREQ;
-		} else if ((code >> 16) == 3 && (code & 0xFFFF) < 2) {
```
JelteF, line 105: to be clear, this was accepting protocol 3.1. It should have been this to only support 3.0:
```diff
		} else if ((code >> 16) == 3 && (code & 0xFFFF) < 2) {
		} else if ((code >> 16) == 3 && (code & 0xFFFF) < 1) {
```

levkk:
How do you test this? I'm trying to send the protocol negotiation packet while connecting with psql and it always returns this error message:

```
psql: error: connection to server at "127.0.0.1", port 6432 failed: expected authentication request from server, but received v
```

which is really weird because when I'm reading the implementation in libpq, it clearly should support `v` at this particular stage in the negotiation: https://github.com/postgres/postgres/blob/b0289574bdf1202248201a3143d1459bdf5727fd/src/interfaces/libpq/fe-connect.c#L3541-L3548

JelteF:
Sounds like you're using a libpq from before bbf9c282ce92272ed7bf6771daf0f9efa209e61b

But even after that libpq will complain. Just with a more helpful message. Better handling is in my patchset here: https://www.postgresql.org/message-id/flat/CAGECzQScQ3N-Ykv2j4NDyDtrPPc3FpRoa%3DLZ-2Uj2ocA4zr%3D4Q%40mail.gmail.com#cd9e8407820d492e8f677ee6a67c21ce

PR #1017, opened 2024-01-29
Add editorconfig rules for our hba_test files

JelteF:
Our hba_test.eval and hba_test.rules file only look well aligned with 8
size tabs. This configures editors (and GitHub) to display them as such.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,16 @@ trim_trailing_whitespace = true
 indent_style = tab
 indent_size = 8
 
+[hba_test.{eval,rules}]
+indent_style = tab
+indent_size = 8
+
+[hba_test.rules]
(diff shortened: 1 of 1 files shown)


petere:
Is it worth codifying this? How often are these files edited?

Also, pg_hba files inherently have no indentation, so it's weird to codify indentation rules for them.

JelteF:
> Is it worth codifying this? How often are these files edited?

Probably not super often, but I also don't really see a harm in codifying it. I've definitely run into this a few times since I've configured my editor (like many people) to display tabs with a width of 4 by default.

> Also, pg_hba files inherently have no indentation, so it's weird to codify indentation rules for them.

While they have no indentation, the columns in the current file are separated by tabs. So the columns only actually look aligned when using tabs of width 8.

PR #1028, opened 2024-02-26
v1.22.1

JelteF:


NEWS.md
```diff
@@ -4,6 +4,19 @@ PgBouncer changelog
 PgBouncer 1.22.x
 ----------------
 
+**2024-02-28  -  PgBouncer 1.22.1  -  "It's summer in Bangalore"**
+
+- Fixes
+  * Fix an issue involving some clients that could introduce memory leaks,
```
emelsimsek, line 10:
```suggestion
  * Fix an issue impacting some clients using `COPY FROM STDIN` queries which could cause memory leaks, performance issues and issues with `max_prepared_statements>0`
```

emelsimsek, line 10:
```suggestion
  * Fix issues caused by some clients using `COPY FROM STDIN` queries that could introduce memory leaks, performance issues and issues with `max_prepared_statements>0`  ([#1025])
    (bug introduced in 1.21.0)
```

emelsimsek (approved):

NEWS.md
eulerto: Let me suggest a different sentence:

* Fix issues caused by some clients using `COPY FROM STDIN` queries that might introduce memory leak, performance regression and prepared statement misbehavior.

JelteF: Thanks for the suggestions. I pushed @eulerto his suggestion, with a small change that splits it into two sentences to make it a bit easier to read.

eulerto (approved):

PR #1032, opened 2024-03-04
Revert "Introduce auth_ident_file setting and enhance cert and peer auth methods with user name maps"

emelsimsek:
Reverts pgbouncer/pgbouncer#996. 

Need to  squash last commits for PR #996 .


JelteF:
Reverting is only going to make commit history even messier. `git reset --hard` + force push to master would be able to clean it up, but it's way too late for that now imho. Chances that people pulled the repo seem extremely high given a release that happened on the same day. And then they would get conflicts when pulling again.

So closing this. And no worries, mistakes like this can happen. I think I'll disable merging with merge commits tomorrow on the repo tomorrow so this doesn't happen again.

PR #1039, opened 2024-03-13
Teach pgBouncer to do auth_query for users defined without a password.

### LINKED ISSUE #484: pgbouncer didn't even try auth_query to verify user password

benchub:
As described in #484, if a user is defined in the [users] section of
the config file (perhaps to take advantage of per-user overrides)
but then that user is *not* defined in auth_file, pgBouncer currently
gets confused when trying to check their password. It sees the user
exists, but fails to notice a password was never defined, resulting
in no running of auth_query.

Add a test case to catch this. The case successfully fails before
this patchset and succeeds after it.

Fixes #484
Fixes #706

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -414,7 +414,7 @@ struct PgPool {
 	 *
 	 * A special case is when there are cancel requests waiting to be forwarded
 	 * to servers in waiting_cancel_req_list. In that case the server bails out
-	 * of the login flow, because a cancel reuest needs to be sent before
+	 * of the login flow, because a cancel request needs to be sent before
 	 * logging in.
 	 *
 	 * NOTE: This list can at most contain a single server due to the way
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -497,7 +497,7 @@ struct PgUser {
 	int pool_mode;
 	int max_user_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
-	struct PgUser *cf_user; /* when set, which configured user this dynamic user shadows */
+	struct PgUser *cf_user;	/* when set, which configured user this dynamic user shadows */
 };
 
 /*
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -347,6 +347,11 @@ struct PgPool {
 	 * Clients that sent cancel request, to cancel another client its query.
 	 * These requests are waiting for a new server connection to be opened,
 	 * before the request can be forwarded.
+	 *
+	 * This is a separate list from waiting_client_list, because we want to
+	 * give cancel requests priority over regular clients. The main reason
+	 * for this is, because a cancel request might free up a connection,
+	 * which can be used for one of the waiting clients.
 	 */
 	struct StatList waiting_cancel_req_list;
 
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -493,9 +493,11 @@ struct PgUser {
 	uint8_t scram_ServerKey[32];
 	bool has_scram_keys;		/* true if the above two are valid */
 	bool mock_auth;			/* not a real user, only for mock auth */
+	bool dynamic_passwd;		/* does the password need to be refreshed every use */
 	int pool_mode;
 	int max_user_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
+	struct PgUser *cf_user; /* when set, which configured user this dynamic user shadows */
 };
 
 /*
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -497,7 +497,10 @@ struct PgUser {
 	int pool_mode;
 	int max_user_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
-	struct PgUser *cf_user;	/* when set, which configured user this dynamic user shadows */
+
+	/* cf_user points at the configured user that a user with a dynamic password is shadowing.
+	   For configured users, cf_user points at itself. */
+	struct PgUser *cf_user;
 };
 
 /*
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -503,8 +503,9 @@ struct PgUser {
 	int max_user_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
 
-	/* cf_user points at the configured user that a user with a dynamic password is shadowing.
-	   For configured users, cf_user points at itself. */
+	/* cf_user points at the configured user that a user
+	 * with a dynamic password is shadowing. For configured
+	 * users, cf_user points at itself. */
 	struct PgUser *cf_user;
 };
 
--- a/src/client.c
+++ b/src/client.c
@@ -453,7 +453,7 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 		}
 	} else {
 		client->login_user = find_user(username);
-		if (!client->login_user) {
+		if (!client->login_user || client->login_user->dynamic_passwd) {
 			/*
 			 * If the login user specified by the client
 			 * does not exist, check if an auth_user is
@@ -554,6 +554,7 @@ bool handle_auth_query_response(PgSocket *client, PktHdr *pkt)
 			length = sizeof(user.passwd) - 1;
 		memcpy(user.passwd, password, length);
 
+		slog_debug(client, "successfully parsed auth_query response for user %s", user.name);
 		client->login_user = add_db_user(client->db, user.name, user.passwd);
 		if (!client->login_user) {
 			disconnect_server(server, false, "unable to allocate new user for auth");
--- a/src/client.c
+++ b/src/client.c
@@ -188,6 +188,15 @@ static void start_auth_query(PgSocket *client, const char *username)
 	}
 	client->link->ready = false;
 
+	/*
+	 * Add outstanding request, so that the server is closed if the client
+	 * disconnects before the auth_query completes.
+	 */
+	if (!add_outstanding_request(client, 'S', RA_SKIP)) {
+		disconnect_server(client->link, true, "out of memory");
+		return;
+	}
+
 	res = 0;
 	buf = pktbuf_dynamic(512);
 	if (buf) {
--- a/src/client.c
+++ b/src/client.c
@@ -456,10 +456,10 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 		if (!client->login_user || client->login_user->dynamic_passwd) {
 			/*
 			 * If the login user specified by the client
-			 * does not exist, check if an auth_user is
-			 * set and if so send off an auth_query.  If
-			 * no auth_user is set for the db, see if the
-			 * global auth_user is set and use that.
+			 * does not exist or if it has no entry in auth_file,
+			 * check if an auth_user is set and if so, send off
+			 * an auth_query.  If no auth_user is set for the db,
+			 * see if the global auth_user is set and use that.
 			 */
 			if (!client->db->auth_user && cf_auth_user) {
 				client->db->auth_user = find_user(cf_auth_user);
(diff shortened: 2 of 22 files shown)


benchub:
I'm not really sure what's going on with these tests, and I think it's clear I'm flailing around more than enough on github. However, this really does seem to address an outstanding bug with auth_query and configured users.

src/loader.c
```diff
@@ -577,6 +577,9 @@ static void copy_quoted(char *dst, const char *src, int len)
 	*dst = 0;
 }
 
+/* This function is only called when parsing the auth file, so
+   all users added by this function do not have a dynamic password,
+   by definition. If the password is empty, so be it. */
 static void unquote_add_user(const char *username, const char *password)
```
JelteF, line 583: let's change the name of this function to something like `authfile_add_user`

src/objects.c
```diff
@@ -1291,7 +1317,10 @@ void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...
 
 	server->pool->db->connection_count--;
 	if (server->pool->user) {
-		server->pool->user->connection_count--;
+		if (server->pool->user->cf_user)
+			server->pool->user->cf_user->connection_count--;
+		else
+			server->pool->user->connection_count--;
```
JelteF, line 1323: Having to do such checks everywhere is quite annoying. And seems easy to forget when adding new config options for users in the future. I feel like we should probably split the PgUser object into two objects, something like PgUser and PgUserConfig.

and then we can add a user->config field of type PgUserConfig, which we then unconditionally access.

JelteF (changes requested):
Thanks for your work on this! This is indeed quite an annoying bug. I think overall direction is good, but I think we should go a bit further with the change because I dislike the cf_user checks that now proliferate the codebase.

Also, I just created while reviewing this PR https://github.com/pgbouncer/pgbouncer/pull/1040 I'd appreciate it if you could take a look at that and let me know if you think that change makes sense.

src/loader.c
benchub: done

src/objects.c
benchub: That makes sense. I was looking to minimize the risk of failing to catch a memory cleanup, but I agree this would be a cleaner path.

JelteF: another option I guess would be to have cf_user point to itself for non dynamic users

benchub: Yeah, I started down the road of having a different datatype but current counts needed to be in there too and it just makes more sense to have a cf_user point to itself when needed, so that we can blindly follow it.

benchub: OK, I think I'm done with this as well.

include/bouncer.h
```diff
@@ -498,9 +498,14 @@ struct PgUser {
 	uint8_t scram_ServerKey[32];
 	bool has_scram_keys;		/* true if the above two are valid */
 	bool mock_auth;			/* not a real user, only for mock auth */
+	bool dynamic_passwd;		/* does the password need to be refreshed every use */
 	int pool_mode;
 	int max_user_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
+
+	/* cf_user points at the configured user that a user with a dynamic password is shadowing.
+	   For configured users, cf_user points at itself. */
```
JelteF, line 507: nit: for such multi line comments please use this style
```c
/*
 * some long comment
 * and some more lines
 */
```

(I'll try to figure out how to make uncrustify fix these automatically)

src/client.c
```diff
@@ -453,7 +453,7 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 		}
 	} else {
 		client->login_user = find_user(username);
-		if (!client->login_user) {
+		if (!client->login_user || client->login_user->dynamic_passwd) {
 			/*
 			 * If the login user specified by the client
 			 * does not exist, check if an auth_user is
```
JelteF, line 459:
```suggestion
			 * If the login user specified by the client
			 * does not exist or if it has no entry in the auth_file, check if an auth_user is
```

test/test_auth.py
```diff
@@ -406,6 +413,86 @@ def test_auth_query_database_setting(
                 )
 
 
+@pytest.mark.skipif(
+    "FREEBSD",
+    reason="FreeBSD does not seem to observe the expected OperationalError description",
```
JelteF, line 418: What error does it give instead?

src/client.c
benchub: done

test/test_auth.py
benchub: The cirrus test says "psycopg.OperationalError: consuming input failed: server closed the connection unexpectedly". The bouncer log does seem to correctly show the error though... you know what, let me rework the test to look at the bouncer log for confirmation the right error is happening.

benchub: ...and it looks like the new test works well.

emelsimsek:
The current logic that adds those users to the `auth_file `users (`struct AATree user_tree;`) seems wrong to me.

I believe the following would be a better approach:

The users in the [user] section are not added to the same list with auth_file users. A new structure is used to maintain them, e.g:

	struct AATree cf_user_user_tree;
	
If a user is not found in auth_file users  

	  Run auth_query
		  Create a PgUser from the query results
		  find_user_user_config()  ---> finds user in cf_user_user_tree.
		  Update the PgUser from auth_query results with the user config in cf_users.

benchub:
Sorry @emelsimsek, I'm afraid I'm not seeing what we gain by having two different user trees? My understanding of user_tree is that it is used as the collection of users that pgBouncer knows about. What do we gain by having one set for users whose password comes from auth_query and another set where the password comes from auth_file? Wouldn't we need to check both trees in every place where we are currently checking just one?

emelsimsek:
> I'm afraid I'm not seeing what we gain by having two different user trees?

@benchub My thinking goes like this: The [Users ](https://www.pgbouncer.org/config.html#section-users:~:text=timezone%20from%20server.-,Section%20%5Busers%5D,-This%20section%20contains)section is for overriding user configurations for users that comes from various sources. The users and their authentication tokens could be defined in either auth_file or remote user directories  such as LDAP. Therefore for flexibility reasons, the user configurations should be maintained in a different list then auth_file's list.

benchub:
OK @emelsimsek, I've spent some time trying to understand more of the pgBouncer codebase and I'm afraid I don't understand how to implement what you're suggesting, or even understand enough to see how it would help. 

I assume when you say, "remote user directories such as LDAP" you are talking about PAM authentication and using LDAP as a PAM provider. I can appreciate that having user credentials being possibly defined in 3 places (pam, auth_file, or auth_query) could get messy, but the documentation is clear that pam is incompatible with the other two options. And indeed, when I look at the code base, there are multiple times that pgBouncer says "Oh, in the case of PAM, do this all differently" - for example, client.c:set_pool(), admin.c:show_one_fd(), or objects.c:use_server_socket().

It's a fair question to ask if pgBouncer _should_ do it all differently, or if it should, as you seem to be suggesting, be able to use auth_query, auth_file, and pam all at the same time. I would suggest that such a change in behavior is beyond the scope of this bug fix.

More importantly, I don't know how to test pgBouncer using PAM. I don't see any unit tests doing so to work from, I don't have PAM set up in any of my environments, and frankly, while I remember PAM coming to linux almost 30 years ago, I've never set it up in all that time. I'm sure somebody is using it, but it's not me. I'm uncomfortable proposing code I don't have the ability to test.

Less importantly, in many of the places that the (currently only) user_tree is queried, it looks like having additional trees would seem to complicate the code for little gain. For example, objects.c:user_client_socket(), admin.c:admin_setup(), or admin.c:show_one_fd().

emelsimsek:
> OK @emelsimsek, I've spent some time trying to understand more of the pgBouncer codebase and I'm afraid I don't understand how to implement what you're suggesting, or even understand enough to see how it would help.
> 
> I assume when you say, "remote user directories such as LDAP" you are talking about PAM authentication and using LDAP as a PAM provider.
Yes, the users may be defined in an LDAP directory and this is integrated into pgbouncer using PAM libraries.

> I can appreciate that having user credentials being possibly defined in 3 places (pam, auth_file, or auth_query) could get messy, but the documentation is clear that pam is incompatible with the other two options. And indeed, when I look at the code base, there are multiple times that pgBouncer says "Oh, in the case of PAM, do this all differently" - for example, client.c:set_pool(), admin.c:show_one_fd(), or objects.c:use_server_socket().

auth_file and LDAP are alternatives to each other. users section in the configuration file on the other hand looks like intended to overrride user configurations for particular users.
I am also not sure about the maturity/extent of LDAP support in pgbouncer.
> 
> It's a fair question to ask if pgBouncer _should_ do it all differently, or if it should, as you seem to be suggesting, be able to use auth_query, auth_file, and pam all at the same time. I would suggest that such a change in behavior is beyond the scope of this bug fix.

I agree that for getting this functionality work in its simplest form, the code change can be kept limited as you had done.
Especially given the absence of test automation for PAM feature, it may be best to keep it as it is.

> 
> More importantly, I don't know how to test pgBouncer using PAM. I don't see any unit tests doing so to work from, I don't have PAM set up in any of my environments, and frankly, while I remember PAM coming to linux almost 30 years ago, I've never set it up in all that time. I'm sure somebody is using it, but it's not me. I'm uncomfortable proposing code I don't have the ability to test.
> 
> Less importantly, in many of the places that the (currently only) user_tree is queried, it looks like having additional trees would seem to complicate the code for little gain. For example, objects.c:user_client_socket(), admin.c:admin_setup(), or admin.c:show_one_fd().

benchub:
@JelteF are you waiting on me for anything?

goksgie:
I feel like `[users]` section should be specific to configurations, rather than mixing it with authentication. Thus, @emelsimsek 's suggestion of utilizing different data structures for both makes more sense, as then you would have a clear separation of "configurations" and "authentication". And, when you have a user that is attempting to connect to the server, you can pull in the configuration settings from the separate data structure if not initialized already and apply them accordingly.

As a side note, adding self referencing pointer and additional boolean to make this work feels hacky to me. But that is of course up to @JelteF  and @emelsimsek to decide.

JelteF:
The more I look into this the more I realize how big of a mess our user management currently is. I definitely think we need some distinction between different types of users. It's super confusing to reason about. I'm trying out a few options for this at the moment.

JelteF:
I created a PR for the auth refactor here: https://github.com/pgbouncer/pgbouncer/pull/1052

PR #1040, opened 2024-03-13
Change SHOW USERS to list connections

JelteF:
While reviewing #1039 I had a hard time seeing if
max_user_connections was applied correctly because `SHOW USERS` didn't
list it. This addresses that by adding a `max_user_connections` and
`current_connections` column to the output of `SHOW USERS`.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -506,6 +506,13 @@ name
 pool_mode
 :   The user's override pool_mode, or NULL if the default will be used instead.
 
+max_user_connections
+:   The user's max_user_connections setting. If this setting is not set
+    for this specific user, then the default value will be displayed.
+
+current_connections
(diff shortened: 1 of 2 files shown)


doc/usage.md
```diff
@@ -506,6 +506,13 @@ name
 pool_mode
 :   The user's override pool_mode, or NULL if the default will be used instead.
 
+max_user_connections
+:   The user's max_user_connections setting. If this setting is not set
+    for this specific user, then the default value will be displayed.
+
+current_connections
+:   Current number of connections that this user has open to the server.
```
benchub, line 514: Instead of "has open to the server", I think it would be more useful (and more accurate, given your code) to be "has open to all servers."

benchub (commented):
Mostly good.

PR #1042, opened 2024-03-15
Update psycopg to an actual release

JelteF:
This starts pulling in a released version of psycopg instead of the version
from GitHub. The support for Close packet has been released, so it's not
necessary to pull from GitHub anymore. 

This fixes a bunch of issues on these official psycopg releases. It also fixes
an issue that I ran into where `test_equivalent_startup_param` failed on my new
laptop because initdb would use 'iso, dmy' as datestyle instead of 'iso, mdy'.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,6 @@ pytest
 pytest-asyncio
 pytest-timeout
 pytest-xdist
-psycopg; python_version < "3.7"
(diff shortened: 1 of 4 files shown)


JelteF:
Closing this as I failed to realise the feature wasn't actually released yet...

PR #1048, opened 2024-04-12
Add values in ignore_startup_parameters

marjuscako:
Currently you can only define the key of a parameter in the `ignore_startup_parameters` section. With this change you can define a `key=value` string in that section. For example you can set the following:

```
[pgbouncer]
ignore_startup_parameters = options,extra_float_digits=2
```

If an application attempts to connect using as startup parameter the extra_float_digits=3, then pgBouncer will produce an error.

This is useful to avoid hidden bugs when using this section between application logic changes. Also it can function for clients as an indication of the expected startup parameters.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -236,11 +236,16 @@ Default: IntervalStyle
 
 By default, PgBouncer allows only parameters it can keep track of in startup
 packets: `client_encoding`, `datestyle`, `timezone` and `standard_conforming_strings`.
-All others parameters will raise an error.  To allow others parameters, they can be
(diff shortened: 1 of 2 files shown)


JelteF:
Thank you for the contribution this sounds like a useful feature indeed. I haven't looked closely at the code yet. But some initial feedback:
1. A test for this should be included
2. There is a memory leak reported by valgrind
3. There are some code formatting errors. You can use `make format` to fix those

JelteF:
@marjus45 are you still planning to update this PR?

PR #1052, opened 2024-04-18
Refactor user management

### LINKED ISSUE #484: pgbouncer didn't even try auth_query to verify user password

JelteF:
Based upon #1039, but changed a lot by splitting the `PgUser` type into two different types. First we introduce a `PgGlobalUser` type which stores user settings from the ini file, credentials in `auth_file` and connection count tracking. And secondly we introduce `PgCredentials` which only stores user credentials, these could either come from the `auth_file` or `auth_query`. This type is embedded in the `PgGlobalUser` to store any global credentials from `auth_file`, and it also links back to the matching `PgGlobalUser`.

This makes user configuration work for dynamically created users and will also list dynamically created users in the output of `SHOW USERS`.

Fixes #484
Fixes #706
Resolves #1039


include/bouncer.h
```diff
@@ -326,7 +327,7 @@ struct PgPool {
 	struct List map_head;			/* entry in user->pool_list */
 
 	PgDatabase *db;			/* corresponding database */
-	PgUser *user;			/* user logged in as, this field is NULL for peer pools */
+	PgAuthInfo *user;			/* user logged in as, this field is NULL for peer pools */
```
benchub, line 330: Given GlobalUser as a type, I am tempted to suggest this field should be called auth_info or credentials instead of user, to reduce confusion. (And yes, this implies a fair amount of substitutions elsewhere.)

In general, how do you feel about never using "user" for pgAuthInfo data, and keeping that reserved for GlobalUser types?

src/client.c
```diff
@@ -317,10 +317,10 @@ static bool finish_set_pool(PgSocket *client, bool takeover)
 	struct HBARule *rule = NULL;
 
 	if (!client->login_user->mock_auth && !client->db->fake) {
-		PgUser *pool_user;
+		PgAuthInfo *pool_user;
```
benchub, line 320: Here, for example - I would think pool_user could have a better name given its new type to reduce confusion.

benchub (commented):
I think it's a fine refactor. I do think there's more work to do if this is the direction - a lot of field names that used to be of type PgUser are still named some permutation of "user", even though they are now storing credential info instead of the new GlobalUser type.

include/bouncer.h
JelteF: Yeah, it's definitely something I think we should be consistent about and was originally trying to do what you are suggesting. But when I got to the `auth_user` field, I didn't really like the resulting `auth_user_auth_info` name.

So I held off on changing everything, because maybe it fits better to rename `PgAuthInfo` to `PgUser`

benchub: What about s/PgAuthInfo/PgCredentials/, so that you end up with `auth_user_credentials` or the like?

JelteF: I like that suggestion. I'll try that out tomorrow.

include/bouncer.h
```diff
@@ -483,13 +484,12 @@ struct PgPool {
  * FIXME: remove ->head as ->tree_node should be enough.
```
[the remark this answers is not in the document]
JelteF: Fixed

include/bouncer.h
JelteF: Okay, I finally found the time to do this. Could you take another look? I'm pretty happy with the result.

benchub (commented):
It's a lot of changes, but it looks good enough at this point that I unleashed my inner grammar nazi for some nits. Assuming the unit tests pass I think it's good enough to get it out there.

clarkbreyman:
> Based upon #1039, but changed a lot by splitting the `PgUser` type into two different types. First we introduce a `PgGlobalUser` type which stores user settings from the ini file, credentials in `auth_file` and connection count tracking. And secondly we introduce `PgCredentials` which only stores user credentials, these could either come from the `auth_file` or `auth_query`. This type is embedded in the `PgGlobalUser` to store any global credentials from `auth_file`, and it also links back to the matching `PgGlobalUser`.
> 
> This makes user configuration work for dynamically created users and will also list dynamically created users in the output of `SHOW USERS`.
> 
> Fixes #484 Fixes #706 Resolves #1039

@JelteF - There are auth modes (e.g. PAM) that pass along client credentials, not  than `auth_file` or `auth_query`.  I'm wondering if there are any issues introduced by the assumption that credentials are static?

Example use case:  passing a dynamic JWT as the client credential. This is how Microsoft Entra (formerly Azure Active Directory) does authentication for Azure Postgres.  A PAM module can work in tandem with PgBouncer to ensure that the client credential is valid even when dynamic.

PR #1057, opened 2024-04-27
Add `server_lifetime` to databases config

livinlefevreloca:
This PR makes it possible to specify `server_lifetime` on a per database basis in the `[databases]` section. 

I have a personal use case for this where our application has a pgbouncer instance that pools connections for several different severs, one of which makes heavy use of partitioning. The partitioning causes connections to become very memory heavy overtime and it would great if we could cycle these connections more often to refresh them. We would prefer not to set the global setting though to avoid the overhead re authenticating to all the servers when for some of them it is not necessary 

It would be great to get this merged into the main trunk but of course I understand it might not be a wide enough use case.

This would be my first contribution to pgbouncer so please let me know if there any issues with the implementation or anything you would like done differently and I will happily make those changes

Thanks,

Adam


JelteF (approved):
I think this change is reasonable. The usecase makes sense, and overall I think it would be good if we could specify most of our global at the database level too.

Overall the code looks good. The only thing I think is necessary to merge this PR is that this new setting should also be listed in `SHOW DATABASES` (and the docs for that command should be updated to include it then too).

JelteF (changes requested):
ugh, wrong button. Its needs the `SHOW DATABASES` change.

livinlefevreloca:
@JelteF I have made the updates to admin to display the `server_lifetime` in the `SHOW DATABASES` command and updated the documentation. It currently displays as seconds but as a unit-less number. I did wonder if it might make sense to change it to display as string like `20 seconds` or to have the column name be `server_lifetime (seconds)` . Let me know if either sounds good to your or if its fine as is.

I also had to make some additional updates to my original implementation specifically around the use of `int` for the type of `sever_lifetime` in the `PgDatabase` struct. This was incorrect due to the value being stored as microseconds which means any setting of over 2147 seconds would have overflowed. The type has been updated to `usec_t` which will handle the microseconds correctly

JelteF (approved):

PR #1058, opened 2024-04-29
Correctly handle large and partial startup packets

### LINKED ISSUE #1056: Fail test test_misc.py::test_options_startup_param on ubuntu-24.04
When running tests on ubuntu-24.04, sometimes the test_misc.py::test_options_startup_param test fails with the error "FATAL: client sent partial pkt in startup phase"
reproduce:
```
a=`printf 'a%.0s' {1..1000}`; \
for i in `seq 100`; do \
echo "iter $i"; \
sudo -u postgres psql -d "host=127.0.0.1 port=15432 client_encoding=UTF8 connect_timeout=10 dbname=postgres options='-c timezone=test'" -c "select 1" &> /dev/null;  \
sudo -u postgres psql -d "host=127.0.0.1 port=15432 client_encoding=UTF8 connect_timeout=10 dbname=postgres options='-c timezone=$a'" &> /tmp/logmsg; \
grep "partial pkt" /tmp/logmsg && { echo "FOUND"; break; }; \
sleep 0.1; \
done
```
If you increase tcp_socket_buffer from 4096 to 8192, the error is not reproduced.

bouncer.ini
```
[databases]
postgres = port=5432 host=127.0.0.1 dbname=postgres user=postgres pool_size=2

[pgbouncer]
logfile = test.log
pidfile = test.pid

listen_addr = 127.0.0.1
listen_port = 15432
unix_socket_dir = /tmp

auth_type = trust
auth_file = userlist.txt
pool_mode = statement

; use a small tcp_socket_buffer so that we trigger EAGAIN more often in tests.
; Otherwise bugs caused by EAGAIN are hard to debug.
tcp_socket_buffer = 4096

```

JelteF:
We were bailing out from a connection if we did not receive the full
startup packet in one go, or if it didn't fit in pkt_buf. This fixes
these issues by reusing the packet buffering logic that was introduced
for prepared statement support for our startup packet parsing.

Fixes #1056

### DIFF
--- a/include/proto.h
+++ b/include/proto.h
@@ -80,10 +80,22 @@ static inline bool incomplete_header(const struct MBuf *data)
 	return data->data[data->read_pos] == 0;
 }
 
-/* rewind the body of a v3 packet. Does not work for v2 packets, e.g. startup packets */
+/*
+ * rewind the body of a v3 packet. Does not work for v2 packets, e.g. startup
+ * packets
+ */
 static inline void pkt_rewind_v3(PktHdr *pkt)
 {
 	pkt->data.read_pos = NEW_HEADER_LEN;
+}
+
+/*
+ * rewind the body of a v2 packet. Does not work for v3 packets, i.e.
+ * everything except a startup packet
+ */
+static inline void pkt_rewind_v2(PktHdr *pkt)
+{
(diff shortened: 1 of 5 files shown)


kav23alex:
Thanks for correction. I checked, this problem is no longer reproduced.

PR #1064, opened 2024-05-07
Support --config=value format in options startup parameter

### LINKED ISSUE #952: ignore_startup_options=options has no effect
Hi, i trying to get pgbouncer work with some software that  send following

image

And nothing set in ignore_startup_parameters=options,client_encoding,standard_conforming_strings will get around it.

Looking at code you try to find first -c option=value and then look if options is set in ignore_startup_parameters

_Originally posted by @torstenfohrer in https://github.com/pgbouncer/pgbouncer/issues/908#issuecomment-1738701265_

JelteF:
The `-c config=value` pattern has been supported in PgBouncer since
PR #878. As described in #952 Postgres also allows to use the
`--config=value`, which honestly seems more readable, but for some
reason is mentioned almost nowhere in the Postgres documentation.

Since #878 this always throws an error, and putting `options` in
`ignore_startup_parameters` wouldn't get rid of this error. So this
should actually be considered a regression for the same reason as #907
was considered a regression (although a much less impactful one given
we've only received two reports of this causing issues).

Fixes #952

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -711,10 +711,14 @@ static bool set_startup_options(PgSocket *client, const char *options)
 		char *equals;
 		mbuf_rewind_writer(&arg);
 		position = cstr_skip_ws((char *) position);
-		if (strncmp("-c", position, 2) != 0)
+		if (strncmp("-c", position, 2) == 0) {
+			position += 2;
+			position = cstr_skip_ws((char *) position);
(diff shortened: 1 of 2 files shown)


emelsimsek (approved):
Testing this locally, and reviewing the change it looks fine to me.

PR #1075, opened 2024-05-18
Reactivate some Cirrus CI tasks with manual trigger

petere:
Reactive some tasks that were commented out in 92b4645.  Instead, set them to trigger manually.  That way, they can still be used individually when desired.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -100,63 +100,65 @@ task:
       path: "config.log"
       type: text/plain
 
-# task:
-#   name: Linux (Red Hat)
-#   container:
-#     matrix:
-#       - image: rockylinux:9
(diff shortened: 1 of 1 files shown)


JelteF:
Good idea, would be good to update the [Making a release](https://github.com/pgbouncer/pgbouncer/wiki/Making-a-release) page of the wiki to mention that we should enable all those builds. Or probably even better, trigger them automatically when the origin branch is of the form `draft-release*` (and mention in the wiki that you should call the branch that way).

petere:
I merged at as is and updated the release checklist.

I don't think what you are describing is easily possible. The expression language on Cirrus CI is not that general. Maybe someone wants to try it as a separate undertaking.

PR #1077, opened 2024-05-20
Meson build system

petere:
I wrote a Meson build system for PgBouncer. Just like for PostgreSQL, I think this could be more pleasant to use and easier to maintain in the long run. Right now, it's still missing a few pieces of functionality, and it's held together with a few hacks. But it does work (`meson setup build`, `meson compile -C build`, `meson test -C build`, `meson install -C build`). Let me know what you think.

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -1,0 +1,26 @@
+hba_test = executable('hba_test',
+                      'hba_test.c',
+                      '../src/hba.c',
+                      '../src/util.c',
+                      libusual_sources,
+                      config_h,
+                      build_by_default: false,
+                      include_directories: ['..', '../include', '../lib'],
+                      dependencies: [libevent, openssl, systemd],
+                     )
+
+libpq = dependency('libpq', required: false, disabler: true)
+
+asynctest = executable('asynctest',
+                       'asynctest.c',
+                       libusual_sources,
+                       config_h,
+                       build_by_default: false,
+                       include_directories: ['..', '../include', '../lib'],
+                       dependencies: [libevent, libpq, systemd],
+                      )
+
+test('hba_test',
+     hba_test,
+     workdir: meson.project_source_root() / 'test',
+    )
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,11 +12,11 @@ task:
     - container:
         image: ubuntu:22.04
       env:
-        configure_args: '--with-cares --with-pam --with-ldap'
+        configure_args: '-Dcares=enabled -Dpam=enabled -Dldap=enabled'
     - container:
         image: ubuntu:22.04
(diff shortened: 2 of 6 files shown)


test/meson.build
```diff
@@ -0,0 +1,24 @@
+hba_test = executable('hba_test',
+                      'hba_test.c',
+                      '../src/hba.c',
+                      '../src/util.c',
+                      libusual_sources,
+                      config_h,
+                      include_directories: ['..', '../include', '../lib', '../uthash/src'],
+                      dependencies: [libevent, openssl],
```
JelteF, line 8: If systemd is available then libusual fails to link without this, due to logging.c calling `sd_journal_print`

```suggestion
                      dependencies: [libevent, openssl, systemd],
```

test/meson.build
```diff
@@ -0,0 +1,24 @@
...
+                     )
+
+libpq = dependency('libpq', required: false, disabler: true)
+
+asynctest = executable('asynctest',
+                       'asynctest.c',
+                       libusual_sources,
+                       config_h,
+                       include_directories: ['..', '../include', '../lib', '../uthash/src'],
+                       dependencies: [libevent, libpq],
```
JelteF, line 18:
```suggestion
                       dependencies: [libevent, libpq, systemd],
```

JelteF:
Overall, great idea! Autoconf is objectively horrible. A few thoughts:

1. Instead of adding meson support for libusual to this repo, maybe we should add that to libusual
2. I think we should throw away autoconf support completely as part of this change. Maintaining two build systems is needed for Postgres because of extensions, but PgBouncer doesn't have this requirement.

rpufky:
Looking into this build system replacement. Are there any thoughts of introducing Visual Studio builds as part of the Windows builds?

I'm willing to contribute to PR, just looking to see if there is anything started here.

petere:
> Looking into this build system replacement. Are there any thoughts of introducing Visual Studio builds as part of the Windows builds?
> 
> I'm willing to contribute to PR, just looking to see if there is anything started here.

Once we have the Meson-based build system, then it would be feasible to consider Visual Studio support. But I'm not going to work on it as part of this PR, and I suggest that we'd keep it a separate follow-up patch.

You could, in the meantime, create your own branch off my branch and then start making VS work.

rpufky:
> Once we have the Meson-based build system, then it would be feasible to consider Visual Studio support. But I'm not going to work on it as part of this PR, and I suggest that we'd keep it a separate follow-up patch.
> 
> You could, in the meantime, create your own branch off my branch and then start making VS work.

Will do. I'll provide support to this PR first, then once we're set with meson build system work on a pr based off of this pr to add visual studio.

PR #1078, opened 2024-05-23
Add support for "all" address type in HBA file

tie:
This change adds support for "all" address type in HBA file that matches both 0.0.0.0/0 and ::/0, similar to PostgreSQL HBA.

https://postgresql.org/docs/16/auth-pg-hba-conf.html#:~:text=You%20can%20also%20write%20all%20to%20match%20any%20IP%20address


JelteF (approved):

JelteF:
Interesting, I honestly didn't even know that this keyword existed in the address column. Totally makes sense to support though. Thanks for the contribution.

Issue #1098, opened 2024-07-02
Memory leaks in parse_database


dmitry-lipetsk:
Hello,

I found few problems with in parse_database function.

My pgbouncer.ini file contains the following data:
```
[databases]
template1 = host=localhost host=localhost2 port=5432 dbname=template1
```
Note that 'host' parameter is defined twice.

Debugger shows that memory for 'host' is allocated twice, too. And the second pointer overrides the first one.

https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/loader.c#L305-L310

The same problem with 'connect_query' parameter:
https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/loader.c#L346-L350

---
The last problem - if we allocate memory for 'host' but cannot allocate memory for 'connect_query' (or vice versa), we do not release an allocated block in 'fail' section.

---
As I understand, parse_database is called each time when we execute "RELOAD" command through an admin connection.

dmitry-lipetsk:
I think, the same problem exists in parse_peer function:

https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/loader.c#L200-L205

1. 'host' can be assigned twice
2. 'host' is not released in 'fail' section

eulerto:
I would suggest you to use `set_param_value` function. See `auth_dbname` or `auth_query` in the same function. It encapsulates the logic of "free old value and allocates the new value".

PR #1099, opened 2024-07-03
Fix for #1098 [memory leaks]

dmitry-lipetsk:
This PR fixes memory leaks in parse_database and parse_peer functions.

See #1098 for details.

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -464,9 +464,6 @@ bool parse_database(void *base, const char *name, const char *connstr)
 	/* remember dbname */
 	db->dbname = (char *)msg->buf + dbname_ofs;
 
-	Assert(host == NULL);
-	Assert(connect_query == NULL);
-
 	free(tmp_connstr);
 	return true;
 fail:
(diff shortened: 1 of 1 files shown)


src/loader.c
```diff
@@ -233,13 +230,17 @@ bool parse_peer(void *base, const char *name, const char *connstr)
 
 	free(peer->host);
 	peer->host = host;
+	host = NULL;
 	peer->port = port;
 	peer->pool_size = pool_size;
 
+	Assert(host == NULL);
+
```
JelteF, line 238: These additions seem a bit unnecessary.
```suggestion
	peer->port = port;
	peer->pool_size = pool_size;

```

src/loader.c
```diff
@@ -470,10 +467,15 @@ bool parse_database(void *base, const char *name, const char *connstr)
 	/* remember dbname */
 	db->dbname = (char *)msg->buf + dbname_ofs;
 
+	Assert(host == NULL);
+	Assert(connect_query == NULL);
+
```
JelteF, line 472: these additions also seem sunnecessary.

JelteF (changes requested):
The setting to NULL in the happy path and then the later Assert(NULL) calls are a bit unnecessary. Other than that this looks good.

JelteF (approved):

Issue #1100, opened 2024-07-03
Different problems in parse_ident_line (hba.c)


dmitry-lipetsk:
Hello,

I found that parse_ident_line (hba.c) does not check strdup results and may exit by error without cleanup.

https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/hba.c#L569

Here it does not check map_name_copy:
https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/hba.c#L595-L597

Here it does not check mapping->system_user_name:
https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/hba.c#L604-L606

Here it does not check mapping->postgres_user_name:
https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/hba.c#L615-L617

Here it exits by error without releasing an allocated 'mapping':
https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/hba.c#L631-L636

Section 'failed' must free map_name_copy, too:
https://github.com/pgbouncer/pgbouncer/blob/5e01e56328201ce8272e4421b550874162508a70/src/hba.c#L646-L650

eulerto:
@emelsimsek it seems you forgot to edit the commit message saying that the PR #1101 closes this issue.

@dmitry-lipetsk some of your contributions already have a patch so there is no need to create an issue and after that a PR, you can create only the PR. If you are fixing an issue that was reported some time ago, you should say so using a [GH keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword).

PR #1101, opened 2024-07-03
Fix for #1100 [parse_ident_line (hba.c)]

dmitry-lipetsk:
This PR fixes different problems in function parse_ident_line (hba.c).

1. This function did not check strdup results
2. This function could return by error without cleanup
3. This function did not free map_name_copy in 'failed' section

Additional DEBUG checks:
1. 'failed' section cannot get a not null value of ident_map. I suggest to remove "ident_map_free(ident_map);" from there.

### DIFF
--- a/src/hba.c
+++ b/src/hba.c
@@ -658,9 +658,6 @@ static bool parse_ident_line(struct Ident *ident, struct TokParser *tp, int line
 	return true;
 
 failed:
-	/* Not null ident_map is not expected here. */
-	Assert(ident_map == NULL);
-
 	mapping_free(mapping);
 	ident_map_free(ident_map);
 	free(map_name_copy);
--- a/src/hba.c
+++ b/src/hba.c
@@ -639,7 +639,7 @@ static bool parse_ident_line(struct Ident *ident, struct TokParser *tp, int line
 	if (find_ident_map(ident, map_name_copy, &ident_map)) {
 		list_append(&ident_map->mappings, &mapping->node);
 		free(map_name_copy);
-		/* map_name_copy = NULL; */
+		map_name_copy = NULL;
 	} else {
 		ident_map = calloc(1, sizeof(*ident_map));
 
@@ -649,7 +649,7 @@ static bool parse_ident_line(struct Ident *ident, struct TokParser *tp, int line
 		}
 
 		ident_map->map_name = map_name_copy;
-		/* map_name_copy = NULL; */
+		map_name_copy = NULL;
 		list_init(&ident_map->mappings);
 		list_append(&ident_map->mappings, &mapping->node);
 		list_append(&ident->maps, &ident_map->node);
--- a/src/hba.c
+++ b/src/hba.c
@@ -593,6 +593,10 @@ static bool parse_ident_line(struct Ident *ident, struct TokParser *tp, int line
 	}
 
 	map_name_copy = strdup(map_name);
+	if (!map_name_copy) {
+		log_warning("ident: no mem for map_name");
+		goto failed;
+	}
 
 	next_token(tp);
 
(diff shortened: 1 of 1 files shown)


src/hba.c
```diff
@@ -644,8 +658,12 @@ static bool parse_ident_line(struct Ident *ident, struct TokParser *tp, int line
 	return true;
 
 failed:
+	/* Not null ident_map is not expected here. */
+	Assert(ident_map == NULL);
```
emelsimsek, line 662: This assert should be removed.  The `failed` block may run when unexpected tokens are encountered in addition to mem allocation failures. 

See

https://github.com/pgbouncer/pgbouncer/pull/1101/files#diff-c0d4b6bb6b74133fcf98cf505edbb8acac315acaa1a885f863b168771acb37efR592:~:text=%26map_name))%20%7B-,if%20(!expect(tp%2C%20TOK_IDENT%2C%20%26map_name))%20%7B,%7D,-map_name_copy%20%3D%20strdup

src/hba.c
```diff
@@ -627,15 +639,17 @@ static bool parse_ident_line(struct Ident *ident, struct TokParser *tp, int line
...
 		ident_map = calloc(1, sizeof(*ident_map));
 
 		if (!ident_map) {
 			log_warning("ident: no mem for parsing ident_map");
-			return false;
+			goto failed;
 		}
 
 		ident_map->map_name = map_name_copy;
+		/* map_name_copy = NULL; */
```
emelsimsek, line 652: Why is this line? `map_name_copy` is definitely not NULL here.

emelsimsek (changes requested):

emelsimsek:
@dmitry-lipetsk thank you for catching the strdup issues here.

src/hba.c
dmitry-lipetsk: > Why is this line? `map_name_copy` is definitely not NULL here.

I general we have to reset (set to NULL) map_name_copy after free (line 641) and assigning in ident_map->map_name (line 651) but it is not necessarily because 'failed' section not be reached from these points.

src/hba.c
dmitry-lipetsk: I am sorry for this question - where ident_map is modified in this point?

I see only ` if (!expect(tp, TOK_IDENT, &map_name))`

emelsimsek: Yes, you are right it is not modified.  But still Assert seems unnecessary to me.  I guess, it demonstrates that no  
`free(ident_map)` is needed. But it serves no other purpose.

emelsimsek (changes requested):

src/hba.c
dmitry-lipetsk: I can save Assert and remove free(ident_map).

free(ident_map) is not required here and more over - it is incorrected.

src/hba.c
```diff
@@ -627,15 +639,17 @@ static bool parse_ident_line(struct Ident *ident, struct TokParser *tp, int line
 	if (find_ident_map(ident, map_name_copy, &ident_map)) {
 		list_append(&ident_map->mappings, &mapping->node);
 		free(map_name_copy);
+		/* map_name_copy = NULL; */
```
emelsimsek, line 642: It looks ok that this can be uncommented.  Although setting to `NULL ` after `free ` is not the convention throughout the code base, this case is somewhat special as the `free `is called in mid function.

src/hba.c
emelsimsek: Ok, Lets uncomment this too.

emelsimsek (changes requested):

emelsimsek (approved):

Issue #1104, opened 2024-07-06
PgCredentials objects are freed incorrectly [through user_cache, not credentials_cache]


dmitry-lipetsk:
Hello,

kill_databases uses slab_free(user_cache, db->forced_user_credentials):
https://github.com/pgbouncer/pgbouncer/blob/bc6bf4fdd0777f283fa9357e999765abf0f2f587/src/janitor.c#L846-L847

but force_user_credentials uses slab_alloc(credentials_cache):
https://github.com/pgbouncer/pgbouncer/blob/bc6bf4fdd0777f283fa9357e999765abf0f2f587/src/objects.c#L596

Is it ok?

dmitry-lipetsk:
Hello,

As I understand it is not OK.

A similar problem exists in credentials_node_release

https://github.com/pgbouncer/pgbouncer/blob/bc6bf4fdd0777f283fa9357e999765abf0f2f587/src/objects.c#L138-L142

This function is used to free db->user_tree:
https://github.com/pgbouncer/pgbouncer/blob/bc6bf4fdd0777f283fa9357e999765abf0f2f587/src/janitor.c#L861

db->user_tree is filed in add_dynamic_credentials:
https://github.com/pgbouncer/pgbouncer/blob/bc6bf4fdd0777f283fa9357e999765abf0f2f587/src/objects.c#L543-L554

So credentials_node_release must use:
`slab_free(credentials_cache, user);`

credentials_node_release is used only for PgDatabase::user_tree.

It seems to me, it is the last point where PgCredentials is freed incorrectly.

dmitry-lipetsk:
This issue leads to the following crash within pgbouncer:

![image](https://github.com/pgbouncer/pgbouncer/assets/4594554/dd90dbd7-9b73-41f9-a3ce-3ce113a00369)

eulerto:
PR was merged.

PR #1105, opened 2024-07-06
Fix for #1104 [free PgCredentials]

dmitry-lipetsk:
After the refactoring of the user managament in #1052, we incorrectly freed credentials by calling `slab_free` on a user_cache, instead of `credentials_cache` in two places. This could later cause segfaults because ththe user type is larger than the credentials type, so out of bound reads would be possible when the credential that was put into the user_cache was taken out again and used as a user instead.

### DIFF
--- a/include/objects.h
+++ b/include/objects.h
@@ -31,6 +31,7 @@ extern struct Slab *peer_cache;
 extern struct Slab *peer_pool_cache;
 extern struct Slab *pool_cache;
 extern struct Slab *user_cache;
+extern struct Slab *credentials_cache;
 extern struct Slab *iobuf_cache;
 extern struct Slab *outstanding_request_cache;
 extern struct Slab *var_list_cache;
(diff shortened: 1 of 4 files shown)


JelteF:
Oooff, that's not good... Thanks for this fix. Any chance you can add a test that triggers the reported crash?

@emelsimsek @eulerto I think this is probably worth a patch release after this is merged. Let's not merge any other (non-critical) PRs so we can create that patch release directly from the master brach.

JelteF:
Ah I see now that you added the test in #1114, could you move that test here?

dmitry-lipetsk:
FAILED test/test_misc.py::test_qa_gh1104 - Failed: **Timeout >30.0s**

dmitry-lipetsk:
Hello, I have a little problem with configuration of development environment on Windows to research and fix a problem in test_ga_gh1104.

So, any help with it will be appreciative.

dmitry-lipetsk:
Done.

Issue #1111, opened 2024-07-08
add_global_user and add_global_credentials can return NULL


dmitry-lipetsk:
Hello,

add_global_user can return NULL:

https://github.com/pgbouncer/pgbouncer/blob/e0937fd5f65a2bdb213a53a4f042b5fc7e786f80/src/objects.c#L492-L499

But these functions do not check it:

- add_dynamic_credentials
- add_pam_credentials
- force_user_credentials

I think, it is incorrected...

add_global_credentials checks the result of add_global_user
https://github.com/pgbouncer/pgbouncer/blob/e0937fd5f65a2bdb213a53a4f042b5fc7e786f80/src/objects.c#L518-L524

But parse_database do not check result of add_global_credentials.

https://github.com/pgbouncer/pgbouncer/blob/e0937fd5f65a2bdb213a53a4f042b5fc7e786f80/src/loader.c#L452-L459

Perhaps, it is not good, too.

PR #1112, opened 2024-07-08
Fix for #1111

dmitry-lipetsk:
Hello,

It is a fix for #1111.

There are added a pair of new functions:
- find_exist_or_add_new_global_user
- find_exist_or_add_new_global_credentials

They replaces calls of:
- find_global_user + add_global_user
- find_global_credentials + add_global_credentials

We checks the result of these new functions to detect OOM.

Also, these functions will be a little fast, because old add_xxx functions call find_xxx again.

Note: add_global_credentials is not used in code and can be removed, I can do it if you want.


include/objects.h
```diff
@@ -71,6 +71,9 @@ bool pop_outstanding_request(PgSocket *client, char *types, bool *skip);
 bool clear_outstanding_requests_until(PgSocket *server, char *types) _MUSTCHECK;
 bool queue_fake_response(PgSocket *client, char request_type) _MUSTCHECK;
 
+PgGlobalUser * find_exist_or_add_new_global_user(const char *name, const char *passwd) _MUSTCHECK;
+PgCredentials * find_exist_or_add_new_global_credentials(const char *name, const char *passwd) _MUSTCHECK;
```
JelteF, line 75: nit let's name these functions as follows:

```suggestion
PgGlobalUser * find_or_add_new_global_user(const char *name, const char *passwd) _MUSTCHECK;
PgCredentials * find_or_add_new_global_credentials(const char *name, const char *passwd) _MUSTCHECK;
```

src/admin.c
```diff
@@ -1746,13 +1746,10 @@ void admin_setup(void)
 		die("cannot create admin pool?");
 	admin_pool = pool;
 
-	/* user */
-	user = find_global_user("pgbouncer");
+	/* find an exist user or create a new fake user with disabled psw */
```
JelteF, line 1749:
```suggestion
	/* find an existing user or create a new fake user with disabled password */
```

src/client.c
```diff
@@ -485,9 +485,7 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 			 * see if the global auth_user is set and use that.
 			 */
 			if (!client->db->auth_user_credentials && cf_auth_user) {
-				client->db->auth_user_credentials = find_global_credentials(cf_auth_user);
-				if (!client->db->auth_user_credentials)
-					client->db->auth_user_credentials = add_global_credentials(cf_auth_user, "");
+				client->db->auth_user_credentials = find_exist_or_add_new_global_credentials(cf_auth_user, "");
```
JelteF, line 488: This should fail too right if it's NULL?

src/objects.c
```diff
@@ -488,33 +488,48 @@ PgDatabase *register_auto_database(const char *name)
...
+		return create_new_global_user(name, passwd);
 
-	passwd = passwd ? passwd : "";
-	safe_strcpy(user->credentials.passwd, passwd, sizeof(user->credentials.passwd));
-	user->credentials.dynamic_passwd = strlen(passwd) == 0;
-	return user;
+	return update_global_user_passwd(user, passwd);
 }
 
+//TODO: is not used.
```
JelteF, line 532: let's remove the function then.

src/objects.c
```diff
@@ -1002,6 +1022,28 @@ bool queue_fake_response(PgSocket *client, char request_type)
 	return res;
 }
 
+/* Find an exist global user or add a new global user */
```
JelteF, line 1025:
```suggestion
/* Find an existing global user or add a new global user */
```

src/objects.c
```diff
@@ -1002,6 +1022,28 @@ bool queue_fake_response(PgSocket *client, char request_type)
...
+{
+	PgGlobalUser *user = find_global_user(name);
+
+	if (!user)
+		user = create_new_global_user(name, passwd);
+
+	return user;
+}
+
+/* Find an exist global credentials or add a new global credentials */
```
JelteF, line 1036:
```suggestion
/* Find an existing global credentials or add a new global credentials */
```

JelteF (changes requested):

JelteF:
Thank you for the contribution. I left some feedback on the code.

src/client.c
dmitry-lipetsk: Should I write something like this

```c
slog_error(client, "set_pool(): failed to allocate a new global credentials");
disconnect_client(client, true, "bouncer resources exhaustion");
return false;
```
?

I will rename these new functoin and will return to this problem.

include/objects.h
dmitry-lipetsk: Done. Thanks!

dmitry-lipetsk:
Hello,

I hope, all the requested changes were done. Or no?

src/objects.c
```diff
@@ -496,39 +496,45 @@ PgDatabase *register_auto_database(const char *name)
...
+
+	aatree_insert(&user_tree, (uintptr_t)user->credentials.name, &user->credentials.tree_node);
+	user->pool_mode = POOL_INHERIT;
+	user->pool_size = -1;
+
+	return update_global_user_passwd(user, passwd);
+}
+
+/* add or update client users */
+PgGlobalUser *add_global_user(const char *name, const char *passwd)
```
JelteF, line 530: This does basically the same as `find_or_add_global_user`, except that it also updates the password of the global user. 

Let's move these functions closer together and make this difference clear from the naming and function comment.

JelteF: Actually, maybe it's better to completely remove `add_global_user` and explicitely call `update_global_user_passwd` in all places where we need to update the password.

JelteF:
@dmitry-lipetsk are you still planning to move this forward? 

On a related note, there's now a discord channel to discuss PgBouncer development, feel free to join there: https://discordapp.com/channels/1258108670710124574/1300532992304742481

dmitry-lipetsk:
> @dmitry-lipetsk are you still planning to move this forward?

Hello @JelteF,

I am sorry but I completely lost the context of this project and do not have free time to get it back :-)

You can do with this PR all what you want.

> On a related note, there's now a discord channel to discuss PgBouncer development, feel free to join there: https://discordapp.com/channels/1258108670710124574/1300532992304742481

Thanks but, it seems to me, this resource is not available in my country.

PR #1114, opened 2024-07-09
QA test for issue #1104

dmitry-lipetsk:
Hello All,

See #1104 for details.

This PR will fail without PR #1105.

> FAILED test/qa/github/test_qa_gh001104.py::test_qa_gh001104 - psycopg.OperationalError: connection is bad: connection to server on socket "/tmp/pytest-of-dima/pytest-124/popen-w0/test_qa_gh0011040/bouncer/.s.PGSQL...

It is an expected behavior.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -6,7 +6,7 @@ PG_LDFLAGS = -L$(shell pg_config --libdir)
 USUAL_DIR = ../lib
 
 SUBLOC = test
-DIST_SUBDIRS = ssl
+DIST_SUBDIRS = ssl qa/github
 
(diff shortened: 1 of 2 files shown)


JelteF:
Can you add this test to `test_misc.py` instead of creating a dedicated file. Also can you include it in #1104 instead of making it a separate PR?.

dmitry-lipetsk:
> Can you add this test to `test_misc.py` instead of creating a dedicated file. Also can you include it in #1104 instead of making it a separate PR?.

Ok.

A name "test_qa_gh1104" is ok, or I have to use another name?

JelteF:
Closing this since #1105 is now merged

Issue #1116, opened 2024-07-10
PGbouncer crashes with put_in_order error


AndrewJackson2020:
PGbouncer crashses with put_in_order error when connecting as auth_user. This appears to effect the most recent release of pgbouncer but not the 1.21.0 release. I am not 100% sure when this error was originally introduced. 

I was able to recreate the error in the below patch using the pytest framework in pgbouncer.

Please let me know if there is any additional information that I can provide about this.

```
index 62bd107..4ad49c1 100644
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -130,6 +130,7 @@ def pg(tmp_path_factory, cert_dir):

     pg.sql("create database unconfigured_auth_database")
     pg.sql("create user bouncer")
+    pg.sql("create user pgbouncer_auth with superuser")
     pg.sql("create user pswcheck with superuser createdb password 'pgbouncer-check';")
     pg.sql("create user someuser with password 'anypasswd';")
     pg.sql("create user maxedout;")
diff --git a/test/test_auth.py b/test/test_auth.py
index e774fe5..3f7bbe4 100644
--- a/test/test_auth.py
+++ b/test/test_auth.py
@@ -890,3 +890,33 @@ def test_peer_auth_ident_map(bouncer):
         host=f"{bouncer.admin_host}",
         user="bouncer",
     )
+
+
+def test_auth_user_crash_issue(bouncer) -> None:
+    """
+    Test that for issue that causes pgbouncer to crash given the following conditions
+      1. Connect to pgbouncer as auth user
+      2. Auth user is not listed in userlist.txt
+    """
+
+    hba_conf_file = bouncer.config_dir / "hba.conf"
+
+    config = f"""
+        [databases]
+        postgres = host={bouncer.pg.host} dbname=postgres port={bouncer.pg.port} min_pool_size=2
+
+        [pgbouncer]
+        listen_addr = {bouncer.host}
+        listen_port = {bouncer.port}
+        auth_type = hba
+        auth_hba_file = {hba_conf_file}
+        auth_user = pgbouncer_auth
+        auth_dbname = postgres
+        logfile = {bouncer.log_path}
+    """
+
+    with bouncer.run_with_config(config):
+        with bouncer.conn(dbname="postgres", user="pgbouncer_auth", password="asdasd") as cn:
+            with cn.cursor() as cur:
+                cur.execute("select 1")
+
```

JelteF:
[...]

PR #1117, opened 2024-07-12
Fix put_in_order crashes

AndrewJackson2020:
This PR attempts to fix issue #1116. 

The user refactor #1052 led to an issue with pgbouncer instances that use auth pass through. Attempting to connect to the auth_db with the auth_user from a sql client through pgbouncer causes pgbouncer to crash. The workaround for is to add the auth_user to the auth_file so that pgbouncer instantiates a non dynamic user and thus does not attempt to acquire the auth_users credentials in a dynamic way.

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -516,7 +516,6 @@ struct PgPool {
  * that use these credentials.
  */
 struct PgCredentials {
-	struct List pool_list;		/* list of pools where pool->user == this user */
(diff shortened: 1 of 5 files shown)


JelteF:
Closing this in favor of #1120

JelteF:
Thank you for the contribution though. I've included your test in the new PR and mentioned you as co-author.

PR #1121, opened 2024-07-19
fixed doc build race

SeleznevPavel:
To run nightly unit tests, we build a bouncer and sometimes encounter an error caused by parallel build, when one part causes files to be cleared.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -186,8 +186,11 @@ htmls:
 		mkdir -p html && $(PANDOC) $$f -o html/`basename $$f`.html; \
 	done
 
-doc/pgbouncer.1 doc/pgbouncer.5:
-	$(MAKE) -C doc $(@F)
+doc/pgbouncer.1:
+	$(MAKE) -C doc pgbouncer.1
+
+doc/pgbouncer.5:
+	$(MAKE) -C doc pgbouncer.5
 
 lint:
 	flake8
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -21,6 +21,7 @@ export PACKAGE_VERSION
 
 pgbouncer.%: pgbouncer_%.md
 	$(PANDOC) -s -t man -o $@ $<
+	rm $<
 
 pgbouncer_1.md: filter.py frag-usage-man.md usage.md
 	$(PYTHON) $^ >$@


Makefile
```diff
@@ -186,8 +186,11 @@ htmls:
 		mkdir -p html && $(PANDOC) $$f -o html/`basename $$f`.html; \
 	done
 
-doc/pgbouncer.1 doc/pgbouncer.5:
-	$(MAKE) -C doc $(@F)
+doc/pgbouncer.1:
+	$(MAKE) -C doc pgbouncer.1
+
+doc/pgbouncer.5:
+	$(MAKE) -C doc pgbouncer.5
```
JelteF, line 193: Is this change really necessary to fix the issue? Or is the `rm` enough.

SeleznevPavel: We do a lot of nightly runs and this fix in our internal repository allowed us to eliminate the time spent on false build errors. I consider it necessary to share with the community

eulerto:
Could you describe the steps you are using to build PgBouncer? Doesn't `make -C doc clean` fix it?

Makefile
JelteF: as far as I can tell this part of the change achieves nothing. It's simply a more verbose (but arguably clearer) syntax for the same thing.

doc/Makefile
```diff
@@ -21,6 +21,7 @@ export PACKAGE_VERSION
 
 pgbouncer.%: pgbouncer_%.md
 	$(PANDOC) -s -t man -o $@ $<
+	rm $<
```
JelteF, line 24: I guess this is the important part of the PR for you. I think I'm fine with this, but I don't fully understand why this matters in your case. Are you running multiple builds at the same time on the same directory?

SeleznevPavel: Let me describe the situation again in more detail.

We have a build process of the product Pangolin, which includes forks of Postgress and Pgbouncer.
After assembly, various tests are launched. Our team is quite large and there are more than 500 similar build-test launches per day.
From time to time we have the problem that the Pgbouncer fork did not build. Having sorted it out, we found out that the problem is in the parallel assembly of documentation,
when one branch of the parallel assembly can delete the necessary data for another.
If you run it once, you won’t see the problem (otherwise everyone would be talking about it), but if you do it in a loop with some extra processes on server
> make
> make clean
then the problem will be revealed.

We have a parallel build process, within which various extensions are also built, and this parallelization creates an additional random load on the hardware, 
which leads to different timings when building documentation

petere: This doesn't provide an analysis of why the exiting rules are faulty, or why this fixes it. Deleting an input file in a make rule almost certainly doesn't make sense.

PR #1124, opened 2024-07-23
Add files pgbouncer_hba.conf and pgident.conf to tarball

vbwagner:
These two files are required by regression tests since 1.23.0 but haven't been packed into tarball during make dist. This commit fixes this problem.


eulerto:
I would add `test/test_replication.py` that is missing.

vbwagner:
Thanks! I 've rebuild all my packages with this test included

PR #1132, opened 2024-08-02
ci: update Postgres version

eulerto:
Use the latest stable version of PostgreSQL. It covers the latest protocol changes.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -215,7 +215,7 @@ task:
     - make -j4
   test_script:
     - brew ls -v postgresql@${PGVERSION}
-    - source /venv/bin/activate && "PATH=/opt/homebrew/opt/postgresql@${PGVERSION}/bin:$PATH make -j4 check CONCURRENCY=4"
+    - source venv/bin/activate && "PATH=/opt/homebrew/opt/postgresql@${PGVERSION}/bin:$PATH make -j4 check CONCURRENCY=4"
   install_script:
     - make -j4 install
   always:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -199,6 +199,8 @@ task:
   env:
     HAVE_IPV6_LOCALHOST: yes
     USE_SUDO: true
+    CPPFLAGS: -I/opt/homebrew/opt/openssl@3/include
+    LDFLAGS: -L/opt/homebrew/opt/openssl@3/lib
     LD_LIBRARY_PATH: /opt/homebrew/opt/postgresql@${PGVERSION}/lib
     PKG_CONFIG_PATH: /opt/homebrew/opt/postgresql@${PGVERSION}/lib/pkgconfig
   setup_script:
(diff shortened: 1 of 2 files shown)


JelteF:
The OSX build seems broken because of this.

eulerto:
It seems the setup errors are gone. Hmm. I don't know why this test is failing like this. @JelteF do you have an idea?

JelteF:
The first error I've seen before when not using the `master` version of `psycopg`. Now that psycopg 3.2 is out though there's no need to use the master version of it anymore. So I think it makes sense to change it to install the released version in `requirements.txt` and then change all the tests that break because we are doing so.

eulerto:
I'm fine with the current PR. I fixed the Mac OS. The alpine is still broken but it is not because of this PR but another issue with TLS. @JelteF objections?

.cirrus.yml
```diff
@@ -195,25 +195,26 @@ task:
...
     - sudo pfctl -f /etc/pf.conf
-  env:
-    CPPFLAGS: -I/opt/homebrew/opt/openssl@3/include
-    LDFLAGS: -L/opt/homebrew/opt/openssl@3/lib
   build_script:
     - ./autogen.sh
     - ./configure --prefix=$HOME/install --enable-werror
     - make -j4
+  env:
+    PATH: /opt/homebrew/opt/postgresql@${PGVERSION}/bin:$PATH
```
JelteF, line 215: can this be merged with the other `env` section?

.cirrus.yml
```diff
@@ -195,25 +195,26 @@ task:
...
   test_script:
-    - source venv/bin/activate && make check -j4 CONCURRENCY=4
+    - source venv/bin/activate && make -j4 check CONCURRENCY=4
```
JelteF, line 217: Was this change actually needed?

JelteF (approved):
LGTM appart from two small comments. Please note that we should update the required jobs in the github repo settings before (or quickly after) merging.

.cirrus.yml
eulerto: I think so. The main reason I added here was that that path exists only after installing the PostgreSQL package.

.cirrus.yml
eulerto: There is no strong reason. Consistency with the similar commands. I should have mentioned it.

eulerto:
> Please note that we should update the required jobs in the github repo settings before (or quickly after) merging.

What settings are you referring to? AFAICS we are not using GH Actions. Cirrus CI is a third-part app that is integrated into GitHub.

.cirrus.yml
JelteF: Let's merge them then, because it's a bit confusing that we have two `env` keys in the same object. Based on some quick searching it doesn't seem specified what that means according the yaml spec (it seems like github merges the objects)

JelteF:
> What settings are you referring to?

I the "Required" tag after the CI job: 
image

This checks are required for merging into the master branch, but they contain the 14 number. They can be changed in the branch protection settings here: https://github.com/pgbouncer/pgbouncer/settings/branch_protection_rules/27973333

JelteF:
FYI: I made the change I suggested to the `env` entry, merged the PR and updated the branch protection rules

PR #1134, opened 2024-08-06
add nix env

sebastianwebber:
current experiment works with:

make CC="zig cc"


sebastianwebber:
I'm sorry. I'm doing some tests with `zig build,` and my intention was not to add a PR for this.

Issue #1136, opened 2024-08-08
Client Connection Limits Per User/DB


AndrewJackson2020:
Is it possible to set client connection limits per user/DB? I know there is the setting max_client_conn but this applies to the entire pgbouncer instance not to individual users/databases.

JelteF:
There is `max_user_connections` that you can set for a user: https://www.pgbouncer.org/config.html#section-users

And there's `max_db_connections` that you can set for a database: https://www.pgbouncer.org/config.html#max_db_connections-1

AndrewJackson2020:
Correct me if I am wrong but both of those options limit the number of users that are allowed connections to the database backends but do not limit the number of users that are allowed to connect to the pgbouncer frontend. 

The issue that I am concerned about for example is if a user is in transaction mode, that user_a has max_user_connections set to 2. max_client_conn is set to 4. user_a can create 4 connections to the pgbouncer frontend, the max_user_connections will disallow the extra 2 users from actually getting database connections but does nothing to prevent that one user from taking up the entirety of the max_client_conn limit. This could potentially lead to an issue where user_a is locking out user_b from database access.

AndrewJackson2020:
I just raised #1137 as a proposal to add a user configuration parameter to limit pgbouncer client connections by user.

eulerto:
Done #1137 and #1138.

PR #1137, opened 2024-08-09
Add client user connection limit

AndrewJackson2020:
This PR adds a user level client connection limit. I believe that currently user connections can only be limited by how many people are given connections  to the database backends but there is no way to limit the number of connections to pgbouncer itself. Combined with the global connection limit this means that one user could use up all connections on the pgbouncer instance. This PR proposes a fix to this problem.

Related to #1136 (because it partially fixes it)


include/bouncer.h
```diff
@@ -540,7 +540,9 @@ struct PgGlobalUser {
 	int pool_mode;
 	int pool_size;				/* max server connections in one pool */
 	int max_user_connections;	/* how much server connections are allowed */
+	int max_user_client_connections;	/* how much server connections are allowed */
```
JelteF, line 543:
```suggestion
	int max_user_client_connections;	/* how many server connections are allowed */
```

include/bouncer.h
```diff
@@ -540,7 +540,9 @@ struct PgGlobalUser {
...
 	int connection_count;	/* how much connections are used by user now */
+	int client_connection_count;	/* how much connections are used by user now */
```
JelteF, line 545:
```suggestion
	int client_connection_count;	/* how many client connections are used by user now */
```

include/bouncer.h
```diff
@@ -540,7 +540,9 @@ struct PgGlobalUser {
 	int pool_mode;
 	int pool_size;				/* max server connections in one pool */
 	int max_user_connections;	/* how much server connections are allowed */
+	int max_user_client_connections;	/* how much server connections are allowed */
 	int connection_count;	/* how much connections are used by user now */
```
JelteF, line 544:
```suggestion
	int connection_count;	/* how many server connections are used by user now */
```

src/client.c
```diff
@@ -489,6 +504,8 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 				if (!client->db->auth_user_credentials)
 					client->db->auth_user_credentials = add_global_credentials(cf_auth_user, "");
 			}
+
+
```
JelteF, line 508: Can you undo this?

JelteF (changes requested):
This seems like a good solution to the described problem. A similar setting would probably be useful per database, but that doesn't need to be in this PR. 

Some things that should be changed about the PR:
1. Github Actions should be green (fix C warning and fix formatting)
2. A test should be added (`test/test_limits.py` seems like a good place to add one)
3. The new option should be documented in `doc/config.md`

src/client.c
AndrewJackson2020: Done

include/bouncer.h
AndrewJackson2020: Done

include/bouncer.h
AndrewJackson2020: Done

include/bouncer.h
AndrewJackson2020: Done

eulerto:
I didn't review this PR yet but I noticed that you did lots of changes in the documentation that is not related to this feature. If you want to propose a whitespace cleanup, create a separate PR. We generally use `/* comment */` instead of `// comment`.

BTW, I'm fine with this feature. It is equivalent to `CREATE ROLE ... CONNECTION LIMIT`.

AndrewJackson2020:
> I didn't review this PR yet but I noticed that you did lots of changes in the documentation that is not related to this feature. If you want to propose a whitespace cleanup, create a separate PR. We generally use `/* comment */` instead of `// comment`.
> 
> BTW, I'm fine with this feature. It is equivalent to `CREATE ROLE ... CONNECTION LIMIT`.

Apologies about that, my editor eliminates trailing whitespace default. Will revert momentarily.

AndrewJackson2020:
> I didn't review this PR yet but I noticed that you did lots of changes in the documentation that is not related to this feature. If you want to propose a whitespace cleanup, create a separate PR. We generally use `/* comment */` instead of `// comment`.
> 
> BTW, I'm fine with this feature. It is equivalent to `CREATE ROLE ... CONNECTION LIMIT`.

I reverted the whitespace removal and changed the comment style to conform with the rest of the codebase.

JelteF (changes requested):
Thanks! That looks a lot better already. I left a review just now on your other #1138 PR.  Most of the feedback there should also be applied here. Other than that it looks pretty good.

I still have want to do some manual tests of both PRs though

AndrewJackson2020:
@JelteF Made all of the documentation changes and added a test of the stats command in the unit tests. Also split the unit test out into 2 separate ones: one that is meant to deny a connection and one that is meant to accept a connection, I feel this is a bit cleaner.

A few additional thoughts:
1. Should this PR and #1138 be merged because they are so similar and have similar feedback?
2. Should the name of the variables be changed from max_user_client_connections/max_user_connections to max_user_client_connections/max_user_server_connections?
3. Should the tests of the stats command  be moved to test/test_admin.py or is it fine as is?

src/objects.c
```diff
@@ -1393,6 +1393,11 @@ void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...
  */
 void disconnect_client(PgSocket *client, bool notify, const char *reason, ...)
 {
+	if (client->login_user_credentials) {
+		if (client->login_user_credentials->global_user) {
+			client->login_user_credentials->global_user->client_connection_count--;
```
JelteF, line 1398: This substraction should only be done when this specific clien actually increased the connection count before. 

Right now if a user is at the limit. The next failing connection attempt (due to hitting the limit) will then also decrease the connection count. So the connection attempt after that will succeed again, eventhough we're still at the client connection limit.

JelteF:
>     1. Should this PR and [Add db client connection limit #1138](https://github.com/pgbouncer/pgbouncer/pull/1138) be merged because they are so similar and have similar feedback?


>     2. Should the name of the variables be changed from max_user_client_connections/max_user_connections to max_user_client_connections/max_user_server_connections?

Let's keep them as is. I don't think the better naming (because I agree it is a bit confusing now) is worth breaking backwards compatibility of the config setting names.

>     3. Should the tests of the stats command  be moved to test/test_admin.py or is it fine as is?

It's fine as is imho.

src/objects.c
AndrewJackson2020: Okay so I added a unit test assert involving the situation you described above. Moved the increment to right before the initial check and changed the inequality accordingly. Seems to satisfy all test cases though there might be another corner case I'm not seeing.

An alternate way of going about this is adding a boolean to each client that specifies if they contribute to the user level count and check that before decrementing the count. I feel like the logic in this commit is better since it adds less variables, as long as the logic is correct.

Another thing that I notice is that this new logic does not apply to PAM authed users though I feel like it should.

JelteF: > An alternate way of going about this is adding a boolean to each client that specifies if they contribute to the user level count and check that before decrementing the count. I feel like the logic in this commit is better since it adds less variables, as long as the logic is correct.

Agreed.

> Another thing that I notice is that this new logic does not apply to PAM authed users though I feel like it should.

Agreed. Everywhere we set `client->login_user_credentials` we should do this incrementing/decrementing correctly. Also the block below where auth_user is used should do this. I think the easiest way to do this is do

```
client->login_user_credentials = /* something */
if (!check_user_connection_count())
    return;

bool check_user_connection_count() {
    /* your check */
}
```

src/client.c
```diff
@@ -476,6 +478,22 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 		}
 	} else {
 		client->login_user_credentials = find_global_credentials(username);
+
+		/* Check client_connection count limit */
+		if (client->login_user_credentials) {
+			client->login_user_credentials->global_user->client_connection_count++;
```
JelteF, line 484: Let's add a comment here for clarity.
```suggestion
			/* increment count now, so that we can decrement it safely in disconnect_client if limit was reached */
			client->login_user_credentials->global_user->client_connection_count++;
```

src/objects.c
AndrewJackson2020: Encapsulated the logic into the function and added the call to everywhere client->login_user_credentials is assigned. Tests all passed but the surface area of the change is now much bigger which may mean more testing is required (limits correctly incremented, decremented and enforced for pam users, admin users, etc). Also the check_user_connection_count function could probably be refactored a bit (lot of nesting).

AndrewJackson2020: @JelteF Expanded test coverage to include admin users. Refactored the check_user_connection_count function to prefer early returrns instead of nested if statements. Refactored the tests so they should be faster, rely less on asynchronously running pg_sleep.

Could also expand test coverage to stats users as I believe are instantiated differently than admin users. Also missing is test coverage around PAM users though I dont believe there is PAM coverage elsewhere in the test codebase either.

Let me know if anything else needs to be done with this PR in order to get it in a mergable state.

JelteF: If tests pass, I think this is probably good. I'm unavailable for the next ~3 weeks though.

AndrewJackson2020: Understood, thanks for the heads up.

JelteF (changes requested):
Sorry for the delay in reviewing. I think this is almost good now, some final suggested changes.

src/admin.c
```diff
@@ -1676,6 +1680,9 @@ bool admin_pre_login(PgSocket *client, const char *username)
 		if (res >= 0 && peer_uid == getuid()
 		    && strcmp("pgbouncer", username) == 0) {
 			client->login_user_credentials = admin_pool->db->forced_user_credentials;
+			if (!check_user_connection_count(client)) {
```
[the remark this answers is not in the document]
AndrewJackson2020: Good point. Looking at the implementation of the server wide max_client_conn setting it behaves in the way you are describing. Will update that.

doc/config.md
```diff
@@ -1420,7 +1432,7 @@ client connections, if a password-based authentication method is
 configured.  Second, they are used as the passwords for outgoing
 connections to the backend server, if the backend server requires
 password-based authentication (unless the password is specified
-directly in the database's connection string).
+directly in the database's connection string).
```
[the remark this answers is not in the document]
AndrewJackson2020: Sorry about that, will fix.

src/client.c
```diff
@@ -504,6 +545,9 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 
 			slog_info(client, "no such user: %s", username);
 			client->login_user_credentials = calloc(1, sizeof(*client->login_user_credentials));
+			if (!check_user_connection_count(client)) {
+				return false;
+			}
```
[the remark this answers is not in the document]
AndrewJackson2020: Will do. Will also make sure test coverage for this feature includes users from the auth DB.

src/admin.c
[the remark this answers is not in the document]
AndrewJackson2020: This is fixed now. Added unit tests to verify that the connection limit applies to the stats users accessing the admin console but not admin users accessing the admin console.

AndrewJackson2020:
@JelteF Pushed a new commit that fixes a bunch of the issues the issues that you mentioned. 

Found that the auth pass through issue that you mentioned in #1138 applies to this PR as well. Implemented a somewhat hacky fix to this. Would appreciate your feedback on this or if you have any alternative ideas.

Also expanded unit test coverage for stats users and auth pass through users. One area that isnt covered is PAM but looking through the unit tests I dont think there is any test coverage of PAM currently so I think adding that here might be out of scope for this PR though it would be very beneficial.

AndrewJackson2020:
@JelteF Fixed the macOS tests by scaling back the run_with_config usage as described in #1138.

Let me know if this solution works for you or if it needs to be reworked more.

AndrewJackson2020:
@JelteF Anything else I can do here to help this PR get merged?

JelteF:
No, it's blocked on me finding some time to review PgBouncer PRs again. Busy busy.

src/client.c
```diff
@@ -420,6 +420,42 @@ static bool finish_set_pool(PgSocket *client, bool takeover)
...
+bool check_user_connection_count(PgSocket *client)
+{
+	int client_connection_count;
+	int max_user_client_connections;
+
+	/* Check client_connection count limit */
+	if (!client->login_user_credentials)
+		return true;
+
+	/* increment count now, so that we can decrement it safely in disconnect_client if limit was reached */
```
JelteF, line 432: That comment seems out of date now.
```suggestion
```

src/client.c
```diff
@@ -420,6 +420,41 @@ static bool finish_set_pool(PgSocket *client, bool takeover)
...
+	int max_user_client_connections;
+
+	/* Check client_connection count limit */
+	if (!client->login_user_credentials)
+		return true;
+
+	if (!client->login_user_credentials->global_user)
+		return true;
+
+	if (!client->user_connection_counted) {
```
JelteF, line 435: I agree this is not pretty. But given the amount of counting bugs found in earlier versions, I prefer this way because this one is obviously correct and will stay correct in the future. I added a check for user_connection_counted in the decrement path too.

AndrewJackson2020:
@JelteF anything else you want done with this PR before merging?

JelteF:
I don't remember why I didn't press merge yet, maybe because CI was failing for unrelated reasons.

PR #1138, opened 2024-08-09
Add db client connection limit

AndrewJackson2020:
Related to #1137. This PR allows the user to configure client connection limits at the database level. This prevents issues where SQL clients all connecting to a single database consume the entire connection limit defined in max_client_conn which effectively locks all other databases from being accessed.


doc/config.md
```diff
@@ -1226,6 +1226,11 @@ they are logged but ignored otherwise.
 Set the pool mode specific to this database. If not set,
 the default `pool_mode` is used.
 
+### max_db_client_connections
```
JelteF, line 1229: nit: let's move this below max_db_connections

doc/config.md
```diff
@@ -1226,6 +1226,11 @@ they are logged but ignored otherwise.
...
+
+Configure a database-wide client connection maximum. Should be used in conjunction with max_client_conn 
+to limit the number of connections that pgbouncer is allowed to accept.
+
 ### max_db_connections
 
 Configure a database-wide maximum (i.e. all pools within the database will
```
JelteF, line 1236:
```suggestion
Configure a database-wide maximum of server connections (i.e. all pools within the database will
```

doc/usage.md
```diff
@@ -570,6 +570,12 @@ max_connections
 current_connections
 :   Current number of connections for this database.
 
+max_client_connections
+:   Maximum number of allowed connections for this pgbouncer instance, as set by max_db_client_connections per database.
```
JelteF, line 574:
```suggestion
:   Maximum number of allowed client connections for this pgbouncer instance, as set by max_db_client_connections per database.
```

doc/usage.md
```diff
@@ -570,6 +570,12 @@ max_connections
 current_connections
 :   Current number of connections for this database.
```
JelteF, line 571:
```suggestion
:   Current number of server connections for this database.
```

doc/usage.md
```diff
@@ -570,6 +570,12 @@ max_connections
...
 
+max_client_connections
+:   Maximum number of allowed connections for this pgbouncer instance, as set by max_db_client_connections per database.
+
+current_client_connections
+:   Current number of connections for this pgboucer instance.
```
JelteF, line 577:
```suggestion
:   Current number of client connections for this database.
```

doc/usage.md
JelteF: and let's make the same change for `max_connections` above (I can't comment on that line because it's outside the diff according to github...)

src/main.c
```diff
@@ -273,6 +274,7 @@ static const struct CfKey bouncer_params [] = {
 	CF_ABS("logfile", CF_STR, cf_logfile, 0, ""),
 	CF_ABS("max_client_conn", CF_INT, cf_max_client_conn, 0, "100"),
 	CF_ABS("max_db_connections", CF_INT, cf_max_db_connections, 0, "0"),
+	CF_ABS("max_db_client_connections", CF_INT, cf_max_db_client_connections, 0, "0"),
```
JelteF, line 277: You didn't document this global option.

src/client.c
```diff
@@ -432,6 +432,15 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 		strlcpy(client->db->name, dbname, sizeof(client->db->name));
 	}
 
+	if (client->db->max_db_client_connections > 0) {
```
JelteF, line 435: This is currently ignoring the global `max_db_client_connections` setting.

```suggestion
	if (database_max_client_connections(client->db) > 0) {
```

JelteF (changes requested):

test/test_limits.py
```diff
@@ -21,6 +21,22 @@ async def test_max_client_conn(bouncer):
...
+    test_db = "conn_limit_db"
+    test_user = "muser1"
+    result = bouncer.asleep(3, dbname=test_db, user=test_user)
+    await asyncio.sleep(1)
+    # should still be allowed, since it's the last allowed connection
+    await bouncer.atest(dbname=test_db, user=test_user)
+    result_last = bouncer.asleep(3, dbname=test_db, user=test_user)
+    await asyncio.sleep(1)
+    with pytest.raises(psycopg.OperationalError, match=r"max_db_client_connections"):
+        await bouncer.atest(dbname=test_db, user=test_user)
```
JelteF, line 35: if it's easy, could you also check the values of `max_client_connections` and `current_client_connections` using `SHOW DATABASES`? Coverage of our stats commands is currentl very minimal (or non-existent). If it's not too hard to add, this would be a good moment to improve that.

doc/config.md
AndrewJackson2020: Done. Was thinking about the naming convention. Currently we are using 
max_db_client_connections/max_db_connections. Perhaps we should change it to max_db_client_connections/max_db_server_connections?

src/main.c
AndrewJackson2020: This should be documented now.

test/test_limits.py
AndrewJackson2020: Will look into this.

doc/usage.md
AndrewJackson2020: Done

test/test_limits.py
```diff
@@ -21,6 +21,224 @@ async def test_max_client_conn(bouncer):
...
+    test_user = "muser1"
+    result = bouncer.asleep(3, dbname=test_db, user=test_user)
+    await asyncio.sleep(1)
+    # should still be allowed, since it's the last allowed connection
+    dbs = bouncer.admin("SHOW DATABASES")
+    db = [db for db in dbs if db[0] == test_db][0]
+    assert db == (
+        test_db,
+        "127.0.0.1",
+        10201,
```
JelteF, line 222: Let's not test that the port is equal to a hardcoded value. As CI already shows this value is not necessarily stable across runs. I think you could use `bouncer.port` or `pg.port` though (whichever is the right port here).

doc/config.md
JelteF: Let's keep as is.

AndrewJackson2020: understood.

test/test_limits.py
AndrewJackson2020: This should be good now.

AndrewJackson2020:
Found a similar issue with this PR as with #1137 wherein the increment was not done in the correct place. This should be fixed now and added an extra assert in one of the test cases.

The implementation of #1137 is a bit more complicated than this one in that databases don't get created in multiple places like users. Because of this I did not encapsulate the limit check into a function. Let me know if you want this in a function anyways for consistency sake.

test/test_limits.py
AndrewJackson2020: Nice, it looks like pg.port fixed the issue.

doc/config.md
```diff
@@ -176,6 +176,22 @@ pool.
 
 Default: 0 (unlimited)
 
+### max_db_client_connections
+
+Do not allow more than this many client connections to pgbouncer per database
```
JelteF, line 181:
```suggestion
Do not allow more than this many client connections to PgBouncer per database
```

doc/config.md
```diff
@@ -1239,9 +1255,14 @@ the default `pool_mode` is used.
...
 ### max_db_connections
 
-Configure a database-wide maximum (i.e. all pools within the database will
+Configure a database-wide maximum of server connections (i.e. all pools within the database will
 not have more than this many server connections).
 
+### max_db_client_connections
+
+Configure a database-wide client connection maximum. Should be used in conjunction with max_client_conn
+to limit the number of connections that pgbouncer is allowed to accept.
```
JelteF, line 1264:
```suggestion
to limit the number of connections that PgBouncer is allowed to accept.
```

src/client.c
```diff
@@ -432,6 +432,17 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 		strlcpy(client->db->name, dbname, sizeof(client->db->name));
 	}
 
+	if (database_max_client_connections(client->db) > 0) {
+		/* increment count now, so that we can decrement it safely in disconnect_client if limit was reached */
+		client->db->client_connection_count++;
```
JelteF, line 437: When using an `auth_user` something is still not correct here. If you try to connect using a user that does not exist the count is not decremented correctly somehow.

`auth_user` does work in a bit of a weird way. See `start_auth_query` for some details, as well as this PR where I'm trying to fix another issue that popped up due to the confusing way in which it works: https://github.com/pgbouncer/pgbouncer/pull/1166

JelteF (changes requested):

src/client.c
AndrewJackson2020: I am noticing the same thing in the #1137. Specifically I'm seeing that it logs in 4 different times initially and of those 4 times 3 logins are permanently retained. Then every additional time you log in it "logs in" 2 times and of those 2 log ins one is not decremented. Definitely something weird with this implementation as is and definitely not something we can merge in as is. 

```console
$ psql  'postgresql://maxedout3@127.0.0.1:10203/pgbouncer' -c 'show users' | grep pswcheck_not_in_auth_file
 pswcheck_not_in_auth_file |         1 |           |                    0 |                   0 |                         100 |                          0

$ psql  'postgresql://pswcheck_not_in_auth_file@127.0.0.1:10203/authdb'  -c select 1;
--
(1 row)


$ psql  'postgresql://maxedout3@127.0.0.1:10203/pgbouncer' -c 'show users' | grep pswcheck_not_in_auth_file
 pswcheck_not_in_auth_file |         1 |           |                    0 |                   0 |                         100 |                          3

$ psql  'postgresql://pswcheck_not_in_auth_file@127.0.0.1:10203/authdb'  -c select 1;
--
(1 row)


$ psql  'postgresql://maxedout3@127.0.0.1:10203/pgbouncer' -c 'show users' | grep pswcheck_not_in_auth_file
 pswcheck_not_in_auth_file |         1 |           |                    0 |                   0 |                         100 |                          4
$ psql  'postgresql://pswcheck_not_in_auth_file@127.0.0.1:10203/authdb'  -c select 1;
--
(1 row)


$ psql  'postgresql://maxedout3@127.0.0.1:10203/pgbouncer' -c 'show users' | grep pswcheck_not_in_auth_file
 pswcheck_not_in_auth_file |         1 |           |                    0 |                   0 |                         100 |                          5
```

AndrewJackson2020: So super hacky solution but I was able to fix this by essentially adding an additional boolean variable to PgSocket that specifies whether or not that client has been contributed to the total count of clients. This ensures that the recursive set_pool calls only increment the count one time. Will push this in a bit, after that we can figure out if there is a better way of doing this.

AndrewJackson2020: @JelteF Pushed the hacky solution mentioned in the last comment. Similar solution implemented for #1137. Also expanded test coverage to admin users, stat users, and auth pass through users. 

Some macOS tests are failing but I believe these are due to flaky tests similar to #1137 as the tests that are failing are different between the two test runs. Its possible that I might have made some changes that impact that global state of the test running but I don't understand why it would not impact linux but would impact macOS. Hard for me to look into this further as I don't have a dev environment set up for MacOS.

AndrewJackson2020: @JelteF I believe that I fixed the macOS tests by removing my usage of the run_with_config method. I think something about how this is implemented makes tests flakier on macOS but I do not have a development environment to recreate and try to fix this.

Either way please lmk what you think about this updated solution.

src/client.c
```diff
@@ -420,6 +420,30 @@ static bool finish_set_pool(PgSocket *client, bool takeover)
...
+{
+	if (!client->contributes_db_client_count) {
+		client->contributes_db_client_count = true;
+		client->db->client_connection_count++;
+	}
+
+	if (database_max_client_connections(client->db) <= 0)
+		return true;
+
+	/* increment count now, so that we can decrement it safely in disconnect_client if limit was reached */
```
JelteF, line 433:
```suggestion
```

src/objects.c
```diff
@@ -1401,6 +1401,9 @@ void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...
  */
 void disconnect_client(PgSocket *client, bool notify, const char *reason, ...)
 {
+	if (client->db) {
```
JelteF, line 1404:
```suggestion
	if (client->db && client->contributes_db_client_count) {
```

test/utils.py
```diff
@@ -975,6 +975,16 @@ async def wait_until_running(self):
...
+    def admin_dict(self, query, params, **kwargs):
+        with self.cur(**kwargs) as cur:
+            cur.execute(query, params=params)
+            try:
+                return cur.fetchall()
+            except psycopg.ProgrammingError as e:
+                if "the last operation didn't produce a result" == str(e):
+                    return None
+                raise
+
```
JelteF, line 987: This seems unused now let's remove it
```suggestion
```

JelteF:
I'm not sure what's up with this one, but every CI job seems to be failing. I seem to remember this was basically good to merge. Was there a bad merge conflict resolution or something??

AndrewJackson2020:
Bad merge on my end, apologies for that.

AndrewJackson2020:
Fixed the tests. The merge created some excessive line by line diffs that I can clean up as well.

JelteF:
Another merge conflict... I guess after I merged #1193

AndrewJackson2020:
> Another merge conflict... I guess after I merged #1193 

I will take care of it.

AndrewJackson2020:
> Another merge conflict... I guess after I merged #1193

@JelteF Fixed the merge conflict. I think that the currently failing test is a flaky test.

AndrewJackson2020:
Again, looks like the failure is a flaky test. I've noticed this seems to happen more on windows than other platforms.

PR #1140, opened 2024-08-11
Print current count of active clients and max_client_conn for an error

achanda:
This will make it easier to debug cases when pgbouncer rejects connections due to low max_client_conn


eulerto:
It is not a good idea to reveal server information to an unauthenticated session. This message is written to the logfile but also sent to the client. You didn't explain the reason but for troubleshooting, it is clear where the problem is. Postgres does the same thing with the `FATAL:  sorry, too many clients already` message that is written to the logfile and also sent to the client; no max_connections information in the message. I'm -1 on this PR.

JelteF:
I agree @eulerto. Thank you for the contribution, but I'll be closing this PR

PR #1146, opened 2024-08-23
Update documentation

AndrewJackson2020:
This PR removes references to load balancing being a todo item. I believe that #692 actioned this todo item.


AndrewJackson2020:
@JelteF Any issues with merging this PR? 

I know its a pretty minor issue but this documentation made me think that pgbouncer did not support load balancing and caused me to use another connection pooler in the past when pgbouncer would have worked just fine. I'm sure other people have had the same confusion as well.

PR #1147, opened 2024-08-25
Add kill_client command

AndrewJackson2020:
This PR adds a kill_client command to the admin console. This command is the equivalent of the Postgres pg_terminate_backend() function. Similar functionality also exists in other Postgres connection poolers.

In general this functionality is a useful last resort to undisciplined users opening a lot of connections and which block out other users. This is especially prominent with front end GUI DB applications that open a new session for every SQL editor tab.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -750,11 +750,13 @@ Immediately drop all client and server connections on given database.
 New client connections to a killed database will wait until **RESUME**
 is called.
 
-#### KILL_CLIENT [ptr]
+#### KILL_CLIENT ptr
 
 Immediately kill specificed client connection along with any server 
 connections for the given client. The client to kill, is identified
 by the ptr value that can be found using the SHOW CLIENTS command.
+
+An example command will look something like `KILL_CLIENT 0x561a71404330`
 
 #### SUSPEND
 
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -750,6 +750,12 @@ Immediately drop all client and server connections on given database.
 New client connections to a killed database will wait until **RESUME**
 is called.
 
+#### KILL_CLIENT [ptr]
+
+Immediately kill specificed client connection along with any server 
+connections for the given client. The client to kill, is identified
+by the ptr value that can be found using the SHOW CLIENTS command.
+
 #### SUSPEND
 
 All socket buffers are flushed and PgBouncer stops listening for data on them.
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -752,11 +752,21 @@ is called.
 
 #### KILL_CLIENT ptr
 
-Immediately kill specificed client connection along with any server 
+Immediately kill specificed client connection along with any server
 connections for the given client. The client to kill, is identified
-by the ptr value that can be found using the SHOW CLIENTS command.
+by the ptr value that can be found using the `SHOW CLIENTS` command.
 
-An example command will look something like `KILL_CLIENT 0x561a71404330`
+An example command will look something like `KILL_CLIENT 0x561a71404330`.
+
+Please note that memory addreses are reused and it is possible to kill
+an unintended client. For example if you run `SHOW CLIENTS`, grab a client
+ptr, that client disconnects, a new client connects, then you run `KILL_CLIENT`
+there is a high likelihood that the new client will have the old clients
+ptr and the new client will be disconnected. As such this command should
+only be used on a client who you are certain will not disconnect by themselves
+and should be run immediately after getting the ptr in `SHOW CLIENTS`, ideally
+in an automated script. This is a limitation that we are looking to addres
+in the future.
 
 #### SUSPEND
 
--- a/include/admin.h
+++ b/include/admin.h
@@ -26,5 +26,3 @@ bool admin_flush(PgSocket *admin, PktBuf *buf, const char *desc) /* _MUSTCHECK *
 bool admin_ready(PgSocket *admin, const char *desc)  _MUSTCHECK;
 void admin_handle_cancel(PgSocket *client);
 void admin_cleanup(void);
-PgSocket *find_client_global(const char *arg);
-PgSocket *find_socket_in_list(const char *arg, struct StatList *sockets);
--- a/include/admin.h
+++ b/include/admin.h
@@ -26,3 +26,5 @@ bool admin_flush(PgSocket *admin, PktBuf *buf, const char *desc) /* _MUSTCHECK *
 bool admin_ready(PgSocket *admin, const char *desc)  _MUSTCHECK;
 void admin_handle_cancel(PgSocket *client);
 void admin_cleanup(void);
+PgSocket *find_client_global(const char *arg);
+PgSocket *find_socket_in_list(const char *arg, struct StatList *sockets);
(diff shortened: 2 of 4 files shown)


include/admin.h
```diff
@@ -26,3 +26,5 @@ bool admin_flush(PgSocket *admin, PktBuf *buf, const char *desc) /* _MUSTCHECK *
 bool admin_ready(PgSocket *admin, const char *desc)  _MUSTCHECK;
 void admin_handle_cancel(PgSocket *client);
 void admin_cleanup(void);
+PgSocket *find_client_global(const char *arg);
+PgSocket *find_socket_in_list(const char *arg, struct StatList *sockets);
```
JelteF, line 30: These two functions can be `static` afaict. When that's changed they wouldn't need to be in the header file.

src/admin.c
```diff
@@ -1330,6 +1330,83 @@ static bool admin_cmd_enable(PgSocket *admin, const char *arg)
...
+{
+	struct List *item;
+	int j;
+	int string_comp;
+	char buffer[strlen(arg) + 1];
+	PgSocket *socket;
+
+	statlist_for_each(item, sockets) {
+		socket = container_of(item, PgSocket, head);
+		j = snprintf(buffer, strlen(arg) + 1, "%p", socket);
```
JelteF, line 1345: I think it would be more efficient to parse the pointer value from the input and compare that to address of the PgSocket. Instead of formatting every client pointer and string comparing them.

P.S. I wouldn't care much about this in most cases, but we're calling this function for every client, so a bit of efficiency seems nice.

JelteF (commented):
This feature Idea seems reasonable to me. But needs some documentation.

include/admin.h
AndrewJackson2020: Fixed this in most recent commit.

src/admin.c
AndrewJackson2020: I agree this solution is cleaner. Have a solution that works on all of the linux test cases. Fails on windows for some reason. Don't have a windows dev environment set up for pgbouncer though.

AndrewJackson2020:
> This feature seems reasonable to me. But needs some documentation.

Added some documentation in the usage.md file. Mostly boilerplate from the kill db command.

src/admin.c
AndrewJackson2020: Okay I got it working now.

src/admin.c
```diff
@@ -1330,6 +1330,79 @@ static bool admin_cmd_enable(PgSocket *admin, const char *arg)
...
+		}
+	}
+	return NULL;
+}
+
+/* Command: KILL_CLIENT */
+static bool admin_cmd_kill_client(PgSocket *admin, const char *arg)
+{
+	PgSocket *kill_client;
+	PgSocket *target_client = NULL;
```
JelteF, line 1395: nit: I would make this a `void *` to make it clear that it should not be dereferenced (which could result in undefined behavior)

doc/usage.md
```diff
@@ -750,6 +750,12 @@ Immediately drop all client and server connections on given database.
 New client connections to a killed database will wait until **RESUME**
 is called.
 
+#### KILL_CLIENT [ptr]
+
+Immediately kill specificed client connection along with any server 
+connections for the given client. The client to kill, is identified
+by the ptr value that can be found using the SHOW CLIENTS command.
```
JelteF, line 757: I think this would benefit from a tiny example, to make it very clear what the pointer value is expected to look like.

src/admin.c
```diff
@@ -1330,6 +1330,79 @@ static bool admin_cmd_enable(PgSocket *admin, const char *arg)
...
+
+	sscanf(arg, "%p", &target_client);
```
JelteF, line 1397: Let's check the return value of `sscanf` and differentiate in the returned error message when the user uses an incorrect formatted value vs an unknown pointer value.

doc/usage.md
AndrewJackson2020: Agreed, non C devs who are pgbouncer end users might not understand or care what a pointer is.

AndrewJackson2020: Addressed in last commit. Added another line with an example in a `` block. I think memory addresses are different lengths in windows vs linux but I think it should be apparent enough what to look for with this line.

src/admin.c
AndrewJackson2020: Addressed in last commit. Added  a unit test for this as well so now there are 3 unit tests associated with this PR.

src/admin.c
AndrewJackson2020: Addressed in last commit

AndrewJackson2020:
@JelteF I believe that all of your feedback should be addressed in the last commit. Let me know if there is anything else you notice.

JelteF:
While playing around with this a bit I found that pointer reuse is very common. i.e. when you connect and disconnect, both clients will have the same pointer. This makes sense because we use a `Slab` for allocation, so a "free" isn't really freeing anything and the piece of memory will be reused for a future client.

That does make pointers a rather bad identifier for clients over time though. This has the downside for this feature that if you're unlucky, and the client disconnects on its own while you are trying to kill it, it's pretty likely that you will kill the next client that tries to authenticate by accident.

@eulerto do you have any thoughts on this? I'm leaning towards merging this PR as-is. Creating a whole new identifier for clients (e.g. an incrementing uint64, or even a uuid) seems a bit excessive just for the functionality of killing them. Especially since people will most likely kill clients that won't disconnect on their own. And if the race condition does happen, chances are it is not the end of the world that the other new connection is killed instead.

eulerto:
> While playing around with this a bit I found that pointer reuse is very common. i.e. when you connect and disconnect, both clients will have the same pointer. This makes sense because we use a `Slab` for allocation, so a "free" isn't really freeing anything and the piece of memory will be reused for a future client.
> 
> That does make pointers a rather bad identifier for clients over time though. This has the downside for this feature that if you're unlucky, and the client disconnects on its own while you are trying to kill it, it's pretty likely that you will kill the next client that tries to authenticate by accident.
> 

That's a weak identifier. I'm afraid you are killing legit client if when you are typing, the app closed that client connection and what you killed was indeed another new client. :(

> @eulerto do you have any thoughts on this? I'm leaning towards merging this PR as-is. Creating a whole new identifier for clients (e.g. an incrementing uint64, or even a uuid) seems a bit excessive just for the functionality of killing them. Especially since people will most likely kill clients that won't disconnect on their own. And if the race condition does happen, chances are it is not the end of the world that the other new connection is killed instead.

I agree that it is excessive design. However, last add a caveat saying that ptr is regularly reused and you might kill another connection if the previous one terminated between SHOW CLIENTS and KILL_CLIENT. If we get various complains about this decision, it is good sign that we need to rethink this decision. Please add a comment (FIXME) about it in the code so we are constantly reminded that it can be an issue.

I didn't review this PR but AFAICS it seems in good shape.

AndrewJackson2020:
Added a TODO comment above the kill_client command. Should a warning about this issue also be included in the usage.md documentation as well?

JelteF:
> Should a warning about this issue also be included in the usage.md documentation as well?

Yeah, let's do that

AndrewJackson2020:
@JelteF I added some language in usage.md that explains the risk. Let me know you want any other changes before this gets merged.

AndrewJackson2020:
Just opened #1172 to add a unique identifier to clients if we decide that this is something we want to address instead of relying on pointer addresses now or in the future.

PR #1148, opened 2024-08-26
Add support for Type=notify-reload with systemd

tie:
This PR adds support for using pgbouncer with `Type=notify-reload` systemd service.

See https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html

Note that `Type=notify-reload` services do not need `ExecReload=` command and use `ReloadSignal=` that conveniently defaults to SIGHUP.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -565,10 +565,31 @@ static void handle_sigusr2(int sock, short flags, void *arg)
 	}
 }
 
+/*
+ * Notify systemd that we are reloading, including a CLOCK_MONOTONIC timestamp in usec
+ * so that the program is compatible with a Type=notify-reload service.
+ *
+ * See https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html
+ *
+ * Note that get_cached_time/get_time_usec from libusual uses gettimeofday
+ * instead of a monotonic clock, otherwise this function would be a single line:
+ *     sd_notifyf(0, "RELOADING=1\nMONOTONIC_USEC=%" PRIu64, get_cached_time());
+ */
+static void notify_reloading(void)
+{
+#ifdef USE_SYSTEMD
+	struct timespec ts;
+	usec_t usec;
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	usec = (usec_t)ts.tv_sec * USEC + (usec_t)ts.tv_nsec / (usec_t)1000;
+	sd_notifyf(0, "RELOADING=1\nMONOTONIC_USEC=%" PRIu64, usec);
+#endif
+}
+
 static void handle_sighup(int sock, short flags, void *arg)
 {
 	log_info("got SIGHUP, re-reading config");
-	sd_notify(0, "RELOADING=1");
+	notify_reloading();
 	load_config();
 	if (!sbuf_tls_setup())
 		log_error("TLS configuration could not be reloaded, keeping old configuration");
--- a/src/main.c
+++ b/src/main.c
@@ -566,14 +566,10 @@ static void handle_sigusr2(int sock, short flags, void *arg)
 }
 
 /*
- * Notify systemd that we are reloading, including a CLOCK_MONOTONIC timestamp in usec
- * so that the program is compatible with a Type=notify-reload service.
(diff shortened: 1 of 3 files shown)


eulerto:
I'm afraid we cannot accept this PR. This service type (`notify-reload`) was added in [version 253](https://github.com/systemd/systemd/releases/tag/v253). There are still supported operating systems that use old `systemd` versions like Ubuntu focal (245), Ubuntu jammy (249), Rocky 8 (239) and Rocky 9 (252). Once the main distributions support this feature we can merge this PR.

tie:
@eulerto, sorry, I don’t understand your point. This PR makes a single change that could be incompatible with older systemd versions, that is, updates Type= in `etc/pgbouncer.service`. If some distributions use the provided example unit as-is, we can document that pgbouncer supports `Type=notify-reload` in the comment above instead and update it later. `sd_notifyf` has been available for decades and `Type=notify` should simply ignore unknown fields like `MONOTONIC_USEC`.

tie:
> etc/pgbouncer.service

And from what I understand, this file is used for reference/documentation purposes only. That is, it is installed under `docdir` and other distributions ship their own service definitions, e.g.
- Debian: https://salsa.debian.org/postgresql/pgbouncer/-/blob/460febb15ab9f040e74da9b2d4023fe3bb5c8481/debian/pgbouncer.service
- (Ubuntu package is mostly derived from Debian)
- Ubuntu (focal): https://git.launchpad.net/ubuntu/+source/pgbouncer/tree/debian?h=applied/ubuntu/focal
- Ubuntu (jammy): https://git.launchpad.net/ubuntu/+source/pgbouncer/tree/debian?h=applied/ubuntu/jammy
- Not sure where Rocky Linux code is hosted.

JelteF:
I think the `sd_notifyf` call should be fine to merge. Assuming that it indeed does not break older systemd setups.

But I do agree with @eulerto and the default systemd config which people copy paste should work with at least the majority of systems that people are expected to run in production environments. Since both the latest RHEL and latest Ubuntu LTS don't even support this feature, currently is not the time for that.

I'd be okay with merging a change to the comments of the config file, mentioning what could be change for systemd 253+. So basically inverting the comment addition that you made.

JelteF:
To be clear, I do agree that most people will use the systemd file from their package management. But even that being the case, showing a systemd file in the docs that is broken for pretty much all systems that we expect people to run at the moment seems like a bad choice.

tie:
@JelteF, rebased and inverted the comment. I’ve also noticed that I’ve wrongly assumed that `get_cached_time`/`get_time_usec` from libusual uses `CLOCK_MONOTONIC` that is required for proper reload synchronization with systemd (currently [`get_time_usec`](https://github.com/libusual/libusual/blob/f24c5ea6abfb549dfec09c3bb2a3f1fbb99faaa6/usual/time.c#L68-L74) uses [`gettimeofday`](https://blog.habets.se/2010/09/gettimeofday-should-never-be-used-to-measure-time.html). I’ve added explicit `clock_gettime` call in this PR.

src/main.c
```diff
@@ -565,10 +565,31 @@ static void handle_sigusr2(int sock, short flags, void *arg)
...
+ *     sd_notifyf(0, "RELOADING=1\nMONOTONIC_USEC=%" PRIu64, get_cached_time());
+ */
+static void notify_reloading(void)
+{
+#ifdef USE_SYSTEMD
+	struct timespec ts;
+	usec_t usec;
+	clock_gettime(CLOCK_MONOTONIC, &ts);
+	usec = (usec_t)ts.tv_sec * USEC + (usec_t)ts.tv_nsec / (usec_t)1000;
+	sd_notifyf(0, "RELOADING=1\nMONOTONIC_USEC=%" PRIu64, usec);
```
eulerto, line 585: Does systemd ignore the MONOTONIC_USEC if the version is less than 253? AFAICS, yes. However, I didn't test.

eulerto (commented):
My main objection is related to the `pgbouncer.service` modification. It is just a comment, I'm fine.

src/main.c
tie: Yes, at least systemd 252 seems to be fine with this.

eulerto (approved):
LGTM

PR #1151, opened 2024-08-29
Add ability to authenticate a user against a specific database

var77:
This PR adds `auth_query_param_count` parameter which can be either `1` or `2`.
 
In case of `1` it will work as before sending binding auth_query with only username parameter, but if it will be set to `2` we will bind the `auth_query` with `username` ($1) and `dbname` ($2) parameters, thus adding ability to check the database in `auth_query` to which the client is trying to connect.

### Usecase

We at Lantern have a usecase where databases are being added dynamically, so under `[databases]` in config file we set `* = host = ..., port = 5432` . But we need to restrict each user to connect to a single database. 

There are currently 2 options to achieve this functionality:
- Using `pg_hba` file and setting `auth_type` to `hba` , but this requires to somehow add and remove lines to `pg_hba` file each time new database is created and deleted and `RELOAD` pgbouncer. This approach also adds additional overhead on backup & restore of the database
- Removing `auth_dbname` config and letting the `auth_db` to fallback to the database user is trying to connect. This approach adds ability to get the database name by using `current_database()` function in the query, but then you can never connect to pgbouncer console anymore as it tries to set the auth_db to pgbouncer and errors as it is forbidden.

This PR will allow to use a central database for auth, table mapping of (rolname , dbname) and use an `auth_query` where we can check if there exists a record in the table where `rolname=$1 and dbname=$2`

### Notes
- When connecting to database directly the pgbouncer auth flow will be bypassed, so in the case described above it will be needed to expose only pgbouncer
- When trying to connect to another database using `\c $dbname` the auth flow will be the same and `auth_query` will prevent connecting to the database if needed.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -496,6 +496,16 @@ is used, it needs to be installed into each database.
 
 Default: `SELECT usename, passwd FROM pg_shadow WHERE usename=$1`
 
+### auth_query_param_count
+
+Parameter count to bind for `auth_query`
+
+Possible values are 1 and 2.
+If the value is set to 1 only username will be passed as $1,
+if the value will be set to 2 dbname will be passed as $2 param.
+
+Default: 1
+
 ### auth_dbname
 
 Database name in the `[database]` section to be used for authentication purposes. This
--- a/etc/pgbouncer.ini
+++ b/etc/pgbouncer.ini
@@ -132,6 +132,10 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; must have 2 columns - username and password hash.
 ;auth_query = SELECT usename, passwd FROM pg_shadow WHERE usename=$1
 
+;; Number of parameters passed to auth_query
+;; 1: (only username will be passed), 2: (username and dbname will be passed)
+;auth_query_param_count = 1
+
 ;; Authentication database that can be set globally to run "auth_query".
 ;auth_dbname =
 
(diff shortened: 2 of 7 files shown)


AndrewJackson2020:
Have you considered using grant connect/revoke connect on the postgres side?

https://www.postgresql.org/docs/current/auth-pg-hba-conf.html

"To connect to a particular database, a user must not only pass the pg_hba.conf checks, but must have the CONNECT privilege for the database. If you wish to restrict which users can connect to which databases, it's usually easier to control this by granting/revoking CONNECT privilege than to put the rules in pg_hba.conf entries."

var77:
> Have you considered using grant connect/revoke connect on the postgres side?
> 
> 
> 
> https://www.postgresql.org/docs/current/auth-pg-hba-conf.html
> 
> 
> 
> "To connect to a particular database, a user must not only pass the pg_hba.conf checks, but must have the CONNECT privilege for the database. If you wish to restrict which users can connect to which databases, it's usually easier to control this by granting/revoking CONNECT privilege than to put the rules in pg_hba.conf entries."

Thanks for the suggestion @AndrewJackson2020 this seems to cover the use-case! 

I have tried it previously, but somehow missed to run `REVOKE CONNECT ON DATABASE dbname FROM PUBLIC` before granting connect to corresponding user.

I will close the PR.

PR #1152, opened 2024-08-29
Attempt to fix inconsistent error

### LINKED ISSUE #1149: Inconsistent Error Message
I am seeing an inconsistent error message when trying to log in as a user with nologin. basically the below test case failed on the last pytest.raises below. Would expect the error message to show 'is not permitted to login' as it does after the first attempt to login. Is this intended behavior?

It looks like the reason for this behaviour is the if statement `if (!pool->last_login_failed)` in the function test_fast_fail. Simply returning true  and skipping the subsequent application logic does not cause any other test failures and fixes the below test case. Not sure if there is some untested behavior that the rest of this function is needed for though.

```python
def test_message(bouncer, pg):
  test_user = "maxedout"
  test_database = "p0a"
  # Connect to database as User, creates existing pool
  conn_1 = bouncer.conn(user=test_user, dbname=test_database)

  # Change user to nologin
  pg.sql(f"ALTER USER {test_user} WITH NOLOGIN;")

  # Kill process on postgres
  terminate_string = f"""
  SELECT pg_terminate_backend(pid)
  FROM pg_stat_activity
  WHERE
    pid <> pg_backend_pid()
    AND usename = '{test_user}'
  """
  pg.sql(terminate_string)

  # login, check error message
  with pytest.raises(psycopg.OperationalError, match=r"is not permitted to log in"):
    bouncer.test(user=test_user, dbname=test_database)

  # login again, check error message
  # TODO This assert fails, No longer raises 'is not permitted to log in', why?
  with pytest.raises(psycopg.OperationalError, match=r"is not permitted to log in"):
    bouncer.test(user=test_user, dbname=test_database)
```

AndrewJackson2020:
This changes `server_login_retry` to also cache the error message. This makes it much easier to debug an authentication problem between PgBouncer and Postgres, because now the client always sees the message, not just the first time. Before you'd have to check the PgBouncer logs to find the actual error message.

Fixes #1149


JelteF:
I think we'd still want the fast fail logic, to be able to quickly surface an error to all waiting clients. But we'd want want that error to contain the original error message too, in addition to the error about the fast failure being triggered.

AndrewJackson2020:
Yeah I think that makes sense. Will figure out how to make that happen.

AndrewJackson2020:
Just pushed a new commit that restores the fast fail logic but saves the previous error message to present to clients. Let me know if this is more in the right direction.

src/objects.c
```diff
@@ -869,7 +869,7 @@ bool check_fast_fail(PgSocket *client)
 		return true;
 
 	/* Else we fail the client. */
-	disconnect_client(client, true, "server login has been failing, try again later (server_login_retry)");
+	disconnect_client(client, true, "%s", pool->last_connect_failed_message);
```
JelteF, line 872:
```suggestion
	disconnect_client(client, true, "server login has been failing, cached error: %s (server_login_retry)", pool->last_connect_failed_message);
```

src/objects.c
```diff
@@ -1342,6 +1342,8 @@ void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...
 		if (!server->ready) {
 			server->pool->last_login_failed = true;
 			server->pool->last_connect_failed = true;
+			safe_strcpy(server->pool->last_connect_failed_message, reason, sizeof(server->pool->last_connect_failed_message));
+			// server->pool->last_connect_failed_message = (char*)reason;
```
JelteF, line 1346:
```suggestion
```

JelteF (approved):
I think this is good overall. For the FreeBSD failure, feel free to just get it to pass by skipping the test. The error reporting on FreeBSD to the client is known to be bad/broken in the test suite.

AndrewJackson2020:
> I think this is good overall. For the FreeBSD failure, feel free to just get it to pass by skipping the test. The error reporting on FreeBSD to the client is known to be bad/broken in the test suite.

Changed the test so that it skips FreeBSD. I'm still seeing FreeBSD fail with the message `ld-elf.so.1: /lib/libc.so.7: version FBSD_1.8 required by /usr/local/libexec/cabal/pandoc not found` on this PR, #1138, and #1137.

eulerto:
Indeed. I asked Cirrus Labs about the FreeBSD image and it seems we are using an EOL image. I created ande merged the PR #1183 to fix it.

AndrewJackson2020:
> Indeed. I asked Cirrus Labs about the FreeBSD image and it seems we are using an EOL image. I created ande merged the PR #1183 to fix it.
> 
> Revert the skip and fix the other tests.

Can confirm that the image update fixed #1138 and #1137. Still seeing some flakiness on some of the test runs in this PR though.

AndrewJackson2020:
@JelteF I believe that I was able to get the tests to work on test/test_auth.py::{test_change_server_password_reconnect,test_change_server_password_server_lifetime} by adding a sleep statement at the end. Possibly that test needed more time in order for the final error message to hit the log before the `log_contains` could register it. Either that or the test is flaky for some reason. 

Either way let me know if this needs more work or it looks good to you.

JelteF (approved):

Issue #1145, opened 2024-08-22
Queueing on unrelated databases after pause/resume


jcoleman:
We use PAUSE/RESUME (with a settings change and RELOAD in between) to atomically cutover between database hosts.

We've been noticing that we regularly see substantial queuing on database A after we've paused and resumed database B (yes, we're using `PAUSE <database_name>` not a global pause). We've eliminated applications side issues like "had a transaction open on database A while trying to query against database B". The real headscratcher here is that we do _not_ see the queueing if we do not resume database B. Not resuming does, of course, cause other app problems (😁), but it suggests that perhaps there's something about reconnecting the pool for database B that's (substantially?) slowing down activity in other pools. We see this even if there's almost zero activity on database B, so this isn't purely a "lots of client connections are waiting on the paused database" problem. The pool size server side is quite small (~10), but there are thousands of client side connections across many databases.

We're not using PAM (I saw an issue that looked very much like this one...except it was a `usleep` in `pam.c`).

We're using TLS (so I imagine some CPU burn).

We're also currently stuck on 1.17 -- we're trying to upgrade but currently noting random crashes on the newest version, and we're trying to bisect to find the first broken release.

Any ideas on locking or CPU intensive code in the event loop that could cause this behavior?

AndrewJackson2020:
Regarding the crashes you were seeing I assume you are referring to v1.23.0. v1.23.1 was just released which should fix all of that, if not that would be good to know in a a separate issue.

cosgroveb:
We’ve seen the process mysteriously stop at a low frequency per-process (but at our scale a daily occurrence because we run many pgbouncers) on pgbouncer 1.22 first when attempting to upgrade from 1.17 to 1.22.

We haven’t been able to reproduce outside of high throughput deployed environments and while we can’t definitively rule out the environment yet it is notable that on otherwise identical containers (all the same bookworm base image) the lowest tagged release of pgbouncer that exhibits the issue in our envs has proven to be 1.20.1 whereas 1.19.1 is apparently green. We plan to bisect the revisions between the two next which is a slow process for us because of the infrequency and deploying new containers etc

AndrewJackson2020:
Is there an error message to that you are seeing with the process stopping? What you are talking about sounds different from the 1.23.1 patch.

cosgroveb:
That’s the baffling bit. No error message in the logs and that was with the logging level at a 2.

cosgroveb:
[...]

PR #1157, opened 2024-09-03
Don't recycle connections on RELOAD if TLS config is unchanged

cosgroveb:
This addresses #1145.

We run pgbouncers client-side that connect to 80+ databases. Separately we are addressing this bottleneck by running more pgbouncers client-side and striping them but in order to eliminate the unnecessary queuing that this kind of aggressive connection recycling can cause I propose taking a more judicious approach.

Essentially, I propose:

1. If going from TLS disabled => enabled, recycle.
2. If changing sslmode, recycle.
3. If changing TLS config, recycle.
4. Otherwise do nothing.

It's kind of gross that I have to inspect each member of the lengthy TLS config struct. Comments in-line.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -796,6 +796,12 @@ Default: not set
 
 ### client_tls_ca_file
 
+If the contents of this file are changed without changing the actual
+setting, the new CA file will be used for new connections. Existing
+connections won't be closed though. If it's necessary for security
+reasons that connections using the old CA are recycled ASAP, it's
+advised to run RECONNECT after the reload.
+
 Root certificate file to validate client certificates.
 
 Default: not set
@@ -869,6 +875,12 @@ verify-full
 ### server_tls_ca_file
 
 Root certificate file to validate PostgreSQL server certificates.
+
+If the contents of this file are changed without changing the actual
+setting, the new CA file will be used for new connections. Existing
+connections won't be closed though. If it's necessary for security
+reasons that connections using the old CA are recycled ASAP, it's
+advised to run RECONNECT after the reload.
 
 Default: not set
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -754,6 +754,16 @@ Default: empty (use operating system defaults)
 
 ## TLS settings
 
+If the contents of any of the cert or key files are changed without
+changing the actual setting filename in the config, the new file
+contents will be used for new connections after a RELOAD. Existing
+connections won't be closed though. If it's necessary for security
+reasons that all connections start using the new files ASAP, it's
+advised to run RECONNECT after the RELOAD.
+
+Changing any TLS settings will trigger a RECONNECT automatically
+for security reasons.
+
 ### client_tls_sslmode
 
 TLS mode to use for connections from clients.  TLS connections
@@ -795,12 +805,6 @@ Certificate for private key.  Clients can validate it.
 Default: not set
 
 ### client_tls_ca_file
-
-If the contents of this file are changed without changing the actual
-setting, the new CA file will be used for new connections. Existing
-connections won't be closed though. If it's necessary for security
-reasons that connections using the old CA are recycled ASAP, it's
-advised to run RECONNECT after the reload.
 
 Root certificate file to validate client certificates.
 
@@ -875,12 +879,6 @@ verify-full
 ### server_tls_ca_file
 
 Root certificate file to validate PostgreSQL server certificates.
-
-If the contents of this file are changed without changing the actual
-setting, the new CA file will be used for new connections. Existing
-connections won't be closed though. If it's necessary for security
-reasons that connections using the old CA are recycled ASAP, it's
-advised to run RECONNECT after the reload.
 
 Default: not set
 
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -1255,20 +1255,20 @@ static bool tls_config_changed(struct tls_config *new_server_connect_conf)
 		(new_server_connect_conf->verify_time == server_connect_conf->verify_time));
 }
 
-static bool skip_tag_pools_dirty(struct tls_config *new_server_connect_conf)
+static bool tls_change_requires_reconnect(struct tls_config *new_server_connect_conf)
 {
 	if (server_connect_sslmode != cf_server_tls_sslmode) {
 		log_noise("new server_tls_sslmode detected");
-		return false;
+		return true;
 	} else if (server_connect_conf == NULL) {
 		log_noise("no existing server tls config detected");
-		return false;
+		return true;
 	} else if (tls_config_changed(new_server_connect_conf)) {
 		log_noise("no server tls config change detected");
-		return true;
+		return false;
 	} else {
 		log_noise("server tls config change detected");
-		return false;
+		return true;
 	}
 }
 
@@ -1357,7 +1357,7 @@ bool sbuf_tls_setup(void)
 	 * connections indefinitely. If TLS is disabled, and it was disabled before
 	 * as well then recycling connections is not necessary, since we know none
 	 * of the settings have changed. */
-	if ((server_connect_conf || new_server_connect_conf) && !skip_tag_pools_dirty(new_server_connect_conf)) {
+	if ((server_connect_conf || new_server_connect_conf) && tls_change_requires_reconnect(new_server_connect_conf)) {
 		struct List *item;
 		PgPool *pool;
 		statlist_for_each(item, &pool_list) {
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -1240,15 +1240,13 @@ static bool tls_config_changed(struct tls_config *new_server_connect_conf)
 {
 	return (strings_equal(new_server_connect_conf->ca_file, server_connect_conf->ca_file) &&
 		strings_equal(new_server_connect_conf->ca_path, server_connect_conf->ca_path) &&
-		strings_equal(new_server_connect_conf->ca_mem, server_connect_conf->ca_mem) &&
-		(new_server_connect_conf->ca_len == server_connect_conf->ca_len) &&
+		memcmp(new_server_connect_conf->ca_mem, server_connect_conf->ca_mem, new_server_connect_conf->ca_len) == 0 &&
 		strings_equal(new_server_connect_conf->ciphers, server_connect_conf->ciphers) &&
 		(new_server_connect_conf->ciphers_server == server_connect_conf->ciphers_server) &&
 		(new_server_connect_conf->dheparams == server_connect_conf->dheparams) &&
 		(new_server_connect_conf->ecdhecurve == server_connect_conf->ecdhecurve) &&
 		strings_equal(new_server_connect_conf->ocsp_file, server_connect_conf->ocsp_file) &&
-		strings_equal(new_server_connect_conf->ocsp_mem, server_connect_conf->ocsp_mem) &&
-		(new_server_connect_conf->ocsp_len == server_connect_conf->ocsp_len) &&
+		memcmp(new_server_connect_conf->ocsp_mem, server_connect_conf->ocsp_mem, new_server_connect_conf->ocsp_len) == 0 &&
 		(new_server_connect_conf->protocols == server_connect_conf->protocols) &&
 		(new_server_connect_conf->verify_cert == server_connect_conf->verify_cert) &&
 		(new_server_connect_conf->verify_client == server_connect_conf->verify_client) &&
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -1259,7 +1259,10 @@ static bool tls_config_changed(struct tls_config *new_server_connect_conf)
 
 static bool skip_tag_pools_dirty(struct tls_config *new_server_connect_conf)
 {
-  if (tls_config_changed(new_server_connect_conf)) {
+  if (server_connect_conf == NULL) {
+    log_noise("no existing server tls config detected");
+    return false;
+  } else if (tls_config_changed(new_server_connect_conf)) {
     log_noise("no server tls config change detected");
     return true;
   } else {
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -1243,7 +1243,7 @@ static bool tls_change_requires_reconnect(struct tls_config *new_server_connect_
 	} else if (server_connect_conf == NULL) {
 		log_noise("no existing server tls config detected");
 		return true;
-	} else if (tls_configs_equal(new_server_connect_conf, server_connect_conf)) {
+	} else if (tls_config_equal(new_server_connect_conf, server_connect_conf)) {
 		log_noise("no server tls config change detected");
 		return false;
 	} else {
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -1236,7 +1236,7 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
 	return true;
 }
 
-static bool tls_config_changed(struct tls_config *new_server_connect_conf)
+static bool tls_config_unchanged(struct tls_config *new_server_connect_conf)
 {
 	return (strings_equal(new_server_connect_conf->ca_file, server_connect_conf->ca_file) &&
 		strings_equal(new_server_connect_conf->ca_path, server_connect_conf->ca_path) &&
@@ -1263,7 +1263,7 @@ static bool tls_change_requires_reconnect(struct tls_config *new_server_connect_
 	} else if (server_connect_conf == NULL) {
 		log_noise("no existing server tls config detected");
 		return true;
-	} else if (tls_config_changed(new_server_connect_conf)) {
+	} else if (tls_config_unchanged(new_server_connect_conf)) {
 		log_noise("no server tls config change detected");
 		return false;
 	} else {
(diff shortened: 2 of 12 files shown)


src/sbuf.c
```diff
@@ -29,6 +29,7 @@
 #include <usual/safeio.h>
 #include <usual/slab.h>
 #include <usual/mbuf.h>
+#include <usual/tls/tls_internal.h>
```
cosgroveb, line 32: The fact that I need to include a header from libusual with "internal" in the name is a smell and a sign that maybe the logic to test whether a new TLS config structure matches an existing struct should actually live in there.

src/sbuf.c
```diff
@@ -1235,6 +1236,44 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
 	return true;
 }
 
+static bool tls_config_changed(struct tls_config *new_server_connect_conf)
+{
+  return (strings_equal(new_server_connect_conf[0].ca_file, server_connect_conf->ca_file) &&
```
cosgroveb, line 1241: I wish C had reflection :) Alas. In any case, there are two members/structs I didn't check here worth noting. [`tls_error`](https://github.com/libusual/libusual/blob/master/usual/tls/tls_internal.h#L50) which seems to be for internal handling of TLS setup errors. Also, [`tls_keypair`](tls_keypair). keypair appears to be a list of keys. 

Maybe it should be checked? I would need to iterate through both the new and the old which is fine.

cosgroveb:
3 out of 4 of my new tests fail in CI under a `--without-openssl`build :| 

```FAILED test/test_operations.py::test_servers_disconnect_when_changing_sslmode - AssertionError
FAILED test/test_operations.py::test_servers_disconnect_when_changing_tls_config - AssertionError
FAILED test/test_operations.py::test_servers_disconnect_when_enabling_ssl - AssertionError
======= 3 failed, 144 passed, 9 skipped, 9 warnings in 96.52s (0:01:36) ========
```

Maybe I'm missing a marker/annotation. 👀

cosgroveb:
Okay, moved my tests into `test_ssl` so they only run in the right builds and addressed the uncrustify build.

src/sbuf.c
```diff
@@ -1235,6 +1236,44 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
 	return true;
 }
 
+static bool tls_config_changed(struct tls_config *new_server_connect_conf)
```
jcoleman, line 1239: Is the struct sufficiently stable or only have binary comparable items in it such that we could simply mem compare? Or do we need to list out all of the attributes?

test/test_ssl.py
```diff
@@ -372,3 +373,65 @@ def test_ssl_replication(pg, bouncer, cert_dir):
...
+
+def test_servers_no_disconnect_on_reload_with_no_tls_change(bouncer):
+    bouncer.default_db = "pTxnPool"
+
+    with bouncer.cur() as cur:
+        # change nothing and RELOAD
+        bouncer.admin("RELOAD")
+
+        with bouncer.log_contains(
+            r"pTxnPool.*closing because: database configuration changed", 0
```
jcoleman, line 386: Maybe we could test that the list of backend connections itself (maybe measured from PG itself?) remains constant? The log message test to show it's 0 makes me a bit nervous about false green tests.

cosgroveb: That's a good idea. We can assert on both. The log assertions seemed to be conventional in this suite so I think there's some confidence in the approach but I like also asserting on connections from PG.

src/sbuf.c
```diff
@@ -1235,6 +1236,44 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+{
+	return (strings_equal(new_server_connect_conf[0].ca_file, server_connect_conf->ca_file) &&
```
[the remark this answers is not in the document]
cosgroveb: Cleaned up this syntax in [0adafa8](https://github.com/pgbouncer/pgbouncer/pull/1157/commits/0adafa86c808be2e7c1dcdb485a65b5e7c2a0679)

src/sbuf.c
cosgroveb: Interesting thought... I suspect that it's stable enough for that kind of comparison. I'll try it.

test/test_ssl.py
cosgroveb: Added [assertions](
f14b97d) to address this feedback

src/sbuf.c
cosgroveb: aha yeah @jcoleman and I chatted and this doesn't work with a couple pointers to other structs and strings in `tls_config`

src/sbuf.c
JelteF: Yeah, I think it's fine for initial review, but before merging the logic to compare a `tls_config` should be created as a PR of `libusual` instead (which can then be used in this PR).

src/sbuf.c
```diff
@@ -1235,6 +1236,44 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+		(new_server_connect_conf->ocsp_len == server_connect_conf->ocsp_len) &&
+		(new_server_connect_conf->protocols == server_connect_conf->protocols) &&
+		(new_server_connect_conf->verify_cert == server_connect_conf->verify_cert) &&
+		(new_server_connect_conf->verify_client == server_connect_conf->verify_client) &&
+		(new_server_connect_conf->verify_depth == server_connect_conf->verify_depth) &&
+		(new_server_connect_conf->verify_name == server_connect_conf->verify_name) &&
+		(new_server_connect_conf->verify_time == server_connect_conf->verify_time));
+}
+
+static bool skip_tag_pools_dirty(struct tls_config *new_server_connect_conf)
```
JelteF, line 1260: nit, let's flip this around to avoid a double negation in the actual comparison and call it something `tls_changes_requires_reconnect`

src/sbuf.c
```diff
@@ -1235,6 +1236,44 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
 	return true;
 }
 
+static bool tls_config_changed(struct tls_config *new_server_connect_conf)
+{
+	return (strings_equal(new_server_connect_conf->ca_file, server_connect_conf->ca_file) &&
+		strings_equal(new_server_connect_conf->ca_path, server_connect_conf->ca_path) &&
```
JelteF, line 1242: There's a nasty problem here with this comparison. It seems likely that for rotations people might change the contents cert/key while keeping the path the same. This comparison wouldn't detect that. I don't know if we can actually compare this (especially when considering race conditions). So I think this limitation is probably fine, but we should document this in the user facing documentation. So for the `server_tls_ca_file` and `client_tls_ca_file` docs add something like this:
> If the contents of this file are changed without changing the actual setting, the new CA file will be used for new connections. Existing connections won't be closed though. If it's necessary for security reasons that connections using the old CA are recycled ASAP, it's advised to run `RECONNECT` after the reload.

test/test_ssl.py
```diff
@@ -372,3 +373,73 @@ def test_ssl_replication(pg, bouncer, cert_dir):
...
     bouncer.psql("IDENTIFY_SYSTEM", **connect_args)
+
+
+def test_servers_no_disconnect_on_reload_with_no_tls_change(bouncer, pg, cert_dir):
+    bouncer.default_db = "pTxnPool"
+
+    with bouncer.cur() as cur:
+        assert pg.connection_count(dbname="p0") == 1
+        # change nothing and RELOAD
+        bouncer.admin("RELOAD")
```
JelteF, line 384: This should be in the next `log_contains` block (same for other tests)

test/test_ssl.py
```diff
@@ -372,3 +373,65 @@ def test_ssl_replication(pg, bouncer, cert_dir):
...
+    bouncer.admin("RELOAD")
+
+    with bouncer.cur() as cur:
+        bouncer.write_ini(f"server_tls_protocols = secure")
+        bouncer.admin("RELOAD")
+
+        with bouncer.log_contains(
+            r"pTxnPool.*closing because: database configuration changed", 1
+        ):
+            time.sleep(0.5)
```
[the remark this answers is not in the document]
JelteF: It indeed seems unnecessary to check this for each of these tests. But yeah let's keep one of them.

JelteF (changes requested):
Overall looks like a good improvement. Currently we indeed recycle connections way too aggressively when TLS is enabled.

src/sbuf.c
```diff
@@ -1235,6 +1236,44 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
...
+		strings_equal(new_server_connect_conf->ca_mem, server_connect_conf->ca_mem) &&
+		(new_server_connect_conf->ca_len == server_connect_conf->ca_len) &&
+		strings_equal(new_server_connect_conf->ciphers, server_connect_conf->ciphers) &&
+		(new_server_connect_conf->ciphers_server == server_connect_conf->ciphers_server) &&
+		(new_server_connect_conf->dheparams == server_connect_conf->dheparams) &&
+		(new_server_connect_conf->ecdhecurve == server_connect_conf->ecdhecurve) &&
```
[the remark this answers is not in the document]
cosgroveb: Ah yes, it was on purpose but not necessarily for "good" reasons. See https://github.com/pgbouncer/pgbouncer/pull/1157#discussion_r1741782953

It can be worked through perhaps in the libusual PR.

src/sbuf.c
cosgroveb: I like it thanks for the feedback on the double negation

test/test_ssl.py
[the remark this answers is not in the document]
cosgroveb: 0143b3d

test/test_ssl.py
cosgroveb: [fb9bed3](https://github.com/pgbouncer/pgbouncer/pull/1157/commits/fb9bed316a0cd7b5605ef849df2654c11034bada). 

In the other tests there are `RELOAD`s in the `log_contains`block. The second and third test have one before the `log_contains`but only as part of test setup and the assertion block also has a `RELOAD` for each of those.

src/sbuf.c
cosgroveb: This only checks the `server_tls_ca_file`, right?

cosgroveb: Ahh yep both are impacted by less aggressive connection recycling on `RELOAD` 11f099b

src/sbuf.c
```diff
@@ -1235,6 +1236,44 @@ static bool setup_tls(struct tls_config *conf, const char *pfx, int sslmode,
 	return true;
 }
 
+static bool tls_config_changed(struct tls_config *new_server_connect_conf)
+{
+	return (strings_equal(new_server_connect_conf->ca_file, server_connect_conf->ca_file) &&
+		strings_equal(new_server_connect_conf->ca_path, server_connect_conf->ca_path) &&
+		strings_equal(new_server_connect_conf->ca_mem, server_connect_conf->ca_mem) &&
+		(new_server_connect_conf->ca_len == server_connect_conf->ca_len) &&
```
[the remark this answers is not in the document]
cosgroveb: thanks fa1c177

src/sbuf.c
cosgroveb: [44b913b](https://github.com/pgbouncer/pgbouncer/pull/1157/commits/44b913b7131abfeb20c05c3e09d0f629d096eb4f)

test/test_ssl.py
cosgroveb: Oops. I misread my tests and was wrong. Fixed. [5fe717a](https://github.com/pgbouncer/pgbouncer/pull/1157/commits/5fe717a3b1fadac6469b0bfe6d3283960591e858)

src/sbuf.c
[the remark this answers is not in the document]
cosgroveb: @JelteF I added some code to do this and a test that fails on [another branch here](https://github.com/pgbouncer/pgbouncer/commit/3e3fca58df9b50281ab2307aa3016911dbbccdd2).

Breakpointing in gdb it would appear that these sub-structs are NULLed out/not hydrated by libusual on config load... on the branch linked above:

```
(gdb) break sbuf.c:1274
Breakpoint 3 at 0x55555557dda3: file src/sbuf.c, line 1274.
(gdb) c
Continuing.
2024-09-04 14:16:43.208 UTC [1003231] LOG S-0x55555566ab80: pTxnPool/bouncer@127.0.0.1:10201 new connection to server (from 127.0.0.1:37876)
2024-09-04 14:16:48.748 UTC [1003231] LOG RELOAD command issued

Breakpoint 3, tls_config_unchanged (new_server_connect_conf=0x55555567e930)
    at src/sbuf.c:1274
1274            return (strings_equal(new_server_connect_conf->ca_file, server_connect_conf->ca_file) &&
(gdb) print new_server_connect_conf[0].keypair[0]
$22 = {next = 0x0, cert_file = 0x0, cert_mem = 0x0, cert_len = 0, key_file = 0x0, key_mem = 0x0, key_len = 0}
(gdb) print server_connect_conf[0].keypair[0]
$23 = {next = 0x0, cert_file = 0x0, cert_mem = 0x0, cert_len = 0, key_file = 0x0, key_mem = 0x0, key_len = 0}
(gdb)```

It's possible that these only get values when libusual needs to read in those files and create these objects.

I'll add a code comment and at this point it's time to move the work over to libusual.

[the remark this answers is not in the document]
Comment (cosgroveb) [reply src/sbuf.c:1250]: [1c373c5](https://github.com/pgbouncer/pgbouncer/pull/1157/commits/1c373c58937bb5a376fba736cb093ceacaea3e81)

File: src/sbuf.c (hunk 1 above)
Comment (cosgroveb) [reply src/sbuf.c]: d6ff3a09a0ae832cdf1c643f39ee626517297059

PRed here https://github.com/libusual/libusual/pull/57

**Comment by cosgroveb** [comment]
Hey @JelteF I think I've addressed any outstanding feedback. I did force push the branch on Wednesday to rewrite my git ùser.email`. Sorry if that was bad form or made reviewing the final state of this harder.

I presume we need to wait for my outstanding [libusual PR](https://github.com/libusual/libusual/pull/57) to be reviewed/merged to get this pull merged? Any tips on getting the libusual PR reviewed? 

Thanks for being so responsive and helpful in getting this operational bottleneck in pgbouncer solved.

**Comment by JelteF** [comment]
> I presume we need to wait for my outstanding https://github.com/libusual/libusual/pull/57 to be reviewed/merged to get this pull merged? Any tips on getting the libusual PR reviewed?

All PgBouncer committers also have commit access on libusual (including me). So this should be fine. But yeah we should merge that one first.

File: doc/config.md [hunk 12]
```diff
@@ -796,6 +796,12 @@ Default: not set
 
 ### client_tls_ca_file
 
+If the contents of this file are changed without changing the actual
+setting, the new CA file will be used for new connections. Existing
+connections won't be closed though. If it's necessary for security
+reasons that connections using the old CA are recycled ASAP, it's
+advised to run RECONNECT after the reload.
```
JelteF, line 803: Let's move this paragraph to the top of the TLS settings section. Also let's change the wording to include RELOAD, because now it sounds like this will happen automatically even without one. Something like this:
```suggestion
If the contents of any of the cert or key files are changed without changing the actual
setting filename in the config, the new file contents will be used for new connections after a RELOAD. Existing
connections won't be closed though. If it's necessary for security
reasons that all connections start using the new files ASAP, it's
advised to run RECONNECT after the RELOAD.
```

JelteF: Probably also good to note that changing any of the settings in the config file will trigger a RECONNECT automatically.

cosgroveb: [5e27dd2](https://github.com/pgbouncer/pgbouncer/pull/1157/commits/5e27dd239c6f3ef22cf0dd379ab9b3ccdbb65ef4)

cosgroveb:
@JelteF I've [branched a cleanup from this](https://github.com/cosgroveb/pgbouncer/tree/move_string_checking_func_to_libusual) to remove `strings_equal()` and use the libusual library function as [discussed there ](https://github.com/libusual/libusual/pull/57#issuecomment-2336713415)and will PR that as soon as the libusual PR and this one are merged.

JelteF:
CI is reporting a failure on one of the newly added tests, looks like it's flaky since only one of the linux builds is hitting the issue.

cosgroveb:
I haven't had success with any specs at all with valgrind enabled locally today to repro but will give it another try tonight

cosgroveb:
Okay @JelteF these specs are genuinely flaky due to timing issues between `full_maint_period` refreshes clients checking out a connection. This spec was one that didn't sleep for `full_maint_period` and didn't expect that style of refresh from the logs but got one:

`2024-09-08 18:04:35.379 UTC [12865] LOG S-0x53ccd50: pTxnPool/bouncer@127.0.0.1:10201 closing because: database configuration changed (age=0s)`

This isn't incorrect behavior. Updating the expectation to avoid this: [e27c774](https://github.com/pgbouncer/pgbouncer/pull/1157/commits/e27c774b07e74372a5b82a5d9c2a235159d9f1dd)

cosgroveb:
Any further changes needed @JelteF ?

JelteF (approved):
LGTM now

PR #1166, opened 2024-09-15
Handle auth_query correctly for replication connections

### LINKED ISSUE #1164: logical replication and auth_query issues
Hello,
I saw that recently support for logical replication was added to 1.23.0 and I was trying to test it but I am having some issues with it. I have the following pgbouncer configuration:

```
[pgbouncer]
listen_addr               = 0.0.0.0
listen_port               = 5432
unix_socket_dir           = /tmp
user                      = postgres
auth_file                 = /etc/pgbouncer/users.txt
auth_type                 = md5
auth_user                 = pgbouncer
auth_query                = SELECT uname,phash FROM pgbouncer.user_lookup($1)
default_pool_size         = 20
pool_mode                 = session
max_client_conn           = 4000
server_idle_timeout       = 600
admin_users               = pgbouncer_admin
client_tls_sslmode        = allow
client_tls_key_file       = /etc/pgbouncer-certs/client/tls.key
client_tls_cert_file      = /etc/pgbouncer-certs/client/tls.crt

[databases]
dbname = host=hostname port=5432 dbname=dbname pool_size=10
```

When I try to connect to the database with the string:

```
postgres://username:password@hostname:5432/dbname?replication=database
```

I get the error (on the database side):

```
2024-09-12 08:38:43.147 GMT [124] FATAL:  permission denied to start WAL sender
2024-09-12 08:38:43.147 GMT [124] DETAIL:  Only roles with the REPLICATION attribute may start a WAL sender process.
```

I suspect it is because of the `auth_query` is trying to connect with the pgbouncer user to the database while forwarding the `replication=database` parameter.

If I add REPLICATION permissions to pgbouncer then I get the error that extended queries are not supported in replication connections:

```
2024-09-12 08:22:13.670 GMT [159] ERROR:  extended query protocol not supported in a replication connection
```

Do you know if/how this is supposed to work?

Thanks for the awesome work!

JelteF:
The way we send an auth_query is a bit strange. We temporarily set the
pool of the client to the pool for the auth_user, and then continue
connection startup like normal. This broke for replication connections
because we would send the replication parameter for that connection too.
This fixes that by only doing our special connection setup logic for
replication connections if we're not currently trying to send an
auth_query.

Fixes #1164

### DIFF
--- a/include/client.h
+++ b/include/client.h
@@ -21,4 +21,5 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 bool handle_auth_query_response(PgSocket *client, PktHdr *pkt);
 bool check_db_connection_count(PgSocket *client);
 bool check_user_connection_count(PgSocket *client);
+bool sending_auth_query(PgSocket *client);
 PgDatabase *prepare_auth_database(PgSocket *client) _MUSTCHECK;
--- a/src/client.c
+++ b/src/client.c
@@ -799,7 +799,12 @@ static bool set_startup_options(PgSocket *client, const char *options)
 		 * GUC_REPORT flag, specifically extra_float_digits which is a
 		 * configuration that is set by CREATE SUBSCRIPTION in the
 		 * options parameter.
+		 *
+		 * First free it, because set_startup_options might be called
+		 * multiple times in some cases. One of these being when
+		 * auth_user is enabled.
 		 */
+		free(client->startup_options);
 		client->startup_options = strdup(options);
 		if (!client->startup_options)
 			disconnect_client(client, true, "out of memory");
--- a/src/client.c
+++ b/src/client.c
@@ -164,6 +164,14 @@ static bool send_client_authreq(PgSocket *client)
 	return res;
 }
 
+/*
+ * Returns true if the currently trying to send an auth query to the server.
+ */
+bool sending_auth_query(PgSocket *client)
+{
+	return client->wait_for_user_conn || client->wait_for_user;
+}
+
 static void start_auth_query(PgSocket *client, const char *username)
 {
 	int res;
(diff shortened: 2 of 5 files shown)


dbarrosop:
Hi, how is it going? just wanted to check if there was any progress here as this is still not working as expected. Thanks!

eulerto (commented):
For some reason, the `test_physical_rep_pg_basebackup` is failing for me. There is no explicit errors in the log (Postgres and PgBouncer).

```
==================================================== short test summary info =====================================================
FAILED test/test_replication.py::test_physical_rep_pg_basebackup - Failed: Timeout >30.0s
=============================== 1 failed, 210 passed, 7 skipped, 13 warnings in 138.12s (0:02:18) ================================
make: *** [Makefile:162: check] Error 1

```

JelteF:
> For some reason, the test_physical_rep_pg_basebackup is failing for me. There is no explicit errors in the log (Postgres and PgBouncer).

Is this only with this PR, or also on master? I cannot reproduce it, and CI is running fine too. So I'm leaning towards merging this unless this is only happening on this PR for you.

Regarding debugging, did you uncomment the following line during the test? That should give much more info into what's happening.
https://github.com/pgbouncer/pgbouncer/blob/master/test/utils.py#L919

benchub:
FWIW, this issue was blocking us (because it turns out HUP'ing pgBouncer also resets connections which were made with credentials found in auth_file, and resetting logical replication connections on a regular basis is not great), but applying this patch seems to work well in our testing. 

With this patch, we can now successfully establish and maintain a logical replication session authenticated with auth_query, even while we HUP pgBouncer every few minutes.

d3qtian:
Hi @JelteF
Thank you for addressing this issue and providing a fix! We are currently in urgent need of this functionality for our project. Could you kindly let us know when this fix will be released in the next version of pgbouncer? Your update will help us plan accordingly.

Thanks again for the great work and support!

PR #1171, opened 2024-09-20
Add current timestamp to SHOW commands

eulerto:
Statistics does not provide "when" and SHOW is not flexible to include a new column. We usually get a timestamp in a script and attach it to each line of the SHOW output. Since it is a common pattern to have a timeseries for a monitoring tool or even for a specific analysis, the first commit just does it for some SHOW commands. Despite of the common requirement is for SHOW POOLS and SHOW STATS, I added the new column (now) for a bunch of SHOW commands that this new column makes sense.

I also added a separate commit to append milliseconds to the 'T' row type (type that prints date and time). This commit might break SHOW commands related to socket such as SHOW CLIENTS|SERVERS|SOCKETS. However, I don't expect that an application cannot parse a legitimate date and time format.


JelteF:
I think including milliseconds definitely makes a lot of sense, even at the slight risk of breaking someone. And I understand why you want the now timestamp, but it does add yet another column to every row for each SHOW comman. And this new column has has the exact same content for each row. I'm wondering if that clutter is worth it. For automated tooling it doesn't matter much, but when running `SHOW` commands manually it's not great.

eulerto:
The main problem is that SHOW is not flexible (like a SELECT that we can choose columns). I think about a new parameter, however, the SHOW command will be dependent of a parameter that might break some tools that parse SHOW commands.

When troubleshooting PgBouncer performance issues, we usually ask SHOW POOLS but there is no way to attach a date/time information for it (unless you externally include it). That's generally what tools do. Hence, you have a few interactions with a customer to collect the statistics and then generate some graphs based on the output. The "now" information is important to correlate events. The PgBouncer is generally in the same host as the Postgres server (no time drift issues) and this additional information is crucial for correlation.

That's true it is the same information for a bunch of rows. We generally don't expect thousands of rows from the SHOW commands. That's true some users have thousands of users and/or databases and all of them needs a pool and you have thousands of rows each time you get statistics. That's the design you choose and the repetitive information is a consequence of this design. There are a few tools that abstracts the date/time information using a "snapshot" table using an identifier to reference the pools table, however, you need another column in the pools table so I don't think what this design buys anything because now you need a join.

Another alternative is to add new SHOW TS_POOLS (where TS stands for timestamp). Every command that has the TS_ prefix contains the same columns as the original command plus a date/time information. I can live with it. The main drawback is that your monitoring tool has to implement support for a new class of SHOW commands.

I wouldn't like to hijack this PR but I'm planning to work on another PR that attack the problem (lack of statistics information) from another angle: have a new parameter to log the SHOW POOLS information into the log file. It basically will follow the same pattern as "LOG stats" that means a periodic dump of the pool statistics. In this case, you can control if you want such information into the log and have at least some information available to tune pool settings or investigate an issue in the past. This approach is complementary to the one proposed here because it might not be available in some cases. For example, cloud vendors don't generally provide access to the log files. The SQL interface can provide such information. (BTW, I've never checked if/how the popular cloud vendors provide the PgBouncer statistics.)

JelteF:
How about we start supporting commands like. So users can customize what they want returned.
```
SHOW POOLS INCLUDE_TIMESTAMP;
```

eulerto:
If we want to expand SHOW command to support options, my suggestion is something like

`SHOW POOLS [ ( option1 [value1], option2 [value2], ... ) ]`

It is easier to parse the options. The value is optional. It means use can use `TIMESTAMP` or `TIMESTAMP true` which means include the timestamp. The value can be a `string` or `number` too. We can also include new options to remove column(s) from the current column list.

I don't know how complicate will be a regex to capture this list of options. I wouldn't like to move to a real parser/lexer for this feature.

JelteF:
I'm fine with that option style. I don't think we need a big parser for this. If we only support string arguments and don't allow spaces in tokens, then we can just do this with a "split by string" tokenizer and doing strcmp.

PR #1173, opened 2024-09-23
Better handling of auth_file and better log messages

eskholm:
In the current handling of auth_file parsing, any error encountered during parsing, will stop further parsing and just run pgbouncer with whatever was read from auth_file until the error was encounted.

This PR will improve the handling, by logging what errors were encountered, skipping the offending lines and continuing trying to load the rest of the auth_file.

### DIFF
--- a/src/loader.c
+++ b/src/loader.c
@@ -639,6 +639,8 @@ static void disable_users(void)
 bool load_auth_file(const char *fn)
 {
 	char *user, *password, *buf, *p;
+	char msg[1000];
+	int n = 0;
 
 	/* No file to load? */
 	if (fn == NULL)
@@ -655,6 +657,9 @@ bool load_auth_file(const char *fn)
 
 	p = buf;
 	while (*p) {
+	
+		n++;
+	
 		/* skip whitespace and empty lines */
 		while (*p && isspace(*p)) p++;
 		if (!*p)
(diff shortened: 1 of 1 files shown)


eulerto:
The error messages need improvement. They are not informative enough to rapidly fix a parsing issue. You provide the line number but maybe it is a good idea to also add the `auth_file` path mainly because it is a parameter and someone may change it to another file.

Another problem with the current code is that it partially applies changes. IMO it should behave like [Postgres](https://github.com/postgres/postgres/blob/master/src/backend/libpq/hba.c#L2575) that parses the file and if there are no errors, it applies all changes. If there was any parse error, it keeps the old configuration and reports all encountered errors. That said, skipping the current line is making the problem even worse. It may apply changes incorrectly if, for example, someone adds a line break and the "new line" turns into a valid rule.

eskholm:
> The error messages need improvement. They are not informative enough to rapidly fix a parsing issue. You provide the line number but maybe it is a good idea to also add the `auth_file` path mainly because it is a parameter and someone may change it to another file.
> 
> Another problem with the current code is that it partially applies changes. IMO it should behave like [Postgres](https://github.com/postgres/postgres/blob/master/src/backend/libpq/hba.c#L2575) that parses the file and if there are no errors, it applies all changes. If there was any parse error, it keeps the old configuration and reports all encountered errors. That said, skipping the current line is making the problem even worse. It may apply changes incorrectly if, for example, someone adds a line break and the "new line" turns into a valid rule.

Great insights - and I think you're right.

I've refactored the code to parse everything, log errors when errors found, and if any errors found log a final line saying that the auth_file has not been loaded due to one or more parsing errors.

If no errors found, the auth_file gets applied. This is absolutely much safer. I've made the error messages more informative as well.

eulerto:
You are almost there... As I said it should parse the whole file and then apply changes. Your patch is still applying changes while parsing the file (see `unquote_add_authfile_user` inside the loop). You should save the information while parsing and after the loop, you create another loop to call `unquote_add_authfile_user` for every single saved information.

IMO we don't need that final extra message. It adds no additional information. Avoid applying unrelated changes (see first hunk and a new empty line).

Since you are improving the `load_auth_file`, add a comment at the top explaining how it works.

PR #1177, opened 2024-09-28
Add target_session_attrs support for PG 14+ servers

cosgroveb:
This is @dpirotte's work to add support for target_session_attrs, allowing pgbouncer to only connect to a database's host if it is in a given state, such as primary or standby.

The implementation relies on specific params from the server that are only passed starting with PG 14, and therefore it only supports PG 14 or later. On older versions, target_session_attrs has no effect.

This is complementary to the proposed `load_balance_hosts` and [was discussed](https://github.com/pgbouncer/pgbouncer/pull/736#discussion_r1222977478) in change in #736

### DIFF
--- a/src/varcache.c
+++ b/src/varcache.c
@@ -88,7 +88,7 @@ static void init_var_lookup_from_config(const char *cf_track_extra_parameters, i
 
 void init_var_lookup(const char *cf_track_extra_parameters)
 {
-	const char *names[] = { "DateStyle", "client_encoding", "TimeZone", "standard_conforming_strings", "application_name", "in_hot_standby", NULL };
+	const char *names[] = { "DateStyle", "client_encoding", "TimeZone", "standard_conforming_strings" "in_hot_standby", "default_transaction_read_only", NULL };
 	int idx = 0;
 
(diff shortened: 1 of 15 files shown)


cosgroveb:
Aha, yes. I jumped the gun and opened this without running the linter. Oof. Changing to draft until I work through the obvious CI failures.

cosgroveb:
I opened this before my morning coffee oof. Thought I had a build passing locally but I made some obvious mistakes in my rebase. 

I'll reopen this later this weekend. Don't want to waste anyone's time.

src/varcache.c
```diff
@@ -258,6 +258,8 @@ void varcache_apply_startup(PktBuf *pkt, PgSocket *client)
 		struct PStr *val = get_value(&client->vars, lk);
 		if (!val)
 			continue;
+		if (strcmp(lk->name, "in_hot_standby") == 0 || strcmp(lk->name, "default_transaction_read_only") == 0)
```
cosgroveb, line 261: This feature was first drafted before the implementation of replication connections. I'm unsure if this adjustment to satisfy test_replication.py is appropriate here, complete, and without edge-cases.

PR #1178, opened 2024-09-29
[admin-c] impl extended query protocol

goksgie:
This PR introduces an initial version of the extended query protocol support to the admin console.

The main driver to implement this protocol is to support Npgsql to connect PgBouncer's admin console, which is something we depend in our daily lives to extract some metrics from pgbouncer.

The current implementation has the following assumptions:
1. Queries that are supported by the PgBouncer's admin console is simple. I.e., we don't actually need to store procedures.
2. We only care about the command provided in the Parse command. And, due to (1), we only store this command and keep it until execution.

The contents have been tested manually using Npgsql.

Remaining TODOs:
* [ ] Add automated tests,
* [ ] See if RowDescription message needs to be returned for Describe packet, or whether it would be safe to skip.
* [ ] When should we drop the stored command, after Sync message is received?


src/admin.c
```diff
@@ -1697,6 +1739,71 @@ static bool admin_parse_query(PgSocket *admin, const char *q)
...
+	  	admin->admin_extended_protocol_cmd = malloc(sizeof(PgParsePacket));
+		if (!admin->admin_extended_protocol_cmd) {
+			admin_error(admin, "failed to allocate parse packet");
+			disconnect_client(admin, true, "out of memory");
+			return false;
+		}
+
+	 	if (!unmarshall_parse_packet(admin, pkt, admin->admin_extended_protocol_cmd)) 
+			return false;
+
```
JelteF, line 1773: should check if the statement has a name, and fail if that's the case. Let's not start supporting named prepared statements unless really necessary.

JelteF (changes requested):
Overall, I think this is a useful change. But there's definitely more work to be done here. 

On a side note, there's now a discord channel to discuss pgbouncer development that I'd recommend you to join: https://discordapp.com/channels/1258108670710124574/1300532992304742481

goksgie:
thanks @JelteF ! I'm experimenting with an acceptable approach for separating "describe" from query execution path. I'm trying to see if we can get away from course endless "if checks" based on some execution flag, but may not be the case.

Right now, I have this: https://github.com/goksgie/pgbouncer/tree/admin_console_extended_p_test, but I kind of don't like. Let me know if you have any suggestions that can speed up my procedure.

JelteF:
Discord invite link is here: https://discord.gg/XZy2DXj7Wz

PR #1179, opened 2024-10-02
Add target_session_attrs support for PG 14+ servers

cosgroveb:
This is @dpirotte's work to add support for target_session_attrs, allowing pgbouncer to only connect to a database's host if it is in a given state, such as primary or standby. 

The implementation relies on specific params from the server that are only passed starting with PG 14, and therefore it only supports PG 14 or later. On older versions, target_session_attrs has no effect.

This is complementary to the proposed load_balance_hosts and https://github.com/pgbouncer/pgbouncer/pull/736#discussion_r1222977478 in change in https://github.com/pgbouncer/pgbouncer/pull/736

Apologies for the PR noise. This has been a challenging rebase.


src/varcache.c
```diff
@@ -258,6 +258,8 @@ void varcache_apply_startup(PktBuf *pkt, PgSocket *client)
 		struct PStr *val = get_value(&client->vars, lk);
 		if (!val)
 			continue;
+		if (strcmp(lk->name, "in_hot_standby") == 0 || strcmp(lk->name, "default_transaction_read_only") == 0)
```
cosgroveb, line 261: This feature was first drafted before the implementation of replication connections. I'm unsure if this adjustment to satisfy test_replication.py is appropriate here, complete, and without edge-cases.

src/varcache.c
```diff
@@ -88,7 +88,7 @@ static void init_var_lookup_from_config(const char *cf_track_extra_parameters, i
 
 void init_var_lookup(const char *cf_track_extra_parameters)
 {
-	const char *names[] = { "DateStyle", "client_encoding", "TimeZone", "standard_conforming_strings", "application_name", NULL };
+	const char *names[] = { "DateStyle", "client_encoding", "TimeZone", "standard_conforming_strings", "application_name", "in_hot_standby", "default_transaction_read_only", NULL };
```
jimbobhickville, line 91: Brian and I were discussing this offline and wondered if we need to make 2 distinct lists. One for values that are readwrite and should be set on the connection startup, and another for readonly values that we want to be able to read from the connection but not attempt to set. Thoughts?

src/admin.c
```diff
@@ -882,7 +890,9 @@ static bool admin_show_pools(PgSocket *admin, const char *arg)
 	usec_t now = get_cached_time();
 	usec_t max_wait;
 	struct CfValue cv;
+	struct CfValue target_session_attrs_lookup;
 	int pool_mode;
+	const char *target_session_attrs_str;
 
 	cv.extra = pool_mode_map;
```
cosgroveb, line 897: Looks like we're missing a `target_session_attrs_lookup.extra = target_session_attrs_map;` here. Will push a fix. (ht to @jimbobhickville for noticing this)

JelteF:
Needs a rebase now that #736 is merged

cosgroveb:
I will spend more time on the current CI failures tomorrow. Sorry for the noise!

AndrewJackson2020:
This feature would be great. Question though: does this PR provide the client with the ability to supply a target_session_attr to pgbouncer and get a matching connection back? 

Based on skimming the code I believe the answer is no but want to be sure.

cosgroveb:
Hey @AndrewJackson2020 that's a very interesting question. The answer is no. In this feature, the pooler is the client in the sense of the emulation of this libpq feature. At a high level it is meant to allow the pool to be defined in a way that allows the pgbouncer operator to specify the [properties that a back-end session must have](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS) to be acceptable to the pool. Just as clients that are directly connecting to PostgreSQL servers can use this in conjunction with multiple host-names to select the first acceptable host the pooler can do that job instead.

I use this further in conjunction with `target_session_attrs=disable`. This provides more deterministic ordering of hosts and "stickiness" with multiple hosts in order to accomplish hassle-free failover/convergence from the pooler (and ultimately the client's perspective) when a standby is automatically promoted.

Hope that helps!

AndrewJackson2020:
> Hey @AndrewJackson2020 that's a very interesting question. The answer is no. In this feature, the pooler is the client in the sense of the emulation of this libpq feature. At a high level it is meant to allow the pool to be defined in a way that allows the pgbouncer operator to specify the [properties that a back-end session must have](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS) to be acceptable to the pool. Just as clients that are directly connecting to PostgreSQL servers can use this setting in libpq in conjunction with multiple host-names to select the first acceptable host the pooler can do that job instead with this feature.
> 
> I use this further in conjunction with `load_balance_hosts=disable`. This provides more deterministic ordering of hosts and "stickiness" with multiple hosts in order to accomplish hassle-free failover/convergence from the pooler (and ultimately the client's perspective) when a standby is automatically promoted.
> 
> Hope that helps!

It does. Looking further into libpq frontend interfaces I don't think my idea above is even possible as target session attribute is something that is handled on the client side. I don't think the target_session_attribute value is even passed from the client to the server.

cosgroveb:
> I don't think the target_session_attribute value is even passed from the client to the server.

Correct. It's a client option that affects the behavior of libpq but is not part of the session or passed to the server.

JelteF:
To be able to use target_session_attribute at the client I think you'd need this PR: https://github.com/pgbouncer/pgbouncer/pull/1205

AndrewJackson2020:
> To be able to use target_session_attribute at the client I think you'd need this PR: #1205
I think you w

cosgroveb:
I see that fixes from @greg-rychlewski reveal a minor test setup issue for me on the Windows build that the previous`pgk-config` failures were hiding. I'll work through that now but feel that this is ready for review.

doc/config.md
```diff
@@ -1285,6 +1285,32 @@ fast retries when multiple hosts are available.
...
 Default: `round-robin`
 
+### target_session_attrs
+
+Note: This configuration only applies to backend servers that report `hot_standby`
+or `default_transaction_read_only` in startup parameters (i.e. Postgres 14+).
+
+Only allow connections with certain properties, similar to libpq's `target_session_attrs`
+connection parameter. For example: given a comma-separated list of hosts where only is
+writable, `target_session_attrs=primary` will disconnect from any hosts that are not writable.
```
JelteF, line 1295: The docs should note that `prefer-standby` is not supported. To be clear, it seems fine that it's not supported, I don't think we can easily implement this with the current way PgBouncer connects to servers.

src/server.c
```diff
@@ -241,6 +246,33 @@ static bool handle_server_startup(PgSocket *server, PktHdr *pkt)
...
+	enum TargetSessionAttrs target = server->pool->db->target_session_attrs;
+
+	// If the server did not return the in_hot_standby and/or
+	// default_transaction_read_only parameters, assume false.
+	bool hot_standby = in_hot_standby && strcmp(in_hot_standby->str, "on") == 0;
+	bool transaction_read_only = default_transaction_read_only && strcmp(default_transaction_read_only->str, "on") == 0;
+
+	return (target == TARGET_SESSION_ANY) ||
+	       (target == TARGET_SESSION_READWRITE && !transaction_read_only) ||
+	       (target == TARGET_SESSION_READONLY && transaction_read_only) ||
```
JelteF, line 271: These are wrong. The correct logic is as follows:
```suggestion
	       (target == TARGET_SESSION_READWRITE && !transaction_read_only && !hot_standby) ||
	       (target == TARGET_SESSION_READONLY && (transaction_read_only || hot_standby)) ||
```

I'd prefer to create separate variable for that though for readability. Something like the following and then use that:
```
bool read_only = default_transaction_read_only || hot_standby;
```

src/varcache.c
JelteF: That's a good question. I think it would probably make sense. This is also very related to #1205, because that also needs to track these read-only settings to be able to send the new versions to the client if they change on the server side. 

Regarding #1205, currently that PR is sending all the ParameterStatus messages again to every client if a new server reported a different one. I think a more correct, aka more in line with postgres behaviour, would be to track these read-only fields per server, and if a client is linked to a server where the values were different than for the previous link, then we'd send ParameterStatus messages for just those ones. We'd have to be a bit careful with the perf implications of that though, because it adds overhead on every connection.

And I think `default_transaction_read_only` is probably a bit special too in this regard, because it can actually be set by the client. So tracking it and then setting it according to the last client setting could actually make sense. But I'm curious how this interacts with servers configuring this setting and then changing it. If we'd set it based on the client its last value, then only new clients would get the new value. Which probably is fine, because the same would actually apply to normal connection to postgres.

@emelsimsek might have some ideas around this too.

src/admin.c
```diff
@@ -518,7 +522,11 @@ static bool admin_show_databases(PgSocket *admin, const char *arg)
 		if (db->host && strchr(db->host, ','))
 			load_balance_hosts_str = cf_get_lookup(&load_balance_hosts_lookup);
 
-		pktbuf_write_DataRow(buf, "ssissiiiissiiiiii",
+		target_session_attrs_lookup.value_p = &db->target_session_attrs;
+		if (db->target_session_attrs != TARGET_SESSION_ANY)
```
JelteF, line 526: Why keep the value NULL for `any`? (same for the other admin command below)

test/conftest.py
```diff
@@ -171,6 +177,22 @@ def pg(tmp_path_factory, cert_dir):
     pg.cleanup()
 
 
+@pytest.fixture(scope="session")
+def replica(pg, tmp_path_factory):
+    """Starts a new Postgres replica db that is shared for tests in this process"""
+    if MACOS:
+        sudo("ifconfig lo0 alias 127.0.0.2 netmask 0xff000000")
```
JelteF, line 184: We should make sure that this test is skipped for MacOS if `USE_SUDO` is not `True`. Otherwise the tests prompt for a sudo password by default.

test/utils.py
```diff
@@ -725,7 +736,13 @@ def apgctl(self, command, **kwargs):
 
     def start(self):
         try:
-            self.pgctl(f'-o "-p {self.port}" -l {self.log_path} start')
+            print("\n\nPG_LOG\n")
```
JelteF, line 739: This added `print` seems accidental

test/test.ini
```diff
@@ -54,6 +54,11 @@ replication = port=6666 host=127.0.0.1 dbname=replication
 
 client_limit_db= port=6666 host=127.0.0.1 dbname=p0 max_db_client_connections=2
 client_limit_db_auth_passthrough = port=6666 host=127.0.0.1 dbname=p0 auth_user=pswcheck max_db_client_connections=2
+primary_first = port=6666 host=127.0.0.1,127.0.0.2 dbname=p0 user=bouncer target_session_attrs=primary
+primary_second = port=6666 host=127.0.0.2,127.0.0.1 dbname=p0 user=bouncer target_session_attrs=primary
+standby_first = port=6666 host=127.0.0.2,127.0.0.1 dbname=p0 user=bouncer target_session_attrs=standby
+standby_second = port=6666 host=127.0.0.1,127.0.0.2 dbname=p0 user=bouncer target_session_attrs=standby
```
JelteF, line 60: Needs some tests for read-only, read-write, and any.

test/conftest.py
```diff
@@ -17,6 +18,11 @@
     sudo,
 )
 
+MACOS = False
+
+if platform.system() == "Darwin":
+    MACOS = True
```
JelteF, line 24: This should be importable from `.utils`

JelteF (changes requested):

test/utils.py
cosgroveb: Yep good catch ccdb6d8c85c09cd7e8a530558ac924b8a0282400

test/conftest.py
cosgroveb: Good cleanup, thanks. fc281b2fb8cad71253a3889420e9dba82739c954

test/conftest.py
cosgroveb: 5b7a91e5bdbc2d87f43e94d2ae072a80d34f9e03

test/test.ini
cosgroveb: Ah yup and those should catch this https://github.com/pgbouncer/pgbouncer/pull/1179/files#r1881005914

cosgroveb: cabc3ee6c6e60fefb99db029bca0b1ca495656a6

src/server.c
cosgroveb: This is indeed the correct logic but am having difficulty exposing this in a test. I think because: in PG14+ PostgreSQL servers begin to put `transaction_read_only` into the session and it is always set to `on` when in hot standby. Is there a scenario I can add coverage for where `transaction_read_only = off` and `in_hot_standby = true`?

Thanks for reasoning through the logic miss and the suggested clean-up 👍

cosgroveb: 8af5a6fdc5c3d7832c8b2fd7ad16c51caf4b0870

JelteF: Hmm, that does make testing one variant impossible I guess. But you can still test that read-only accepts both a hot_standby, and a database specifially marked as read only by default using `ALTER DATABASE ... SET transaction_read_only = true` (without putting it in hot_standby mode).

src/admin.c
cosgroveb: This seems to be incorrect. Removed the conditional.

src/server.c
cosgroveb: psycopg makes this annoying. I might need to use psql to `ALTER DATABASE` as `Postgres#configure` is annoyingly giving `psycopg.errors.CantChangeRuntimeParam: parameter "transaction_read_only" cannot be changed`

cosgroveb: Ah silly me `configure` is `ALTER SYSTEM`. Not psycopg's fault at all.

cosgroveb: I believe this feedback is addressed in f752642b97781680f07fd8232deda7dcef16cae0 and 4ef999a6a2d7a0cc99d5449698b875c1284a2203 please unresolve if you disagree.

src/varcache.c
cosgroveb: After thinking about this more and reading the docs again I'm not sure that it should be a design goal here should be to enforce whether or not a parameter tracked by the varcache is read-only or read-write. If we set the varcache only after getting the pertinent `ParameterStatus` message from the server we can be certain that it is the valid value for the back-end connection.

From the docs (emphasis mine):

```ParameterStatus messages will be generated whenever the active value changes for any of the parameters the backend believes the frontend should know about. Most commonly this occurs in response to a SET SQL command executed by the frontend, and this case is effectively synchronous — but it is also possible for parameter status changes to occur because the administrator changed a configuration file and then sent the SIGHUP signal to the server. Also, *if a SET command is rolled back, an appropriate ParameterStatus message will be generated to report the current effective value.*```

I think that it's none of our business if the client issues a `SET` command that will fail and we can continue to transparently return whatever error the server gave the client.

Comment (cosgroveb) [reply src/varcache.c:91]: The docs go on to talk about the hard-wired set of `ParameterStatus` messages and warn:

> This set might change in the future, or even become configurable.

So, IMHO, let's leave it.

Comment (cosgroveb) [reply src/varcache.c:91]: I think that, then, separates this concern from #1205 though I think that propagating `ParameterStatus` out to all connected clients _may_ be a good change. I am not 100% sure, though, because if an administrator issues a `reload` would it not already inform all connected clients?

Comment (cosgroveb) [reply src/varcache.c:91]: https://www.postgresql.org/docs/current/protocol-flow.html

Comment (cosgroveb) [reply src/varcache.c:91]: @ya-mouse OK thanks, your comment made me think a bit further.

At a minimum I need to add tests wherein the server role changes and observe what happens to the server connections that no longer satisfy `target_session_attrs` and what the client sees. Will look tonight.

Comment (cosgroveb) [reply src/varcache.c:91]: Strike everything I said above about read-only parameters and target_session_attrs 🙂 

The feature is broken with the default load_balance_hosts=round-robin and the default target_session_attrs=any (if you have multiple hosts in the hostlist) because of the PR's varcache usage.

Comment (cosgroveb) [reply src/varcache.c:91]: This test demonstrates the issue at hand. I need to think more about this but wanted to share something concrete:

https://github.com/pgbouncer/pgbouncer/pull/1179/commits/2879cfe42344a5310c4a45f993ecd2f927ba82c6

Comment (cosgroveb) [reply src/varcache.c:91]: The most straightforward way to not fail to connect to read-onlys in round-robin is to simply not attempt to `SET` `in_hot_standby`. 1dcdb42d197ab389ec4825c36a160ae4638a2705

Comment (cosgroveb) [reply src/varcache.c:98]: I took a pass at making the concept of read-only vars more real. [Please have a look](https://github.com/pgbouncer/pgbouncer/pull/1179/commits/39a11f3144d1e7c645bd02f1d43032f0835507e0) @JelteF and @ya-mouse.

File: test/test_ssl.py [hunk 11]
```diff
@@ -289,29 +289,6 @@ def test_client_ssl_reload_change_ca(bouncer, cert_dir):
     bouncer.psql_test(host="localhost", sslmode="verify-full", sslrootcert=new_root)
 
 
-def test_client_ssl_auth(bouncer, cert_dir):
```
cosgroveb, line 292: Obviously this is not the solution to the problem I've found with test isolation here. Just wanted to be sure I could get the PR clean on CI. Details in #1280

cosgroveb: Renaming our test file so that it sorts alphabetically before "test_ssl.py" "fixes" it https://github.com/cosgroveb/pgbouncer/commit/c9bf1b1c02ea867adf81a2195895b3015b0bc4d0

🤮

cosgroveb:
Day Job has been pretty busy so I've let slip a needed rebase here to make this more reviewable. This is helpful for Day Job, though, so I'll try and work in a rebase in the next few weeks.

saolof:
If this is close to done and this just needs a rebase and some polish, would you be fine with external help for the last push? This feature would be pretty nice for replacing haproxy with just pgbouncer plus a few A records, I'm willing to help in any way I can

cosgroveb:
@saolof contributions welcome. I would be super happy to get this over the line. I've been running a forked pgbouncer at scale for a few years now but I keep getting pulled in other directions at work. Merging this gets me back onto mainline pgbouncer.

saolof:
Okay cool. I'll squash and reorganize the commits in a somewhat ahistorical order and replace the readonly struct member with a function to make the rebase work easier. (edit: keeping the readonly struct member, that part was already easy to rebase)

src/varcache.c
saolof: Oh right, the varcache thing was mentioned here and its strictly internal to src/varcache.c, and varcache seems straightforward to rebase actually

AndrewJackson2020:
I'm happy to help here as well, would love to see this get over the line. 

Built and tested. Was curious how it handles multiple hosts, seems to do it per libpq implementation where you cannot mix standbys/primaries on a single host. If you try to mix and it encounters a mismatch target_session_attrs it will not go onto the next one. This is annoying for anyone wanting to just define their entire cluster on a single DNS address with overloaded A records and let target_session_attr fine the right one but since it conforms to libpq probably it needs to be kept that way in this implementation.

doc/config.md
```diff
@@ -1313,6 +1313,34 @@ fast retries when multiple hosts are available.
 
 Default: `round-robin`
 
+### target_session_attrs
+
+Note: This configuration only applies to backend servers that report `hot_standby`
+or `default_transaction_read_only` in startup parameters (i.e. Postgres 14+).
+
+Only allow connections with certain properties, similar to libpq's `target_session_attrs`
+connection parameter. For example: given a comma-separated list of hosts where only is
```
AndrewJackson2020, line 1408: > For example: given a comma-separated list of hosts where only is
writable, 

I think this sentence is missing a word. Maybe "For example: given a comma-separated list of hosts where only **one** is writable"?

test/utils.py
```diff
@@ -805,7 +838,7 @@ def nossl_access(self, dbname, auth_type, user="all"):
         with self.hba_path.open(mode="w") as pghba:
             if USE_UNIX_SOCKETS:
                 pghba.write(f"local {dbname}   {user}                {auth_type}\n")
-            pghba.write(f"hostnossl  {dbname}   {user}  127.0.0.1/32  {auth_type}\n")
```
AndrewJackson2020, line 884: Why is this being changed?

AndrewJackson2020: Okay I get it, its because the read replica is being put on the same port, different internal IP so we need to expand the range. Also pgbouncer currently does not support a list of ports so every hostname needs to be on the same port. Makes sense.

AndrewJackson2020: Raised a PR #1496 which adds the ability to specify multiple ports. If #1496 get merged before this PR it will allow this PR to test on MacOS without sudo as we can use different ports instead of different IPs

src/varcache.c
cosgroveb: > Brian and I were discussing this offline and wondered if we need to make 2 distinct lists.

I've been giving this more thought. The separation idea here is right. Maybe not two lists.

This PR mixes backend selection state with replayable client state in VarCache; the [attempted](https://github.com/pgbouncer/pgbouncer/commit/1dcdb42d197ab389ec4825c36a160ae4638a2705) `SET in_hot_standby` exposed the problem.

default_transaction_read_only has two roles: backend selection and optional client replay through track_extra_parameters. This PR conflates them.

This doesn't need to solve #1205. With per-server state outside VarCache, #1205 can compare the backend with what the client last saw.

I'm reworking this separately.

saolof:
Ah, I see that this got rebased. Very nice work!

cosgroveb:
> Ah, I see that this got rebased. Very nice work!

In fact I have an improved version of this feature in the works and expect to close this PR in favor of it rather soon.

cosgroveb:
@saolof my attempt in this PR to rebase and upstream my team's original work here was absolute dogshit.

I opened a [much improved, more complete, and more reviewable](https://github.com/pgbouncer/pgbouncer/pull/1534) version that is hopefully acceptable to @JelteF and @eulerto... @JelteF has given a lot of important feedback on this one that took me an embarrassingly long time to process 🤞 😅 ❤️ . 

I think also that #1534 lines up well with the aim of #1205.

PR #1180, opened 2024-10-02
Expand warnings around the implications of transaction pooling 

cosgroveb:
The motivation for this PR is to better-educate users who, _reasonably_, may want to use transaction pooling but may also be running one of the world's [most popular web frameworks](https://github.com/rails/rails/blob/fa152fba87f3a11790783207e6e84e58d345c012/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L949-L990) which would be, cryptically, considered a "broken setup" by the current docs since it manipulates the session extensively out of the box with no levers that I am aware of to prevent that! 😄

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -610,7 +610,8 @@ session features over a transaction-pooled PgBouncer.  It changes non-determinis
 to deterministic breakage: Clients always lose their state after each
 transaction. Client applications should be designed to not use session features
 over a transaction-pooled PgBouncer but this can provide a guardrail against
-unintentional breakage in an application or its' dependencies (in, say, a hypothetical ORM).
+unintentional breakage in an application or its' dependencies (for instance ActivRecord
+uses session features out of the box).
 
 Default: 0
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -605,12 +605,23 @@ setting is off (default), the `server_reset_query` will be run only in pools
 that are in sessions-pooling mode.  Connections in transaction-pooling mode
 should not have any need for a reset query.
 
-This setting is for working around broken setups that run applications that use session features
-over a transaction-pooled PgBouncer.  It changes non-deterministic breakage
+This setting is for working around ill-advised setups which run applications that use
+session features over a transaction-pooled PgBouncer.  It changes non-deterministic breakage
 to deterministic breakage: Clients always lose their state after each
-transaction.
+transaction. Client applications should be designed to not use session features
+over a transaction-pooled PgBouncer but this can provide a guardrail against
+unintentional breakage in an application or its' dependencies (in, say, a hypothetical ORM).
 
 Default: 0
+
+Note: Using `DISCARD ALL` also works around the very non-obvious potential performance land-mine
+that can happen to query plan caching with the server-side `plan_cache_mode` default of `auto`.
+Clients can be stuck with a cached plan that was efficient for a previous client that is
+no longer because the dataset being queried has changed. With transaction pooling it may
+be worth considering setting `plan_cache_mode` to `custom`. When using transaction pooling it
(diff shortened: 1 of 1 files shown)


cosgroveb:
The motivation for this PR is to better-educate users who, _reasonably_, may want to use transaction pooling but may also be running one of the world's [most popular web frameworks](https://github.com/rails/rails/blob/fa152fba87f3a11790783207e6e84e58d345c012/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L949-L990) which would be, cryptically, considered a "broken setup" by the current docs since it manipulates the session extensively out of the box with no levers that I am aware of to prevent that! 😄

doc/config.md
```diff
@@ -605,13 +605,25 @@ setting is off (default), the `server_reset_query` will be run only in pools
...
 Default: 0
 
+Note: Using `DISCARD ALL` also works around the very non-obvious potential performance land-mine
+that can happen to query plan caching with the server-side `plan_cache_mode` default of `auto`.
+Clients can be stuck with a cached plan that was efficient for a previous client that is
+no longer because the dataset being queried has changed. With transaction pooling it may
+be worth considering setting `plan_cache_mode` to `custom`. When using transaction pooling it
+is at a minimum [strongly adised to review the Postgres docs][postgres-plan-cache-mode-docs].
+
+[postgres-plan-cache-mode-docs]: https://www.postgresql.org/docs/current/runtime-config-query.html#GUC-PLAN-CACHE-MODE
```
JelteF, line 625: I think this note should maybe be part of the `max_prepared_statements` config, since that is enabling prepared statements in the first place. At minimum we should link there from here.

doc/config.md
```diff
@@ -605,13 +605,25 @@ setting is off (default), the `server_reset_query` will be run only in pools
...
-This setting is for working around broken setups that run applications that use session features
-over a transaction-pooled PgBouncer.  It changes non-deterministic breakage
+This setting is for working around ill-advised setups which run applications that use
+session features over a transaction-pooled PgBouncer.  It changes non-deterministic breakage
 to deterministic breakage: Clients always lose their state after each
-transaction.
+transaction. Client applications should be designed to not use session features
+over a transaction-pooled PgBouncer but this can provide a guardrail against
+unintentional breakage in an application or its' dependencies (for instance ActivRecord
+uses session features out of the box).
```
JelteF, line 614: I think it would be good to clarify exactly what the session level features are unsupported? Because a bunch of session features actually are supported these days, see `max_prepared_statements` and `track_extra_parameters`. So I think it would be good to call out what is and what isn't supported. And it might be useful to say what exactly ActiveRecord is doing that is unsupported.

doc/config.md
cosgroveb: Ah maybe I didn't outline this gotcha very well because you don't need to use prepared statements to get a backend conn with a suboptimal cached plan

cosgroveb: Rather implicit prepared statements

cosgroveb: Ah but yes that warning may be better fit if moved per your suggestion

doc/config.md
eulerto: Do we really need to mention tools here? These tools might eventually change and this information is obsolete. I prefer the other way, open a PR for the referred product saying that it doesn't work with PgBouncer if you are using feature X. Once they fix this issue, their documentation is updated accordingly saying that PgBouncer is supported.

cosgroveb:
I want to think more about helpful ways to communicate the things about transaction pooling that have been surprising to me and also plan caching so that I can make very clear and easy to understand updates to the documentation here. Thank you for reviewing these thoughts @eulerto and @JelteF

Issue #954, opened 2023-10-03
Getting 'E' packet on "auth_conn query" instantly


ylacancellera:
Hi

I currently have a repeatable issue where users can connect to every db, except one (with appropriate pg privileges).
Users are able to connect to other databases and switch between them with \c, but not to the "broken" one. 
 
I notice I have a "error response from auth_query", I tried debugging it:
```
2023-10-03 06:47:02.635 UTC [7] NOISE C-0x561ead097b10: some_user/(nouser)@<client_ip>:<client_port> doing auth_conn query
2023-10-03 06:47:02.635 UTC [7] DEBUG pktbuf_dynamic(512): 0x561ead0a24a0
2023-10-03 06:47:02.635 UTC [7] NOISE tls_write: req=108 out=108
2023-10-03 06:47:02.635 UTC [7] DEBUG pktbuf_free(0x561ead0a24a0)
2023-10-03 06:47:02.635 UTC [7] NOISE resync(10): done=0, parse=0, recv=0
2023-10-03 06:47:02.635 UTC [7] NOISE tls_read: req=4096 out=103
2023-10-03 06:47:02.635 UTC [7] NOISE S-0x561ead0a83f0: some_user/_pgbouncer@<pg_ip>:5432 read pkt='E', len=103
2023-10-03 06:47:02.635 UTC [7] LOG S-0x561ead0a83f0: some_user/_pgbouncer@<pg_ip>:5432 closing because: error response from auth_query (age=0s)
2023-10-03 06:47:02.635 UTC [7] LOG C-0x561ead097b10: some_user/(nouser)@<client_ip>:<client_port> closing because: bouncer config error (age=0s)
2023-10-03 06:47:02.635 UTC [7] WARNING C-0x561ead097b10: some_user/(nouser)@<client_ip>:<client_port> pooler error: bouncer config error

```

I have some questions regarding it:
- is it possible something else than postgres is answering this "E" packet ? (a pgbouncer cache, anything)
- is it weird taht everything is instantaneous here ? It is not on successful logins
- could that be a pg_catalog issue somehow ?
- anything particular logging you would advise on postgres side to dive further ?

ylacancellera:
Found it.
It was a permission issue on pgbouncer schema. The error is obfuscated unfortunately, it would be really nice to display it as pgbounceruser was getting a proper "permission denied" when trying to use select * from pgbouncer.get_auth(x) directly

JelteF:
[...]

PR #1187, opened 2024-10-20
Log server error when issuing auth query

greg-rychlewski:
Closes https://github.com/pgbouncer/pgbouncer/issues/954

This is my attempt to log the error message from auth queries coming from Postgres.

I tried to follow what happens here: https://github.com/pgbouncer/pgbouncer/blob/master/src/server.c#L151. When receiving an error packet they call `log_server_error` before `disconnect_server` and the former's note has `S:` prepended to it. I am guessing that is to distinguish it as a note coming from the server rather than something generated from pgbouncer.


JelteF (approved):
Thanks for this change, it looks good but needs a small rebase.

JelteF:
If you have time to fix the Windows build in another PR that would be greatly appreciated.

PR #1188, opened 2024-10-22
Add user level idle transaction timeout

AndrewJackson2020:
This PR implements the user level equivalents of the server level timeouts `idle_transaction_timeout` and `query_timeout`. I implemented both of these timeouts together because the server level implementations of these timeout are implemented together as well.


doc/config.md
```diff
@@ -1295,6 +1295,12 @@ database or default `pool_mode` is used.
 Configure a maximum for the user (i.e. all pools with the user will
 not have more than this many server connections).
 
+### idle_transaction_timeout
```
JelteF, line 1298: Missing docs on `query_timeout`

src/janitor.c
```diff
@@ -595,6 +596,38 @@ static void pool_server_maint(PgPool *pool)
 		}
 	}
 
+	/* handle user level idle_transaction_timeout */
+	if (any_user_level_timeout_set) {
+		statlist_for_each_safe(item, &pool->active_server_list, tmp) {
```
JelteF, line 601: I think this loop should be combined with the loop above. The user level timeout should override the global one, so if the user has a higher timeout than the global one, the connection should not disconnect when the global one expires, only when the user one expires. Right now this isn't the case afaict.

JelteF (commented):
Looks like a useful feature. Some notes on the implementation & docs, but overall looks reasonable.

On another note: It seems like you're interested in contributing regularly to PgBouncer. We're always on the lookout for people who want to help maintain the project. Since today, there's now a discord channel on the Postgres Hackers discord, specific for PgBouncer development. That hopefully makes discussions/questions/mentoring a bit easier to manage, please join if you're interested: https://discordapp.com/channels/1258108670710124574/1300532992304742481

src/janitor.c
AndrewJackson2020: Agreed. Will modify accordingly.

doc/config.md
AndrewJackson2020: Will fix.

AndrewJackson2020: Added query_timeout documentation.

src/janitor.c
AndrewJackson2020: Combined the loops as suggested with the override logic. Though it would probably make sense to unit test that user level timeouts correctly overwrite the global level timeouts.

doc/config.md
```diff
@@ -1304,6 +1304,17 @@ database or default `pool_mode` is used.
 Configure a maximum for the user of server connections (i.e. all pools with the user will
 not have more than this many server connections).
 
+### query_timeout
+
+Set the maximum length of time that a user query can run for.
```
JelteF, line 1309:
```suggestion
Set the maximum number of seconds that a user query can run for.
```

doc/config.md
```diff
@@ -1304,6 +1304,17 @@ database or default `pool_mode` is used.
...
+If set this timeout overrides the server level query_timeout described above.
+
+### idle_transaction_timeout
+
+Set the maximum length of time a user can have an idle transaction open.
```
JelteF, line 1314:
```suggestion
Set the maximum number of seconds that a user can have an idle transaction open.
```

src/janitor.c
```diff
@@ -27,6 +27,7 @@
 /* do full maintenance 3x per second */
 static struct timeval full_maint_period = {0, USEC / 3};
 static struct event full_maint_ev;
+extern bool any_user_level_timeout_set;
```
JelteF, line 30: Let's rename this variable, because its name is too generic now that #1189 is merged.
```suggestion
extern bool any_user_level_server_timeout_set;
```

AndrewJackson2020:
@JelteF Rebased. All checks pass. Let me know if there is anything else that can be changed.

src/janitor.c
```diff
@@ -550,6 +551,11 @@ static void pool_server_maint(PgPool *pool)
 	usec_t now = get_cached_time();
 	PgSocket *server;
 
+	usec_t effective_query_timeout;
+	usec_t effective_idle_transaction_timeout;
+	usec_t user_query_timeout;
+	usec_t user_idle_transaction_timeout;
+
```
JelteF, line 558: nit: let's move this inside the loop. No need to declare these variables with this large of a scope.

JelteF (commented):
Sorry for the delayed re-review. Overall, this looks great. Left a final nitpick comment.

src/janitor.c
AndrewJackson2020: Moved these variable declarations into the loop with the latest commit.

PR #1189, opened 2024-10-22
implement user client idle timeout

AndrewJackson2020:
This PR implement a user level version of the client_idle_timeout.


doc/config.md
```diff
@@ -1295,6 +1295,14 @@ database or default `pool_mode` is used.
 Configure a maximum for the user (i.e. all pools with the user will
 not have more than this many server connections).
 
+### client_idle_timeout
+
+Set the maximum amount of time that a client is allowed to idly connect to
```
JelteF, line 1300:
```suggestion
Set the maximum amount of time in seconds that a client is allowed to idly connect to
```

doc/config.md
```diff
@@ -1295,6 +1295,14 @@ database or default `pool_mode` is used.
...
+the pgbouncer instance. If set this timeout applies in addition to the
+pgbouncer level config described above.
```
JelteF, line 1302: The global setting notes this:

> This should be larger than the client-side connection lifetime settings, and only used for network problems.

I guess that's not true for your usecase, otherwise you probably wouldn't need this at the user level. But that does make me curious what your usecase is.

JelteF (commented):
Overall looks like a fine change. But I'm curious why you need this feature. Also needs a merge conflict resolution.

doc/config.md
AndrewJackson2020: The usecase is that I manage a multi tenant database with users of various levels of trustworthiness. Users who connect via a client connection Pooled application are more trustworthy. Users who connect via psql or pgadmin are less trustworthy. Would like the ability to give untrustworthy users more strict settings around stuff like idle sessions or idle transactions. 

Honestly idle sessions are not the biggest deal, especially compared to idle transaction, but I figured I would put together a PR as it is a fairly easy implementation.

AndrewJackson2020:
One thing about this PR is that it applies in addition to the global level setting. Most of the more granular settings appear to replace the global level settings. I realized this after raising the PR. If you want this can be reworked to act in the same manner.

src/janitor.c
```diff
@@ -392,6 +394,19 @@ static void pool_client_maint(PgPool *pool)
 		}
 	}
 
+	/* check user client_idle_timout */
+	if (any_user_level_client_timeout_set) {
+		statlist_for_each_safe(item, &pool->active_client_list, tmp) {
```
JelteF, line 399: > Most of the more granular settings appear to replace the global level settings. I realized this after raising the PR. If you want this can be reworked to act in the same manner.

Yeah, let's merge this with the loop above so that the user level setting replaces the global one.

AndrewJackson2020: Agreed. Will do.

AndrewJackson2020: Merged the loops as described and added additional tests to cover that the user level setting overrides the global level setting.

JelteF (approved):

PR #1191, opened 2024-10-25
Improve SHOW CLIENTS by adding "idle" state

### LINKED ISSUE #1071: Pgbouncer PAUSE command  getting executed without completing the transaction in transaction mode.
Screenshot from 2024-05-15 16-34-46

There are still connection existing which are active after giving pause.I'm using pgbouncer version 1.19

rajaryanece:
Closes #1071
As @JelteF suggested in #1071 -

> I do think that reporting could be improved here though, because this report definitely confused me too. Maybe we can introduce an idle state.


JelteF:
Code looks good. But docs need to be updated to explain this new value (and update the docs for `active`)

rajaryanece:
Sure , will update the doc and add the changes here.

> (and update the docs for active)

what does this mean ? in addition to updating the docs for admin commands with this state, where else would we need to update this reporting?

eulerto:
@rajaryanece I think @JelteF is referring to `cl_active` (SHOW POOLS). I didn't review this PR yet but in a quick look, isn't it better to change the state instead of creating a fake state based on 2 values? If you want to have a real idle state, you need to set the client state as idle.

JelteF:
I kinda agree that having a dedicated state is more in line with how all the other states work in PgBouncer. But we set `client->link` to `NULL` in many places, so we'd have to change all those places to also set the state to `CF_IDLE`. And I'd be pretty worried that something would break because it did not check for CF_IDLE. 

I think this 1 line change is far less risky, because it only involves the reporting code for the admin commands. And it does solve a real confusing thing for users. So I would prefer this change over a change that introduces a new state.

JelteF:
This still needs a documentation update

rajaryanece:
working on the test and documentation will add the commit

rajaryanece:
tests failing due to unrelated - [issue](https://github.com/pgbouncer/pgbouncer/pull/1231#issuecomment-2594457051)

AndrewJackson2020:
> tests failing due to unrelated - [issue](https://github.com/pgbouncer/pgbouncer/pull/1231#issuecomment-2594457051)

Opened #1249 with a potential fix to this.

rajaryanece:
pulled fix of the [issue](https://github.com/pgbouncer/pgbouncer/pull/1231#issuecomment-2594457051) from master and the update pr - the tests are passing now.

rajaryanece:
Hi @JelteF, As I added the test and documentation, I think this PR is ready to be reviewed,  please help review it again. Thanks a lot.

JelteF (approved):
This is good now. I'll wait with merging until after the next patch release is done though.

PR #1192, opened 2024-10-28
feat: expose usage counters for prepared statements

sboschman:
With prepared statements tracking enabled in PgBouncer we internally track the following counters:
- ps_client_parse_count
- ps_server_parse_count
- ps_bind_count

But we only expose them in log/systemd messages, but not in any `SHOW` command, which are used by external monitoring tools like pgbouncer_exporter. These counters are imo important to gain insight in how well 'prepared statement caching' in PgBouncer is working for your workload and your configuration. I think of them as a sort of 'cache hit ratio'.


sboschman:
@JelteF hopefully you have time to take a look at this addition to the prepared statement implementation

JelteF:
LGTM, but needs some manual testing on my or someone else's side.

JelteF:
If you could add some tests for this that would be great btw.

sboschman:
> If you could add some tests for this that would be great btw.

What kind of tests are you thinking about? The existing `test_admin` tests already check the `SHOW` commands. So f.e. a mismatch between header columns and data columns is detected.

JelteF:
Basically a modification to `test_show_stats()` to show that the returned counts make sense: https://github.com/pgbouncer/pgbouncer/blob/585a6300995e7e3e714e9b55d29caf16bb350ddd/test/test_admin.py#L226

sboschman:
A small test case has been added to the prepared statement test suite: [test_prepared.py](https://github.com/pgbouncer/pgbouncer/pull/1192/files#diff-e424c5b27444d6f06805fb0d28508c50a5706520a99653a2996a0d1b1200de49)

Anything else I can do to help out on this issue @JelteF ?

JelteF (approved):

PR #1193, opened 2024-10-28
Use PqMsg_ symbols from PostgreSQL

petere:
Use the descriptively-named macros for identifiers used in wire protocol messages, introduced in PostgreSQL 17.  The header file `protocol.h` is copied from PostgreSQL.

---

Separate AUTH_* symbols for protocol and configuration purposes

PgBouncer has used the same set of AUTH_* symbols for handling the authentication protocol (AuthenticationRequest message) and for storing configuration settings (auth_type, hba).  The two sets have diverged significantly over time, and it's gotten quite confusing and cumbersome.

This creates two separate sets of symbols.  For the protocol, we use the AUTH_REQ_* symbols imported from PostgreSQL in protocol.h.  For the configuration settings, we make a separate enum with AUTH_TYPE_* symbols.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -129,11 +129,13 @@ static bool handle_server_startup(PgSocket *server, PktHdr *pkt)
 	/* ignore most that happens during connect_query */
 	if (server->exec_on_connect) {
 		switch (pkt->type) {
-		case 'Z':
-		case 'S':	/* handle them below */
+		case PqMsg_ReadyForQuery:
+		case PqMsg_ParameterStatus:
+			/* handle them below */
 			break;
 
-		case 'E':	/* log & ignore errors */
+		case PqMsg_ErrorResponse:
+			/* log & ignore errors */
 			log_server_error("S: error while executing exec_on_query", pkt);
 		/* fallthrough */
 		default:	/* ignore rest */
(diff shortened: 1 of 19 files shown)


src/server.c
```diff
@@ -469,55 +473,55 @@ static bool handle_server_work(PgSocket *server, PktHdr *pkt)
...
-	case 'V':		/* FunctionCallResponse */
+	case PqMsg_ParameterDescription:
+	case PqMsg_CopyDone:
+	case PqMsg_CopyFail:
+	case PqMsg_FunctionCallResponse:
 		break;
-	case 'I':		/* EmptyQueryResponse == CommandComplete */
-	case 's':		/* PortalSuspended */
-		pop_outstanding_request(server, "E", &ignore_packet);
+	case PqMsg_EmptyQueryResponse:	/* EmptyQueryResponse == CommandComplete XXX */
```
JelteF, line 517: Agreed that this comment is confusing

```suggestion
	case PqMsg_EmptyQueryResponse:	/* EmptyQueryResponse is similar to CommandComplete, which is handled above */
```

JelteF (approved):
Thanks for this one. When I saw those new constants in Postgres I was very jealous of them but I hadn't found the time to update PgBouncer to use the same.

JelteF:
Let's try to get this merged quickly btw, it seems like a one with a high possibility for merge conflicts (there's already one now).

src/server.c
petere: done

petere:
fixed the comment and rebased over the conflict

JelteF:
Merged

Issue #1174, opened 2024-09-25
could not read pidfile while running in daemon mode.


Monk-Liu:
I have set pidfile in config.ini in order to run in daemon mode.
```
pidfile = /var/run/pgbouncer/pgbouncer.pid
```
However, when I start pgbouncer, it raise an error.
```
# pgbouncer -v /etc/pgbouncer/pgbouncer.ini
2024-09-25 18:18:12.897 CST [14956] FATAL could not read pidfile '/var/run/pgbouncer/pgbouncer.pid': No such file or directory
```
 the error is raised from "src/main.c:722".

It seems that we can't just create an empty pgbouncer.pid file, instead, we should add one number to the pidfile.  But why? Shouldn't the code handle the situation that the pidfile is empty when we first start pgbouncer?

petere:
The error handling in the code is sloppy. It should not treat a return of 0 from `read()` as a system call error, but handle it separately.

One of the purposes of a PID file is to protect against multiple instances interfering with each other. So being restrictive about how faulty PID files are treated seems sound to me.

Monk-Liu:
get it, thanks Petere!

JelteF:
I'll re-open this, as we should handle the error better. But manually creating pidfiles yourself is definetly not something that's recommended.

PR #1195, opened 2024-10-29
Separate error handling for empty pidfile

greg-rychlewski:
Closes https://github.com/pgbouncer/pgbouncer/issues/1174

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -721,8 +721,10 @@ static void check_pidfile(void)
 	}
 	res = read(fd, buf, sizeof(buf) - 1);
 	close(fd);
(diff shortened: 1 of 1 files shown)


eulerto (changes requested):
This PR seems reasonable to me. However, I have some suggestion. Instead of having a different message for this case (empty file), use `goto locked_pidfile`. It reports exact situation. I also suggest that you change the goto message to include the pidfile name.

greg-rychlewski:
Thank you :). Updated.

eulerto (approved):

PR #1204, opened 2024-11-05
implement hostname in pg hba conf

AndrewJackson2020:
This PR implements hostnames in the pg_hba.conf file. In many environments ip addresses can be very volatile and often ephemeral and it is much more reliable to be able to use dns addresses instead of IP addresses directly.

2 issues with the current state of this PR:
1. There is no unit testing. Not sure exactly how to unit test this without knowledge of CIs DNS. Possibly I can just put an entry into /etc/hosts?
2. The current implementation of this is network blocking/synchronous. Tried to get this to work using the DNS resolver on src/dns.c but was not able to get this to work exactly.


AndrewJackson2020:
Added a test that helps cover the new functionality somewhat, though it doesn't cover the cases of a hostname that maps back to ipv4 and a hostname that maps back to ipv6.

src/hba.c
```diff
@@ -940,17 +966,50 @@ static bool match_inet6(const struct HBAAddress *haddress, PgAddr *addr)
...
+		case AF_INET:
+			return match_inet4(haddress, addr);
+		case AF_INET6:
+			return match_inet6(haddress, addr);
+		default:
+			return false;
+		}
+	} else {
+		/* match DNS name */
+		ret = getaddrinfo(haddress->hostname, NULL, NULL, &gai_result);
```
JelteF, line 990: This seems to be doing only half of what postgres does to validate the hostname:

> If a host name is specified (anything that is not an IP address range or a special key word is treated as a host name), that name is compared with the result of a reverse name resolution of the client's IP address (e.g., reverse DNS lookup, if DNS is used). Host name comparisons are case insensitive. If there is a match, then a forward name resolution (e.g., forward DNS lookup) is performed on the host name to check whether any of the addresses it resolves to are equal to the client's IP address. If both directions match, then the entry is considered to match. 

Source: https://www.postgresql.org/docs/current/auth-pg-hba-conf.html

Your code is only doing the forward resolution of the hostname, not the reverse lookup of the IP address.

src/hba.c
```diff
@@ -940,17 +966,50 @@ static bool match_inet6(const struct HBAAddress *haddress, PgAddr *addr)
 	       (src[2] & mask[2]) == base[2] && (src[3] & mask[3]) == base[3];
 }
 
+
```
JelteF, line 969: nit: unnecessary added newline

src/hba.c
JelteF, line 990: I'm not sure what problems you ran with our async DNS. But we should definitely try to use that. We have the libraries available. Blocking the event loop is something we should only do in cases where we cannot avoid it.

JelteF (changes requested):

src/hba.c
AndrewJackson2020: I did attempt to use async DNS but was not able to figure it out. Probably skill issues on my end. Will give it another shot.

AndrewJackson2020:
Closing this PR for now as I was not able to figure out the async DNS resolution and even by the pg spec using DNS entries in HBA doesnt solve the issue I was hoping to mitigate involving the HBA file being resilient to server moves which would require resolution of CNAMEs and not just hostnames.

Issue #859, opened 2023-06-05
Cached in_hot_standby value


careless6666:
We had a situation when a switchover occurred on the pantroni cluster and when trying to connect via pgbouncer, problems started when connecting with target session attribute=read-write, after authorization, a message arrives in the socket, including where there are parameters in_hot_standby and default_transaction_read_only (https://www .postgresql.org/docs/current/protocol-flow.html#PROTOCOL-ASYNC). The parameter in_hot_standby = on, although it should have been off, the problem was solved only when the pgbouncer was reloaded. 

Additional complexity here is imposed by the fact that psql itself cannot work normally with the target session attribute
in the example, it opens a connection with the primary type to a synchronous replica and naturally falls on the insert
It's hard to debug such situations

```
psql "postgresql://ic-da-service-ml-titles-dev@xx.xx.xx.xx:6532/ic-da-service-ml-titles-0002?target_session_attrs=primary"                                                                                    

Password for user ic-da-service-ml-titles-dev:
psql (14.5 (Homebrew), server 14.6 (Ubuntu 14.6-1.pgdg20.04+1))
Type "help" for help.

ic-da-service-ml-titles-0002=> insert into version_info (version, applied_on, description) VALUES ('00', '2023-04-13 15:25:59.000000', 'test');
ERROR:  cannot execute INSERT in a read-only transaction
```

Is it possible to disable caching of these parameters?

```
pgbouncer --version
PgBouncer 1.16.1
libevent 2.1.11-stable
adns: c-ares 1.15.0
tls: OpenSSL 1.1.1f  31 Mar 2020
```

JelteF:
I'm surprised by this bugreport. Because afaik PgBouncer doesn't actively cache in_hot_standby or default_transaction_read_only (a grep through the codebase for either of those doesn't show up anything). Could you share some steps to reproduce the issue? That would help with debugging and finding out what the cause is here (e.g. maybe pgbouncer should cache these settings and report them to the client).

Melkij:
Probably somewhere around here https://github.com/pgbouncer/pgbouncer/blob/master/include/bouncer.h#L369
variable not explicitly excluded from welcome message and forwarded to clients as is. Some time ago I checked why if you update the database, then psql connected via pgbouncer will report that the version is old until you command `reconnect` from pgbouncer console (among other things, this resets the welcome_msg state)

eulerto:
PR that @JelteF is referring to is #867 .

solry:
[...]

PR #1205, opened 2024-11-05
Update welcome message and push non-cachable variables on client link (#859)

ya-mouse:
**Summary:** These changes aim to keep client connections accurately synchronized with the server state changes (i.e. standby role change).

This update addresses the need to propagate server-side changes in the cached welcome message to each client connection immediately, rather than waiting for a subsequent connection. Currently, when variables that affect the welcome message change (e.g., in_hot_standby), these updates only reach the client after a reconnection. This is particularly relevant for specific drivers, such as Npgsql, which rely on real-time updates within the same client connection to avoid inconsistencies and unexpected behavior ([related issue in Npgsql](https://github.com/npgsql/npgsql/issues/5766)).

The change updates the logic to send the latest welcome message variables when a client is linked, ensuring that any changes in non-cached server variables are consistently applied to active client connections.

Fixes: https://github.com/pgbouncer/pgbouncer/issues/859 https://github.com/npgsql/npgsql/issues/5766

### DIFF
--- a/include/varcache.h
+++ b/include/varcache.h
@@ -34,3 +34,9 @@ void varcache_clean(VarCache *cache);
 void varcache_add_params(PktBuf *pkt, VarCache *vars);
 void varcache_deinit(void);
 void varcache_set_canonical(PgSocket *server, PgSocket *client);
+
+bool welcome_vars_set(WelcomeVarLookup **vars, const char *key, const char *val, bool when_different);
+void welcome_vars_add_params(PktBuf *pkt, WelcomeVarLookup *vars);
+const WelcomeVarLookup *welcome_vars_get(WelcomeVarLookup *vars, const char *key);
+void welcome_vars_apply(PgSocket *client);
+void welcome_vars_deinit(WelcomeVarLookup **vars);
--- a/src/objects.c
+++ b/src/objects.c
@@ -948,26 +948,12 @@ bool find_server(PgSocket *client)
 
 	/* link or send to waiters list */
 	if (server) {
-		PktBuf *msg;
-		const WelcomeVarLookup *lk, *tmp;
-
 		slog_noise(client, "linking client to S-%p", server);
 		client->link = server;
 		server->link = client;
 		server->pool->stats.server_assignment_count++;
 
-		if (client->state == CL_ACTIVE && !client->wait_for_welcome) {
-			msg = pktbuf_temp();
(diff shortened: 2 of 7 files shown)


src/proto.c
```diff
@@ -22,6 +22,7 @@
 
 #include "bouncer.h"
 #include "scram.h"
+#include "common/uthash_lowercase.h"
```
JelteF, line 25: This should not be added to each file, it should be isolated to `varcache.c` (or some other dedicated file). Any `HASH_` macros should only be used in that file, otherwise the hash algorithm differs between invocations of `HASH_`

src/objects.c
```diff
@@ -948,10 +948,25 @@ bool find_server(PgSocket *client)
...
 		slog_noise(client, "linking client to S-%p", server);
 		client->link = server;
 		server->link = client;
 		server->pool->stats.server_assignment_count++;
+
+		msg = pktbuf_temp();
+		HASH_ITER(hh, server->pool->welcome_vars, lk, tmp) {
+			if (lk->value != NULL)
+				pktbuf_write_ParameterStatus(msg, lk->name, lk->value);
+		}
```
JelteF, line 963: Doing this every link seems quite excessive.

JelteF (commented):
Thanks a lot for the contribution. Could you update the PR description with a description of the reasoning why the changes you made were necessary. It currently only describes what is done, but not really why. And that is a bit hard to grasp for me.

Also, is it possible to add some tests for this?

JelteF:
To run tests you can follow the instructions here: https://github.com/pgbouncer/pgbouncer/blob/master/test/README.md#test_xxxpy

include/varcache.h
```diff
@@ -15,13 +15,28 @@ struct VarCache {
 	struct PStr **var_list;
 };
 
+struct WelcomeVarLookup {
+	char *name;			/* key (string is WITHIN the structure) */
+	char *value;		/* value (string is WITHIN the structure) */
+	UT_hash_handle hh;	/* makes this structure hashable */
+};
```
JelteF, line 22: Instead of having this be "welcome vars". I think instead we should maybe support syncing of read-only parameters. More details in this comment, let's keep the discussion there because it impacts both PRs: https://github.com/pgbouncer/pgbouncer/pull/1179#discussion_r1881016656

JelteF:
Given the PR with the other approach has gone a bit stale maybe we can go with this approach for now. It still needs tests though.

PR #1208, opened 2024-11-14
Yuhan  patch pgboucner rr

yuhan-zhang-zip:


eulerto:
Could you explain what is the goal of this patch? If this patch is related to the [pgbouncer-fast-switchover](https://github.com/awslabs/pgbouncer-fast-switchover) repository, I have strong objections.

* License: there are multiple licenses and copyright into the patch. I'm not a lawyer but I wouldn't like to see multiple licenses (Apache license, PostgreSQL license, Amazon license) in this project. The current license is a [ISC license](https://github.com/pgbouncer/pgbouncer/blob/master/COPYRIGHT).
* Routing: PgBouncer is a connection pooler and that's it. It is **not** a routing tool and I wouldn't like to add this capability. The main reason is that it won't be fast and waste some cycles deciding which server it should send the query. Besides, that it is not clear how it works for session pooling mode. If it is the repository that I linked above, it seems it will decide the server **per query** which will only work for statement pooling mode. It is an unacceptable limitation.
* Query rewrite: again PgBouncer is a simple connection pooler. It does not manipulates query or even parse it. Implementing an SQL parser is not a goal of this project. It is not an easy task and it will certainly slowdown PgBouncer quite a bit.
* Fast switchover: PgBouncer already has parameters and commands to do a graceful switchover. It doesn't execute the switchover automatically but another tool can orchestrate the switchover (a DNS can be used to avoid changing the pgbouncer.ini). It is not clear how this feature works but if it manipulates the `host` under the hood, it seems dangerous.

My impression is that this proposal is just a matter of "let's provide our fork to the main project because we don't have interest to maintain it anymore". If you have a fork and wants to see some of the cool features from it into the main project, you should at least create separate pull requests for each feature so we can discuss it separately.

PR #1209, opened 2024-11-17
GSS Auth Implementation

AndrewJackson2020:
This PR implements GSS authentication in pgbouncer. This is a very important institutional feature that allows a client to verify its identity to a server (and vice versa) without the client ever sending its password to the server.

Since the underlying gss c functions are network blocking and synchronous I used the threaded design common to the PAM implementation and the LDAP PR #731 use for this functionality. For the CI and testing aspects of this PR I found the GSS backend encryption PR #743 helpful as well.

Sources:
- [GSS auth postgres docs](https://www.postgresql.org/docs/current/gssapi-auth.html)
- [Libpq backend GSS implementation](https://github.com/postgres/postgres/blob/master/src/backend/libpq/auth.c)


doc/config.md
```diff
@@ -527,6 +527,8 @@ Database name in the `[database]` section to be used for authentication purposes
 option can be either global or overridden in the connection string if this parameter is
 specified.
 
+### krb_server_keyfile
```
JelteF, line 530: I guess you forgot to fill in the docs

AndrewJackson2020: sorry about that will fix.

JelteF:
Just as with the LDAP PR it would be very helpful if someone that uses GSS could try this PR out. I can do a review to see if the code looks sensible, but I don't really know enough about GSS to do an indepth review of this (nor do I have the time to learn).

JelteF:
FYI while cleaning up old PRs I found this one: https://github.com/pgbouncer/pgbouncer/pull/441

It would probably be good to check if there are any good ideas in there.

doc/config.md
AndrewJackson2020: Added documentation to the PR.

AndrewJackson2020:
Expanded the test coverage and feature set. Tried to match all of the GSS options that are available in postgres (case insensitive match, realm match, etc). I believe that I have covered currently with the exception of user mapping.

src/client.c
```diff
@@ -151,6 +152,10 @@ static bool send_client_authreq(PgSocket *client)
 		SEND_generic(res, client, PqMsg_AuthenticationRequest, "i", AUTH_REQ_PASSWORD);
 	} else if (auth_type == AUTH_TYPE_SCRAM_SHA_256) {
 		SEND_generic(res, client, PqMsg_AuthenticationRequest, "iss", AUTH_REQ_SASL, "SCRAM-SHA-256", "");
+#ifdef HAVE_GSS
+	} else if (auth_type == AUTH_TYPE_GSS) {
+		SEND_generic(res, client, 'R', "i", AUTH_REQ_GSS);
```
zhaorui20005, line 157: Better to use PqMsg_AuthenticationRequest for value 'R'.

zhaorui20005 (commented):
I believe this is another quite important feature. Hope I can provide some helps.

src/client.c
AndrewJackson2020: Yeah this is a good point.

src/gss.c
```diff
@@ -0,0 +1,594 @@
...
+	safe_strcpy(request->username, client->login_user_credentials->name, MAX_USERNAME);
+// safe_strcpy(request->password, passwd, MAX_PASSWORD);
+	request->context = &client->gss.ctx;
+	request->flags = &client->gss.flags;
+
+	free_gss_parameters(request);
+
+	gss_first_free_slot = next_free_slot;
+
+	pthread_mutex_unlock(&gss_queue_tail_mutex);
```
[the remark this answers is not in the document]
AndrewJackson2020: This should be fixed now.

src/gss.c
```diff
@@ -0,0 +1,594 @@
...
+		 * sockets and thus save some time.
+		 */
+		if (!is_valid_socket(request)) {
+			log_debug("gss_auth_worker(): invalid socket in slot %d", current_slot);
+			request->status = GSS_STATUS_FAILED;
+			continue;
+		}
+
+		if (gss_check_passwd(request)) {
+			request->status = GSS_STATUS_SUCCESS;
```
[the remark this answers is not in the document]
AndrewJackson2020: This should be fixed now. Went ahead and tried to implement this as closely as possible to your LDAP PR.

zhaorui20005:
> I believe this is another quite important feature. Hope I can provide some helps.

I will continue reading the logic of GSS as it differs a lot for LDAP, I need sometime to compare this with postgres when I got time. And maybe test on my local vm.

src/gss.c
```diff
@@ -0,0 +1,619 @@
...
+		/* We have at least one request in the queue */
+		request = &gss_auth_queue[current_slot];
+		current_slot = (current_slot + 1) % GSS_REQUEST_QUEUE_SIZE;
+
+		/* If the socket is already in the wrong state or reused then ignore it.
+		 * This check is not safe and should not be trusted (the socket state
+		 * might change exactly after it), but it helps to quickly filter out invalid
+		 * sockets and thus save some time.
+		 */
+		if (!is_valid_socket(request)) {
```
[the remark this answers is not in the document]
AndrewJackson2020: Does this issue apply to your LDAP PR as well or did you do something there to mitigate the issue?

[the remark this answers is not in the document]
zhaorui20005: I just removed the calling here as I think this is not a must check.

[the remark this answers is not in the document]
AndrewJackson2020: Okay I similarly removed the above mentioned call to `is_valid_socket`. This sounds like another item that should be implemented in a separate PR for PAM as well.

AndrewJackson2020:
Just a heads up: opened PR #1275 in order to keep track of some of these thread safety concerns on the PAM side.

FF-Ninja:
Hi, can I ask what the current state of this feature is? I am interested in using this

AndrewJackson2020:
> Hi, can I ask what the current state of this feature is? I am interested in using this

The  state of this is "it works on my machine". Right now this PR needs reviews, feedback, and testing from people who are interested in this feature. Please feel free to build this from source, test it for your use case and report any issues that you find, would be greatly appreciated.

FF-Ninja:
> > Hi, can I ask what the current state of this feature is? I am interested in using this
> 
> The state of this is "it works on my machine". Right now this PR needs reviews, feedback, and testing from people who are interested in this feature. Please feel free to build this from source, test it for your use case and report any issues that you find, would be greatly appreciated.

Great thanks. Will build and test it

AndrewJackson2020:
> Great thanks. Will build and test it

Appreciate it, thank you.

test/conftest.py
```diff
@@ -17,6 +20,75 @@
...
+KDC_KADMIN_SERVER = socket.gethostname()
+
+LOGDIR = "log"
+PG_LOG = f"{LOGDIR}/krb.log"
+# Assumes packages are installed; krb5-kdc and krb5-admin-server on debian
+KADMIN_PRINCIPAL_FULL = f"{KADMIN_PRINCIPAL}@{REALM}"
+MASTER_PASSWORD = "master_password"
+
+
+def setup_krb():
```
zhaorui20005, line 36: Where is this function called?

src/objects.c
```diff
@@ -45,6 +45,14 @@ struct AATree user_tree;
  */
 struct AATree pam_user_tree;
 
+/*
+ * All GSS users are kept here. We need to differentiate two user
+ * lists to avoid user clashing for different authentication types,
+ * and because gss_user_tree is closer to PgDatabase.user_tree in
+ * logic.
+ */
+struct AATree gss_user_tree;
```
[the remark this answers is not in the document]
AndrewJackson2020: Huh, not sure why I did it like that. Will review what you did in LDAP and what exists with PAM and will modify this accordingly.

test/conftest.py
AndrewJackson2020: Its probably not called anymore. I originally set up kerberos in python during tests but decided python tests shouldn't do anything with su, that should all be on the container setup. Will remove.

src/gss.c
```diff
@@ -0,0 +1,608 @@
...
+
+	gbuf.length = request->token_length;
+	gbuf.value = request->token;
+
+	log_debug("Received GSSAPI token.");
+	server_credentials = GSS_C_NO_CREDENTIAL;
+	delegated_creds = GSS_C_NO_CREDENTIAL;
+
+	/* Accept the context. */
+	maj_stat = gss_accept_sec_context(&acc_sec_min_stat,
```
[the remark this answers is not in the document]
AndrewJackson2020: Correct, good catch.

AndrewJackson2020:
@zhaorui20005 just a heads up, I'm going to try to rewrite the GSS part of this PR entirely to make it as similar as possible to Postgres's GSS implementation. I think this will address 3 of your comments above and make the implementation more maintainable.

test/conftest.py
AndrewJackson2020: Went ahead and removed the unused test assets from this file.

src/gss.c
```diff
@@ -0,0 +1,608 @@
...
+					  cf_auth_gss_accept_delegation ? &delegated_creds : NULL);
+
+	log_debug("Evaluation of GSSAPI token");
+
+	log_debug("gss_accept_sec_context major: %u, "
+		  "minor: %u, outlen: %u, outflags: %x",
+		  maj_stat, min_stat,
+		  (unsigned int) send_tok.length, *request->flags);
+
+	if (GSS_S_COMPLETE != maj_stat) {
```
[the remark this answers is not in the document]
AndrewJackson2020: This should be good after my refactor.

src/gss.c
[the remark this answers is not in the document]
AndrewJackson2020: This should be fixed after my refactor.

src/gss.c
```diff
@@ -0,0 +1,608 @@
...
+					return false;
+				}
+			} else {
+				if (strcasecmp(cp, request->krb_realm) != 0) {
+					log_warning("gss_name_mismatch");	/* Just for the info */
+					return false;
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
AndrewJackson2020: I restructured these if statements to look a lot more like how it functions in the postgres source code. I think this issue should be fixed now.

AndrewJackson2020:
@zhaorui20005 I have completed my response to all of your feedback. Tests are all passing. The entire interaction with the GSS API should now look a lot more like the postgres source code which I believe will make it much easier to review. 

When you get a chance please take a look and let me know if you have additional feedback.

src/gss.c
```diff
@@ -0,0 +1,607 @@
...
+
+	gbuf.length = request->token_length;
+	gbuf.value = request->token;
+
+	log_debug("Received GSSAPI token.");
+	server_credentials = GSS_C_NO_CREDENTIAL;
+	delegated_creds = GSS_C_NO_CREDENTIAL;
+
+	/* Accept the context. */
```
zhaorui20005, line 442: The overall logic looks fine for me. 
But only one question from here, in Postgres, there is a loop keep calling function `gss_accept_sec_context ()`. But here there is only one round, can you explain why one round of calling is enough, and in which case there will be multiple round of exchange between the client and server?

zhaorui20005:
> @zhaorui20005 I have completed my response to all of your feedback. Tests are all passing. The entire interaction with the GSS API should now look a lot more like the postgres source code which I believe will make it much easier to review.
> 
> When you get a chance please take a look and let me know if you have additional feedback.

The logic looks fine to me now. I think the next step should request @JelteF to review on the basis of these changes.

src/gss.c
AndrewJackson2020: I cannot explain in what case multiple rounds would be required. All of my testing works with only a single round. I imagine it could have to do with a server configuration. I was a hesitant adding the loop without ever having seen a test case where the second iteration would be triggered but I could just go ahead and add the loop anyway to confirm with the postgres implementation.

zhaorui20005: I think there is no need to add the loop if we know the loop is not really needed. Better to know this in more details before make the changes. 
The first time I read the code of Postgres's implementation, I was confused about the loop too.

AndrewJackson2020: So looking at my code it looks like it assumes the auth is complete when `maj_stat == GSS_S_CONTINUE_NEEDED` whereas in the postgres codebase it goes on to do another iteration until `maj_stat == GSS_S_COMPLETE`. In manual testing I do not see instances where it actually does want to continue. I think I need to understand what causes gss to want to continue in order to move forward with this PR.

zhaorui20005: So it is better to add that loop for better portablility?

AndrewJackson2020: Sounds like it, though I'm not sure exactly what portability means in this context. Also I think if we were to modify this PR to enable the multiple rounds of token exchange it wouldn't be a loop locally in the GSS.c file but it would return to the main sequence. This makes me more concerned about implementing it without test coverage.

JelteF:
FYI this has a bunch of merge conflicts now

JelteF:
Also same change as in #1297 should be applied here

AndrewJackson2020:
> Also same change as in #1297 should be applied here
Changed the log_debug to log_warning.

> FYI this has a bunch of merge conflicts now
These were all due to the LDAP PR getting merged. These merge conflicts are all resolved now.

src/gss.c
```diff
@@ -0,0 +1,607 @@
...
+
+	request->token = token;
+	request->token_length = token_length;
+
+	request->client = client;
+	request->connect_time = client->connect_time;
+	request->status = GSS_STATUS_IN_PROGRESS;	/* This is protected by gss_queue_tail_mutex */
+	memcpy(&request->remote_addr, &client->remote_addr, sizeof(client->remote_addr));
+	safe_strcpy(request->username, client->login_user_credentials->name, MAX_USERNAME);
+// safe_strcpy(request->password, passwd, MAX_PASSWORD);
```
zhaorui20005, line 314: The password is not needed, so we should remove this line?

AndrewJackson2020: Correct, just pushed an update with this line removed.

src/gss.c
zhaorui20005: So you are worried about the loop may block the whole working thread?
If you did not add that loop, maybe the authentication should fail when return value is `GSS_S_CONTINUE_NEEDED `.
But I think it is better to add that loop, as we now do not know in which case the loop may happen or even block.

src/gss.c
```diff
@@ -0,0 +1,606 @@
...
+	char *princ;
+
+	maj_stat = gss_display_name(&min_stat, request->gss_name, &gbuf, NULL);
+
+	if (maj_stat != GSS_S_COMPLETE) {
+		log_warning("retrieving GSS user name failed major: %u, minor: %u", maj_stat, min_stat);
+		return false;
+	}
+
+	princ = malloc(gbuf.length + 1);
```
zhaorui20005, line 519: After a little testing, I found that the memory that `princ` points to be set to 0, otherwise it will fail because of the last character not match.

AndrewJackson2020: Okay, that's not good. I thought that I tested every case around this. Will take another look.

src/client.c
```diff
@@ -588,6 +624,20 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const
 		if (!check_user_connection_count(client)) {
 			return false;
 		}
+#ifdef HAVE_GSS
+	} else if (check_if_need_gss_authentication(client, dbname, username) || cf_auth_type == AUTH_TYPE_GSS) {
+		client->login_user_credentials = find_or_add_new_global_credentials(username, NULL);
+		if (!check_db_connection_count(client))
+			return false;
+		if (!client->login_user_credentials) {
+			slog_error(client, "set_pool(): failed to allocate new PAM user");
```
zhaorui20005, line 633: `PAM` to `GSS`

src/client.c
```diff
@@ -1359,6 +1409,25 @@ static bool handle_client_startup(PgSocket *client, PktHdr *pkt)
...
+		} else if (client->client_auth_type == AUTH_TYPE_GSS) {
+			uint8_t *token;
+			uint32_t length;
+
+			length = mbuf_avail_for_read(&pkt->data);
+			/*
+			 * Get the token that the client sends as part of the context
+			 * initialization process.
+			 */
+			if (!mbuf_get_bytes(&pkt->data, length, (const uint8_t **)&token)) {
```
zhaorui20005, line 1421: If no disconnect_client() is called, how to handle this error in the following procedure?

include/gss.h
```diff
@@ -0,0 +1,42 @@
...
+#include <gssapi/gssapi_ext.h>
+#include <gssapi/gssapi_krb5.h>
+#endif
+
+/* Name of the service to be passed to GSS */
+#define PGBOUNCER_GSS_SERVICE "pgbouncer"
+
+/*
+ * Defines how many authentication requests can be placed to the waiting queue.
+ * When the queue is full calls to pam_auth_begin() will block until there is
```
zhaorui20005, line 35: `pam` to `gss`

src/gss.c
zhaorui20005: I re-read the code, it is difficult for pgbouncer to receive the token in gss.c
So I am OK with current state of the code now.
If there are bad cases reported, then we can think of other ways to add that loop.

include/gss.h
AndrewJackson2020: Good catch will fix this.

src/gss.c
AndrewJackson2020: Okay sounds good. I do think that a return of `GSS_S_CONTINUE_NEEDED` should lead to a failure though. Will update this next week.

zhaorui20005: In the current function of `gss_recvauth()`. There are 3 resources that needs to be cleared after return, as without the loop, some resources will be left uncleared.
The `server_credentials`,calling of `gss_delete_sec_context` and `gss_release_buffer(&lmin_s, &send_tok)`.
The release of `send_tok` is handled properly, but "`server_credentials`,calling of `gss_delete_sec_context`" are not, as there are return statement before clearing these 2 resources.
So I suggest to handle them properly in your next update.

src/client.c
AndrewJackson2020: I'm not sure about this one to be honest. Looking at the other calls in `handle_client_startup` 2 of them do not call disconnect_client explicitly and one of them does.

include/gss.h
AndrewJackson2020: This is fixed now.

src/client.c
AndrewJackson2020: This is fixed now.

src/gss.c
AndrewJackson2020: I changed this from malloc (uninitialized) to calloc (zero initialized). Let me know if you think this works.

src/gss.c
AndrewJackson2020: So I modified the function so that if `GSS_S_CONTINUE_NEEDED` is received then authentication is failed. I think it is better to do it this way, at least until we figure out how to incorporate the continue procedure and get a test to trigger it.

Regarding the memory deallocation, I tried to follow the postgres code as closely as possibly with this. Is this an issue with postgres as well? Also maybe we should consider adding gss (and maybe ldap) to the valgrind test pipeline?

zhaorui20005: It is better to have a valgrind testing result if possible.

src/gss.c
```diff
@@ -0,0 +1,607 @@
...
+		pktbuf_static(&_buf, _data, sizeof(_data));
+		pktbuf_put_char(&_buf, 'R');
+		pktbuf_put_uint32(&_buf, send_tok.length + 4 + 4);
+		pktbuf_put_uint32(&_buf, AUTH_REQ_GSS_CONT);
+		pktbuf_put_bytes(&_buf, send_tok.value, send_tok.length);
+
+		if (!pktbuf_send_immediate(&_buf, request->client)) {
+			gss_release_buffer(&lmin_s, &send_tok);
+			log_debug("Failed to send gss token to client");
+			return false;
```
zhaorui20005, line 476: In postgres, there is no such return. So if this return happens, I think there is a need to call `gss_delete_sec_context(&lmin_s, &request->context, GSS_C_NO_BUFFER);` and `gss_release_cred(&min_stat, &server_credentials);` if the if block is met.

src/gss.c
```diff
@@ -0,0 +1,607 @@
...
+		}
+		gss_release_buffer(&lmin_s, &send_tok);
+	}
+
+	if (maj_stat != GSS_S_COMPLETE && maj_stat != GSS_S_CONTINUE_NEEDED) {
+		gss_delete_sec_context(&lmin_s, &request->context, GSS_C_NO_BUFFER);
+		log_debug("accepting GSS security context failed major: %u, minor: %u",
+			  maj_stat, min_stat);
+		return false;
```
zhaorui20005, line 485: The same necessary checking for calling `gss_release_cred(&min_stat, &server_credentials);` if the return happens. Or if this is not met, then there is no need to call `gss_release_cred(&min_stat, &server_credentials);` ?
I am not 100% sure for that.

src/gss.c
zhaorui20005: Yes, it works.

AndrewJackson2020:
> @JelteF @zhaorui20005 Is there any way I can help to add this feature in the pgbouncer? This would be great to add. Thanks!

Any help testing would be greatly appreciated. Looks like this PR needs a rebase as well but thats on me.

zhaorui20005:
> > @JelteF @zhaorui20005 Is there any way I can help to add this feature in the pgbouncer? This would be great to add. Thanks!
> 
> Any help testing would be greatly appreciated. Looks like this PR needs a rebase as well but thats on me.

I would like to give an approval to this PR. I have tested with the testing cases, it all passed.

AndrewJackson2020:
> > @JelteF @zhaorui20005 Is there any way I can help to add this feature in the pgbouncer? This would be great to add. Thanks!
> 
> Any help testing would be greatly appreciated. Looks like this PR needs a rebase as well but thats on me.

This has been rebased.

AndrewJackson2020:
> > > @JelteF @zhaorui20005 Is there any way I can help to add this feature in the pgbouncer? This would be great to add. Thanks!
> > 
> > 
> > Any help testing would be greatly appreciated. Looks like this PR needs a rebase as well but thats on me.
> 
> I would like to give an approval to this PR. I have tested with the testing cases, it all passed.

The major reservation that I still have with this architecture is that the man page for GSS says that the auth should be run in a loop with the expectation of multiple back and forth messages. In testing I have never been able trigger the necessity of the extra message. I also am not sure how this loop type flow could be implemented in the pgbouncer async framework, all of the other auth types are much more open/close compared to GSS.

[0]: https://man.freebsd.org/cgi/man.cgi?query=gss_accept_sec_context&sektion=3&apropos=0&manpath=FreeBSD+7.1-RELEASE

zhaorui20005:
> > > > @JelteF @zhaorui20005 Is there any way I can help to add this feature in the pgbouncer? This would be great to add. Thanks!
> > > 
> > > 
> > > Any help testing would be greatly appreciated. Looks like this PR needs a rebase as well but thats on me.
> > 
> > 
> > I would like to give an approval to this PR. I have tested with the testing cases, it all passed.
> 
> The major reservation that I still have with this architecture is that the man page for GSS says that the auth should be run in a loop with the expectation of multiple back and forth messages. In testing I have never been able trigger the necessity of the extra message. I also am not sure how this loop type flow could be implemented in the pgbouncer async framework, all of the other auth types are much more open/close compared to GSS.

The whole logic looks fine for me. I am ok with this PR.

zhaorui20005 (approved):

karankhanchandani:
@JelteF @zhaorui20005 Thanks for the approvals. Is it possible to merge this feature? Please let me know if I can be of any help.. Thanks!

AndrewJackson2020:
> @JelteF @zhaorui20005 Thanks for the approvals. Is it possible to merge this feature? Please let me know if I can be of any help.. Thanks!

There is a memory error in this PR that I still need to figure out.

AndrewJackson2020:
> > @JelteF @zhaorui20005 Thanks for the approvals. Is it possible to merge this feature? Please let me know if I can be of any help.. Thanks!
> 
> There is a memory error in this PR that I still need to figure out.

The memory error in question appears to also be an issue with LDAP and PAM. The only reason CI passes is because those code paths aren't tested with Valgrind. Went ahead and removed the Valgrind check from this PR, will open an issue for existing auth types later with more detail in case this needs to be addressed for existing auth types.

PR #468, opened 2020-02-28
Fix application_name_add_host to handle SET application_name

pinaraf:
The PostgreSQL protocol sends us a message when the application_name
is changed. It's thus possible to use this message to add back the
host and port information in the session.

Old behaviour (transaction pooling):
```
test=> show application_name;
   application_name    
------------------------
psql - 127.0.0.1:52396
(1 row)
test=> set application_name to 'toto';
SET
test=> show application_name;
   application_name    
------------------------
toto
(1 row)
```

New behaviour:
```
test=> show application_name;
   application_name    
------------------------
psql - 127.0.0.1:52396
(1 row)
test=> set application_name to 'toto';
SET
test=> show application_name;
   application_name    
------------------------
toto - 127.0.0.1:52396
(1 row)
```

PR #1218, opened 2024-12-04
Fix application_name_add_host to handle SET application_name

pinaraf:
The PostgreSQL protocol sends us a message when the application_name is changed. It's thus possible to use this message to add back the host and port information in the session.

Old behaviour (transaction pooling):
```
test=> show application_name;
   application_name    
------------------------
psql - 127.0.0.1:52396
(1 row)
test=> set application_name to 'toto';
SET
test=> show application_name;
   application_name    
------------------------
toto
(1 row)
```

New behaviour:
```
test=> show application_name;
   application_name    
------------------------
psql - 127.0.0.1:52396
(1 row)
test=> set application_name to 'toto';
SET
test=> show application_name;
   application_name    
------------------------
toto - 127.0.0.1:52396
(1 row)
```

(This replaces PR #468)

### DIFF
--- a/include/client.h
+++ b/include/client.h
@@ -23,3 +23,5 @@ bool check_db_connection_count(PgSocket *client);
 bool check_user_connection_count(PgSocket *client);
 bool sending_auth_query(PgSocket *client);
 PgDatabase *prepare_auth_database(PgSocket *client) _MUSTCHECK;
+void set_appname(PgSocket *client, const char *app_name);
+
--- a/include/varcache.h
+++ b/include/varcache.h
@@ -25,3 +25,5 @@ void varcache_clean(VarCache *cache);
 void varcache_add_params(PktBuf *pkt, VarCache *vars);
 void varcache_deinit(void);
 void varcache_set_canonical(PgSocket *server, PgSocket *client);
+const char *varcache_get(VarCache *cache, const char *lk);
+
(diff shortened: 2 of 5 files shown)


pinaraf:
I fixed the formatting issue spotted by uncrustify

JelteF:
Can you add a test for this? Here's docs on how our testing works: https://github.com/pgbouncer/pgbouncer/blob/master/test/README.md

JelteF:
I played around with this a bit, and it definitely works better than the current master branch. But the behaviour also a bit more confusing. Because if you set the application_name inside session mode, then it doesn't work. Nor does the ip address get added to the application_name in the transaction that the application_name is set by the user, it only gets added for the transaction that follows.

I don't think there's an easy way around this, and in practice I think this new behaviour will be very nice for a lot of people. But I do think that we should add these caveats in the documentation for `application_name_add_host`.

JelteF:
Apart from my feedback above, this also needs some tests.

JelteF:
@pinaraf do you still intend to add the tests and docs changes? The code in this PR is good.

mNTTAliev:
This will not work correctly if the SET is performed within rollback transactions (In transaction pooling mode).

When rolling back a transaction, PostgreSQL will send ParameterStatus packet with the old application_name value. This application_name already has an IP address added. Therefore, you will have a double concatenation of the IP address (or triple concatenation, depends on the number of rollbacks).

For example:
image

PR #1216, opened 2024-12-03
Install pkg-config in Windows CI

greg-rychlewski:
This looks like it is solving the error that has been occurring in the recent Windows CI runs.

For reference, the errors are of this form:

```
C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build>call sh -l -c "./autogen.sh" 
Using autoconf: 2.72
Using aclocal: 1.17
ac-wrapper: autoheader: warning: auto-detected versions not found ( ); falling back to latest available
ac-wrapper: autoconf: warning: auto-detected versions not found ( ); falling back to latest available
configure.ac:15: warning: The macro 'AC_PROG_CC_STDC' is obsolete.
configure.ac:15: You should run autoupdate.
../autoconf-2.72/lib/autoconf/c.m4:1669: AC_PROG_CC_STDC is expanded from...
lib/m4/usual.m4:69: AC_USUAL_PROGRAM_CHECK is expanded from...
configure.ac:15: the top level
configure.ac:17: error: possibly undefined macro: PKG_PROG_PKG_CONFIG
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
configure.ac:58: error: possibly undefined macro: PKG_CHECK_MODULES

C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build>if 1 NEQ 0 exit /b 1 
```

PR #1220, opened 2024-12-08
Disable error message test on Windows

JelteF:
The new error message test introduced by #1152 is failing on Windows.
This slipped past CI, because Windows CI was broken completely
before #1216.

For now this disables the test on Windows to make CI green, but it might
be nice to actually fix the behaviour.


JelteF:
cc: @AndrewJackson2020

JelteF:
Interestingly this is also failing on the Rocky Linux CI runs (which we don't run by default).

PR #1224, opened 2024-12-16
Fix version  in configure.ac

AndrewJackson2020:
I noticed that in the most recent build the version number still displays as v1.23.0. I believe that the below number should have been incremented with the last update.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,18 @@ PgBouncer changelog
 
 PgBouncer 1.23.x
 ----------------
+**2024-08-02 - PgBouncer 1.23.1 - "Everything is put back in order"**
+
+- Fixes
+  * Fix a possible segmentation fault after PgBouncer reloads its configuration.
+    (#1105) (bug introduced in 1.23.0)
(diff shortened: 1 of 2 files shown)


AndrewJackson2020:
> Yes, but we will increment it again when we create a new release.

Do you want this PR closed then? I also notice that th NEWS.md  file was not updated.

eulerto:
IMO we should always bump a new version when we cut a new release so the reported version in a development build will reflect the reality. I mean, if you are building from git, the current version is always ahead of the latest stable version.

Yeah, we should have pushed the commits from the 1.23.1 into master. Could you keep the exact change into this PR? I noticed the order is different.

`git diff stable-1.23~2..stable-1.23`

AndrewJackson2020:
Reverted the changes and applied the diff that you provided.

I agree that he building a development build should not yield the same version number as the official release. Although I also don't think it should yield the same version number of the next release as that could be misleading as well. I have seen some projects use git hashes for development builds. Let me know if you want a different version number in this PR.

eulerto (approved):

PR #1229, opened 2024-12-20
Add git hash to version number

AndrewJackson2020:
This PR adds a git hash to the pgbouncer version number. Currently if you build pgbouncer from source the only identifier available is the previous version number which is very misleading as you may be building a version based off of a totally different commit than the last release. This PR would decrease ambiguity around situations like these.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 dnl Process this file with autoconf to produce a configure script.
 
 AC_INIT([PgBouncer],
-        [1.23.1],
+         [m4_esyscmd([./version_script.sh])],
         [https://github.com/pgbouncer/pgbouncer/issues], [],
         [https://www.pgbouncer.org/])
 AC_CONFIG_SRCDIR(src/janitor.c)
--- a/version_script.sh
+++ b/version_script.sh
@@ -1,1 +1,7 @@
-echo 1.23.1-$(git describe --always --abbrev=5 --exclude "*" || echo "unknown" )
+git_hash=$(git describe --always --abbrev=5 --exclude "*" || echo unknown)
+numeric_version="1.23.1"
+if [[ $git_hash == "unknown" ]]; then
+  printf $numeric_version
+else
+  printf $numeric_version-$git_hash
+fi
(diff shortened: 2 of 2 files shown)


df7cb:
Fwiw these "git in version number" things are a major PITA for packaging. Packages are often built from a different git repo, and the hash in there might not have anything to do with the upstream version. Or there might not be any .git directory at build time.

The proposed change looks like it would handle that case gracefully, though.

AndrewJackson2020:
> Fwiw these "git in version number" things are a major PITA for packaging. Packages are often built from a different git repo, and the hash in there might not have anything to do with the upstream version. Or there might not be any .git directory at build time.
> 
> The proposed change looks like it would handle that case gracefully, though.

Interesting. For anyone that builds from source not having the git in version number a PITA. 

Thanks for mentioning this though. Will keep in mind the packaging stakeholders.

eulerto:
I tend to agree with @df7cb . I thought about this for a while and proposed #1278 . We don't need to identify the exact commit we built PgBouncer; instead, a version that says it is a development version for the next release is sufficient. One downside of the referred proposal (#1278) is that we never know if the next version will change the major and/or minor version, hence, we should always increment the minor version but the next one can be a major version. There wasn't a release that increments the major and it means this proposal is almost accurate; some releases the version with "devel" suffix might not match a real release.

AndrewJackson2020:
> I tend to agree with @df7cb . I thought about this for a while and proposed #1278 . We don't need to identify the exact commit we built PgBouncer; instead, a version that says it is a development version for the next release is sufficient. One downside of the referred proposal (#1278) is that we never know if the next version will change the major and/or minor version, hence, we should always increment the minor version but the next one can be a major version. There wasn't a release that increments the major and it means this proposal is almost accurate; some releases the version with "devel" suffix might not match a real release.

Feel free to close this PR if you prefer the approach taken in #1278.

AndrewJackson2020:
Closing as it seems like this negatively impacts packagers. Possibly the `-devel` suffix is the better path forward.

PR #1230, opened 2024-12-23
Output diff of linter failures in 'make check'

cosgroveb:
It would improve my workflow to see the diffs `black` and `isort` are
prescribing while running `format-check`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,8 +194,10 @@ lint:
 	flake8
 
 format-check: uncrustify
-	black --check .
-	isort --check .
+	black --check --diff .
+	isort --check --diff .
+
+Co-authored-by: Brian Cosgrove <bcosgrove@paypal.com>
 	./uncrustify -c uncrustify.cfg --check include/*.h src/*.c -L WARN
 
 format: uncrustify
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -194,8 +194,10 @@ lint:
 	flake8
 
 format-check: uncrustify
-	black --check .
-	isort --check .
+	black --check --diff .
+	isort --check --diff .
+
+Co-authored-by: Brian Cosgrove <bcosgrove@paypal.com>
```
JelteF, line 200:
```suggestion
```

JelteF (approved):
Seems fine to me, I'm assuming this is useful to see what CI complains about?

Makefile
JelteF, line 200: These lines seem to have been added by accident
```suggestion
```

cosgroveb:
Yes because unfortunately I am so far having trouble running black on my very ancient distro (stretch) of Debian that I typically do development in.

It would be better to catch failures before pushing. cirrus cli + this change helps me with that.

Makefile
cosgroveb: My mistake.

PR #1228, opened 2024-12-20
Add user specific reserve pool size

greg-rychlewski:
Closes https://github.com/pgbouncer/pgbouncer/issues/166

PR #1232, opened 2024-12-23
Change db param to reserve_pool_size

greg-rychlewski:
Based on discussion here: https://github.com/pgbouncer/pgbouncer/pull/1228

Updating the database parameter name from `reserve_pool` to `reserve_pool_size` to be consistent with everything else. 

Do you prefer to advertise the deprecation through the release notes only or also make a note in the parameter's documentation?

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -1249,7 +1249,7 @@ Only enforced if at least one of the following is true:
 ### reserve_pool_size
 
 Set additional connections for this database. If not set, the global `reserve_pool_size`
-is used.
+is used. For backwards compatibilty reasons `reserve_pool` is an alias for this option.
 
 ### connect_query
 
--- a/doc/config.md
+++ b/doc/config.md
@@ -1246,7 +1246,7 @@ Only enforced if at least one of the following is true:
 * there is at least one client connected to the pool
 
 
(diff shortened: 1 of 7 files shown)


src/loader.c
```diff
@@ -338,6 +338,8 @@ bool parse_database(void *base, const char *name, const char *connstr)
 			min_pool_size = atoi(val);
 		} else if (strcmp("reserve_pool", key) == 0) {
```
JelteF, line 339: Would be good to add a comment here that we continue supporting this for backwards compatability.

doc/config.md
```diff
@@ -1246,10 +1246,10 @@ Only enforced if at least one of the following is true:
 * there is at least one client connected to the pool
 
 
-### reserve_pool
+### reserve_pool_size
 
-Set additional connections for this database. If not set, `reserve_pool_size` is
-used.
+Set additional connections for this database. If not set, the global `reserve_pool_size`
+is used.
```
JelteF, line 1252:
```suggestion
is used. For backwards compatibilty reasons `reserve_pool` is an alias for this option.
```

JelteF (approved):

PR #1233, opened 2024-12-23
Update default health check query to do a null query

AndrewJackson2020:
This PR changes the setup of the `server_check_query` slightly to allow the use of null queries in `server_check_query`. I believe this is slightly more efficient as it does not need to go through the query optimizer.

See this PR from psycopg for a similar change https://github.com/psycopg/psycopg/pull/791


JelteF (changes requested):
I'm not entirely convinced that the benefit this brings is worth changing the meaning of everyone currently using the empty string to mean "don't send server_check_query". i.e. Let's avoid this being a breaking change. Also, instead of introducing a dedicated option for disabling, it probably makes more sense to dp this when server_check_delay is set to -1

How about we have a special sequence like `<empty>` to mean "send an empty query". We could then do the comparison to this special value when parsing the config to avoid the comparison on every query.

AndrewJackson2020:
I 

> I'm not entirely convinced that the benefit this brings is worth changing the meaning of everyone currently using the empty string to mean "don't send server_check_query". i.e. Let's avoid this being a breaking change. Also, instead of introducing a dedicated option for disabling, it probably makes more sense to dp this when server_check_delay is set to -1
> 
> How about we have a special sequence like `<empty>` to mean "send an empty query". We could then do the comparison to this special value when parsing the config to avoid the comparison on every query.

Good points. I went ahead and reversed the changes and implemented kept the interface of blank string disabling query check, <empty> sends the null query check.

etc/pgbouncer.ini
```diff
@@ -180,6 +180,9 @@ auth_file = /etc/pgbouncer/userlist.txt
 ;; When taking idle server into use, this query is run first.
 ;server_check_query = select 1
 
+;; This disables server_check_query check
+;disable_server_check = 0
+
```
JelteF, line 185: This should be removed now.

doc/config.md
```diff
@@ -650,6 +650,8 @@ Simple do-nothing query to check if the server connection is alive.
 
 If an empty string, then sanity checking is disabled.
 
+If `<empty>` then send empty query as sanity check.
+
 Default: `select 1`
```
JelteF, line 655: The default is not changed anymore, was that intended? I think it might be fine to change the default to `<empty>`, but we need some test coverage in that case. Because currently we don't seem to have any `server_check` tests.

src/janitor.c
```diff
@@ -160,7 +160,10 @@ static void launch_recheck(PgPool *pool)
 		/* send test query, wait for result */
 		slog_debug(server, "P: checking: %s", q);
 		change_server_state(server, SV_TESTED);
-		SEND_generic(res, server, PqMsg_Query, "s", q);
+		if (strcmpeq(q, "<empty>"))
```
JelteF, line 163: Instead of doing this check over and over in the janitor, I'd rather change the config parsing to handle this as a special case and store it as an the empty string.

AndrewJackson2020: Agreed. This should be good now.

etc/pgbouncer.ini
AndrewJackson2020: Agreed. Reverted all changes to `etc/pgbouncer.ini`.

doc/config.md
AndrewJackson2020: Changed it back so `<empty>` is now the default. I still owe you some unit tests on this PR though.

AndrewJackson2020:
Looks like the valgrind pipeline picked up some memory errors. Will look into these as well as add unit testing for the functionality.

doc/config.md
AndrewJackson2020: I added a couple of tests and fixed the valgrind errors mentioned below. 

One thing is that I could not validate the checks actually finding a bad connection, only validating a good connection. I tried terminating a connection with pg_terminate_backend but it seems that pgbouncer knew the connection was dead before the health check query. Could possibly use socat or something like that to simulate a network break that pgbouncer might not be able to validate without a health check.

AndrewJackson2020: @JelteF Was able to implement negative tests by using socat. Added this as a dependency in CI and only works with Linux. There might be a better way of doing this. Either way let me know if these tests are what you had in mind or if you know of a better way of doing this.

JelteF (approved):
LGTM, but you accidentally committed a `configure~` file. Can you remove that one? Would be good to add that one to `.gitignore` in a separate PR.

AndrewJackson2020:
> LGTM, but you accidentally committed a `configure~` file. Can you remove that one? Would be good to add that one to `.gitignore` in a separate PR.

Raised #1292 to add this file to .gitignore. Very

PR #1235, opened 2024-12-24
v1.24.0

JelteF:
It's almost half a year since the last release, and quite a bunch of good stuff has been merged since. Let's do another release.

Fixes that should still be merged before releasing:
- [x] #1238 
- [x] #1241

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,86 @@
 PgBouncer changelog
 ===================
 
+PgBouncer 1.24.x
+----------------
+
+**2025-01-10  -  PgBouncer 1.24.0  -  "New year, new bouncer"**
+
+- Features
+  * Add support for `Type=notify-reload` for systemd. This requires systemd
+    version 253 or later. ([#1148])
+  * Add `KILL_CLIENT` command to the admin console. This allows terminating a
+    client connection by force. ([#1147])
+  * Add `max_user_client_connections` setting, both globally and at the user level. ([#1137])
+  * Add `max_db_client_connections` setting, both globally and at the database level. ([#1138])
+  * Add `current_client_connections` counter to `SHOW USERS` and `SHOW DATABASES` output. ([#1137], [#1138])
+  * Add `load_balance_hosts` parameter, to support **not** load balancing between hosts. ([#736])
+  * Expose prepared statement usage counters in `SHOW STATS`. ([#1192])
+  * Add `client_idle_timeout` setting. ([#1189])
+  * Add user level `query_timeout` and `reserve_pool_size`. ([#1180], [#1228])
+  * Enable `pam` authentication support in HBA file. ([#326])
+
+- Changes
+  * Don't recycle connections on RELOAD if TLS config is unchanged. Previously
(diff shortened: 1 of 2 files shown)


NEWS.md
```diff
@@ -1,6 +1,54 @@
...
+    ([#1232])
+
+- Fixes
+  * Handle various unlikely error cases better, such as OOM errors. These could
+    previously cause crashes or memory leaks. ([#1108], [#1101], [#1099], [#1169], [#1202])
+  * Correct default value for `server_tls_sslmode` in sample config file. ([#1133])
+  * Remove mention in docs of invalid alias for `server_tls_protocols`. ([#1155])
+  * Fix bug when using `auth_query` and replication connections together. This
+    bug would cause connection failures in such setups. ([#1166])
+  * Ignore client cancel requests while PgBouncer is configuring server setting. ([#298])
```
JelteF, line 49: The actual links still need to be added for each of the issue numbers.

NEWS.md
```diff
@@ -1,6 +1,81 @@
...
+  * Add user level `query_timeout` and `reserve_pool_size`. ([#1180], [#1228])
+  * Enable `pam` authentication support in HBA file. ([#326])
+
+- Changes
+  * Don't recycle connections on RELOAD if TLS config is unchanged. Previously
+    if you had TLS connections they would all be recycled on RELOAD, which could
+    cause a temporary but serious performance degradation. Now this only
+    happens when the TLS settings are actually changed. ([#1157])
+  * Enable prepared statement support by default, `max_prepared_statements` is
+    now set to 200 by default. ([#1144])
```
eulerto, line 29: Isn't it a good idea to say that this change might break some apps? Maybe it is a good idea to add a link to [FAQ](https://www.pgbouncer.org/faq.html#how-to-use-prepared-statements-with-transaction-pooling).

NEWS.md
```diff
@@ -1,6 +1,81 @@
 PgBouncer changelog
 ===================
 
+PgBouncer 1.24.x
+----------------
+
+**2025-01-01  -  PgBouncer 1.24.0  -  "New year, new bouncer"**
```
AndrewJackson2020, line 7: Just a reminder: this date should be changed when the release happens.

NEWS.md
JelteF: I expanded the changelog entry a bit with some more context and a link to the relevant docs.

But honestly, I don't think this has any greater chance of breaking apps than any other feature we introduce in this release. If you don't use prepared statements, this change is essentially a no-op. And if you did use prepared statements, then your application was almost certainly broken to begin with. Also given the amount of time this has baked and the feedback we've received, I'm much more confident in this change not breaking people than most of the other changes.

JelteF:
Rocky linux tests are failing on the test added by #1152

I'm not blocking the release on this, since the failure is simply not reporting the error to the client (so the behaviour that the previous release had). 

This is also happening on FreeBSD an Windows, so we ignore the test there. I think we want to probably do the same here. Or even better, fix whatever the root cause of the message not appearing is. @AndrewJackson2020 it would be great if you could look into

PR #1238, opened 2024-12-28
Fix bug when decrementing the user connection count

### LINKED ISSUE #1237: User Count not Decrementing Correctly
See below for a reproducible example.

```diff
diff --git a/test/test.ini b/test/test.ini
index ca8afc5..05bb565 100644
--- a/test/test.ini
+++ b/test/test.ini
@@ -61,6 +61,7 @@ client_limit_db_auth_passthrough = port=6666 host=127.0.0.1 dbname=p0 auth_user=
 maxedout = max_user_connections=3
 maxedout2 = max_user_connections=2 max_user_client_connections=2
 maxedout3 = max_user_client_connections=2
+maxedout6 = max_user_client_connections=2
 poolsize1 = pool_size=1
 respoolsize1 = pool_size=1 reserve_pool_size=2
 pswcheck_not_in_auth_file = max_user_client_connections=2
diff --git a/test/test_limits.py b/test/test_limits.py
index 6a03dc4..7c9f0c4 100644
--- a/test/test_limits.py
+++ b/test/test_limits.py
@@ -454,6 +454,23 @@ def test_user_client_count_db_connect_fail(pg, bouncer) -> None:
     assert user["current_client_connections"] == 0


+def test_user_client_count_db_connect_fail_2(pg, bouncer) -> None:
+    test_user = "maxedout6"
+    test_dbname = "user_passthrough"
+
+    users = bouncer.admin("SHOW USERS", row_factory=dict_row)
+    user = next(user for user in users if user["name"] == test_user)
+    assert user["current_client_connections"] == 0
+
+    connect_args = {"dbname": test_dbname, "user": test_user}
+    with pytest.raises(psycopg.OperationalError):
+        _ = bouncer.conn(**connect_args)
+
+    users = bouncer.admin("SHOW USERS", row_factory=dict_row)
+    user = next(user for user in users if user["name"] == test_user)
+    assert user["current_client_connections"] == 0
+
+
 def test_min_pool_size_with_lower_max_user_connections(bouncer):
     # The p0x in test.init has min_pool_size set to 5. This should make
     # the PgBouncer try to create a pool for maxedout2 user of size 5 after a
```

AndrewJackson2020:
Fixes #1237

The user connection count was not being decremented correctly if the user did not exist in the auth file, but did exist in the pgbouncer config.

Relevant docs:

> Note that when auth_file is configured, if a user is defined in this section but not listed in auth_file, pgBouncer will attempt to use auth_query to find a password for that user if auth_user is set. If auth_user is not set, pgBouncer will pretend the user exists and fail to return "no such user" messages to the client, but neither will it accept any provided password.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -490,7 +490,6 @@ bool check_user_connection_count(PgSocket *client)
 
 bool set_pool(PgSocket *client, const char *dbname, const char *username, const char *password, bool takeover)
 {
-	PgGlobalUser *global_user;
 	Assert((password && takeover) || (!password && !takeover));
 
 	/* find database */
@@ -564,6 +563,7 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 			return false;
 
 		if (!client->login_user_credentials || client->login_user_credentials->dynamic_passwd) {
+			PgGlobalUser *global_user;
 			/*
 			 * If the login user specified by the client
 			 * does not exist or if it has no entry in auth_file,
@@ -602,6 +602,14 @@ bool set_pool(PgSocket *client, const char *dbname, const char *username, const 
 			slog_info(client, "no such user: %s", username);
 			client->login_user_credentials = calloc(1, sizeof(*client->login_user_credentials));
 
+			/*
+			 * For users that we are already tracking, we want to
+			 * track this correctly as a connection count. But for
+			 * users that we don't know about at all, we don't want
+			 * to create a new global user. That's why we use
+			 * find_global_user instead of
+			 * find_or_add_new_global_user.
+			 */
 			global_user = find_global_user(username);
 			if (global_user)
 				client->login_user_credentials->global_user = global_user;
--- a/src/client.c
+++ b/src/client.c
@@ -490,6 +490,7 @@ bool check_user_connection_count(PgSocket *client)
 
 bool set_pool(PgSocket *client, const char *dbname, const char *username, const char *password, bool takeover)
 {
+	PgGlobalUser *global_user;
 	Assert((password && takeover) || (!password && !takeover));
 
 	/* find database */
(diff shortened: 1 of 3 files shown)


JelteF:
Would a fix similar to this work: https://github.com/pgbouncer/pgbouncer/pull/1166

AndrewJackson2020:
> Would a fix similar to this work: #1166

Was able to fix this by adding a call to find_or_add_new_global_user.

AndrewJackson2020:
Any chance we can merge this in before the next release (assuming the fix looks reasonable)?

AndrewJackson2020:
> I didn't check this PR but if it is a recent issue, I would say, fix it before release. If it is an old issue, we might postpone it for the next release. Of course, we also need to consider how critical it is. Since it is just a wrong number, it is not a critical one. I would say, fix it if we are sure the patch is correct. I can take a look in a couple of hours.

This is a fix related to #1137 which has not made it to a release yet (new functionality). Also if someone tries to use max_user_client_connections this will cause users to not be able to connect, so in that case its a bit more critical than an incorrect number.

JelteF:
Yeah agreed, planning to look at this soonish. Test looks good, I'm mainly wondering if the fix is in the correct spot.

AndrewJackson2020:
Okay looking at your test my fix just shuffled the issues around a bit. Will look into this.

JelteF:
Hmm, the additional test passed locally on my machine I believe. I might have messed something up right before pushing it.

AndrewJackson2020:
I believe that I fixed it with an if statement and a call to `find_global_user` instead. This way non-existent users aren't being created but users that were created because they had entries in the conf file are assigned. Possibly there is a better way of doing this though.

JelteF (approved):
Added a few comments for clarity. I think this is good now.

AndrewJackson2020:
Merged with main branch. All tests green. lgtm

PR #1242, opened 2025-01-08
Add transaction timeout

AndrewJackson2020:
This PR adds a transaction timeout global setting to pgbouncer. In the current state there is an idle_in_transaction_timeout and a query_timeout but no transaction_timeout. This is meant to combat issues where a user starts a transaction and then proceeds to wait and issue `SELECT 1;` repeatedly thus making the transaction not idle. There is an equivalent setting on the postgres side but it is easily overridden by any user.

Not looking to get this in before the next release but I believe that this functionality could be very useful and just want to put a PR out here for visibility.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -991,6 +991,13 @@ it will be disconnected.  [seconds]
 
 Default: 0.0 (disabled)
 
+### transaction_timeout
+
+If a client has been in "in transaction" state longer,
+it will be disconnected.  [seconds]
+
+Default: 0.0 (disabled)
+
 ### suspend_timeout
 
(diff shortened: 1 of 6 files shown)


AndrewJackson2020:
Added a test here. Tried to do a 2 second timeout at first but it seems like that didn't play nice with the valgrind pipeline so I upped it to 6.

AndrewJackson2020:
Need to update this PR to accomodate the new error reporting in psycopg3.

AndrewJackson2020:
> Need to update this PR to accomodate the new error reporting in psycopg3.

This is fixed now.

JelteF (approved):

PR #1249, opened 2025-01-17
Pin psycopg version

AndrewJackson2020:
This PR attempts to fix CI by pinning the psycopg version. This will make it  a manual process to upgrade psycopg but at least it will not cause CI failures on the master branch anytime psycopg does a release that changes its behaviour in unit tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ pytest
 pytest-asyncio
 pytest-timeout
 pytest-xdist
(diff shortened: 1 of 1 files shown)


AndrewJackson2020:
Looks like all of the test pipelines passed. Let me know if this is a good enough solution or if you want the psycopg version updated to yesterday's release in this PR as well.

PR #1250, opened 2025-01-21
Fix assertion failures introduced by psycopg 3.2.4

cosgroveb:
These failures were previously addressed by @AndrewJackson2020
in #1249 by pinning psycopg to 3.2.3. I became aware of this because I
just re-enabled the python test suite in Debian[1] [2].

This is a good change in psycopg[3]. Let's update the assertions and bump
the version.

Open question around whether or not we should pin here? It would be good if
we can sort these issues out before they block the upload of new
pgbouncer packages to downstream distros (although most folks probably
care more about pgdg?) No worries either way.

[1]: https://www.postgresql.org/message-id/CAOMoQbROhu7WG9LnyjHb1RXJrax_HKKMBbwaou%2Bjps0uZgqHFw%40mail.gmail.com
[2]: https://www.postgresql.org/message-id/CAGKCzNAQKMKJ_BQcF%3DuzcfcM%3Dvx8QYnT%2Bjtzuv_yp%3D76zbhn7g%40mail.gmail.com
[3]: https://github.com/psycopg/psycopg/compare/3.2.3...3.2.4#diff-87f7a7bb7fe8dbed76e16b67b87f27bda7b779b56e7d93ed3350d77ee4a3f217R20-R23

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -271,7 +271,8 @@ async def test_repeated_sigterm(bouncer):
         bouncer.sigterm()
         await bouncer.wait_for_exit()
         with pytest.raises(
-            psycopg.OperationalError, match="server closed the connection unexpectedly"
+            psycopg.OperationalError,
+            match="database removed|server closed the connection unexpectedly",
         ):
             cur.execute("SELECT 1")
         assert not bouncer.running()
@@ -298,7 +299,8 @@ async def test_repeated_sigint(bouncer):
         bouncer.sigint()
         await bouncer.wait_for_exit()
         with pytest.raises(
-            psycopg.OperationalError, match="server closed the connection unexpectedly"
+            psycopg.OperationalError,
+            match="database removed|server closed the connection unexpectedly",
         ):
             cur.execute("SELECT 1")
         assert not bouncer.running()
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -28,7 +28,7 @@ def test_fast_close(bouncer):
 
             with pytest.raises(
                 psycopg.OperationalError,
-                match=r"database configuration changed|Software caused connection abort",
+                match=r"database configuration changed|server closed the connection unexpectedly|Software caused connection abort",
             ):
                 cur.execute("select 1")
 
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -28,7 +28,7 @@ def test_fast_close(bouncer):
 
             with pytest.raises(
                 psycopg.OperationalError,
-                match=r"server closed the connection unexpectedly|Software caused connection abort",
+                match=r"database configuration changed|Software caused connection abort",
             ):
                 cur.execute("select 1")
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ pytest
 pytest-asyncio
 pytest-timeout
 pytest-xdist
-psycopg==3.2.3
+psycopg==3.2.4
 filelock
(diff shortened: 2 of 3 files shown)


AndrewJackson2020:
Thank you for doing this.

I am on team pin. Further I think we should pin every version of every package in the requirements.txt file as long as it passes CI. I think that in an ideal world 2 runs of CI should produce the exact same results every single time, and pinning gets us closer to that.

df7cb:
Fwiw, pinning won't help with Debian, we will be using the package that is present in the release targeted, ignoring any pins.

test/test_misc.py
```diff
@@ -28,7 +28,7 @@ def test_fast_close(bouncer):
 
             with pytest.raises(
                 psycopg.OperationalError,
-                match=r"server closed the connection unexpectedly|Software caused connection abort",
```
df7cb, line 31: Does this still work with older psycopg versions?

cosgroveb:
@df7cb 

> Fwiw, pinning won't help with Debian, we will be using the package that is present in the release targeted, ignoring any pins.

Yeah, that's a more explicit way of stating what I was trying to say here:

> Open question around whether or not we should pin here? It would be good if
we can sort these issues out before they block the upload of new
pgbouncer packages to downstream distros (although most folks probably
care more about pgdg?) No worries either way.

test/test_misc.py
cosgroveb: aa40020480ed03dbbbbb5fd8ceea46e255fcb8c5

cosgroveb:
> I am on team pin. Further I think we should pin every version of every package in the requirements.txt file as long as it passes CI.

This approach works well for internal platform teams that have mirrors and aren't disrupted e.g. when a package is pulled but IMHO it's less realistic for an OSS project being distributed across several OSes and packaging systems. I agree at least that there's a time and place for it.

cosgroveb:
If pinning is more realistic for the pgbouncer project, I can perhaps work with the Debian package maintainer to override that in `debian/patches` though I think we would need to have ~unimpeachable reasons to do that.

df7cb:
Afaict requirements.txt is just ignored by most parts of the Debian toolchain. If you insist on pinning to a version that behaves differently, we will poke you with a stick until the regression tests are working again for Debian and apt.postgresql.org. :grin:

df7cb:
That said, https://jengus.postgresql.org/job/pgbouncer-binaries/102/architecture=amd64,distribution=sid/console is unhappy and we'd appreciate a new release with this PR merged. :smile_cat:

AndrewJackson2020:
Is the idea with removing pins that any upstream packager will always pull the most recent versions of any python packages so the pgbouncer requirements.txt should always pull the most recent version even if that means a successful build yesterday could mean a failed build today?

cosgroveb:
Yes. We should aim for our test suite to be compatible with all of the versions of our dependencies that downstream packagers will use and a build failure tomorrow is precisely how that is accomplished.

JelteF (approved):

JelteF:
> Is the idea with removing pins that any upstream packager will always pull the most recent versions of any python packages so the pgbouncer requirements.txt should always pull the most recent version even if that means a successful build yesterday could mean a failed build today?

Yeah, I think that's a fair assesment. And I think that's a reasonable tradeoff. Packaging is definitely a case where we cannot force a specific python dependency version on people, so we should just make sure that we work with all. And then we'll just have to live with the fact that we need to fix our build sometimes (for which a temporary workaround can be to pin the version).

PR #1259, opened 2025-01-27
fix clang readability warning (inconsistent declarataion parameter name)

mjacobs:
clang-tidy identified a minor naming inconsistency between parameter naming in server.h/server.c

https://clang.llvm.org/extra/clang-tidy/checks/readability/inconsistent-declaration-parameter-name.html


eulerto:
This is not the only inconsistency reported by `readability-inconsistent-declaration-parameter-name`.

```
for f in $(find src -type f); do
  clang-tidy --checks="-*,readability-inconsistent-declaration-parameter-name" $f -- -Iinclude -Ilib -Iuthash/src;
done
```

If we are going to fix this case, it is better to fix other cases too. (I'm not sure about the SCRAM code because it is a copy from upstream.)

PR #1264, opened 2025-02-04
Queue notification

AndrewJackson2020:
This PR adds a configuration item `client_queue_notify_seconds`. By default this is disabled but when set it will alert a client that is awaiting a server connection after they have been queued for the amount of time in `client_queue_notify_seconds` with the message "No server connection available in postgres backend, client being queued".

As is it can be confusing for users that are queued as the terminal just looks blank and appears to hang. This PR allows pgbouncer to provide some visual feedback to the client as to why their query hasn't returned instantly.


src/janitor.c
```diff
@@ -188,7 +189,33 @@ static void per_loop_activate(PgPool *pool)
...
+			&& client->queued_user_notified == 0
+			&& (age / USEC) > cf_client_queue_notify_seconds
+			&& cf_client_queue_notify_seconds > 0) {
+
+			buf = pktbuf_dynamic(256);
+			pktbuf_write_Notice(
+				buf,
+				"No server connection available in postgres backend, client being queued"
+			);
+			pktbuf_write_CommandComplete(buf, "QUEUED");
```
JelteF, line 212: This seems incorrect, no command has completed.

AndrewJackson2020: Let me add some tests here. Thought that I had run this manually and it worked.

AndrewJackson2020: Okay I see what your point was above now. Removed the `pktbuf_write_CommandComplete(buf, "QUEUED");` line from this PR. Also added a unit test.

src/janitor.c
```diff
@@ -188,7 +189,31 @@ static void per_loop_activate(PgPool *pool)
...
+		    && (age / USEC) > cf_client_queue_notify_seconds
+		    && cf_client_queue_notify_seconds > 0) {
+			buf = pktbuf_dynamic(256);
+			pktbuf_write_Notice(
+				buf,
+				"No server connection available in postgres backend, client being queued"
+				);
+			res = pktbuf_send_queued(buf, client);
+			if (!res)
+				log_warning("Queue warning failed");
```
JelteF, line 213:
```suggestion
				log_warning("Sending queue warning failed");
```

doc/config.md
```diff
@@ -776,6 +776,15 @@ syntax and directives.
...
 
+### client_queue_notify_seconds
+
+The number of seconds that a client will be queued for before
+pgbouncer sends a notification message that they are being
+queued.
+
+A value of 0 disables this notification message.
+
+Default: 0
```
JelteF, line 787: I feel like enabling this by default is probably a good idea. How about setting it to 5 seconds by default?

AndrewJackson2020: I enabled it by default but it looks like it might be messing with SCRAM authentication.

AndrewJackson2020:
I was able to come up with a test that deterministically catches the scram issue. I think what is happening is that if the client is queued and receives the queue warning before completing the scram exchange the client is not expecting a message but is instead expecting the next part of the scram exchange which leads to an error. I suspect there will be similar problems with this feature and GSS as well.

 I think the way of fixing this is to just add an if statement so that if the client is currently being authenticated with scram then do not send the notification.

AndrewJackson2020:
I was able to fix this functionality by only sending the notifications after the welcome message has been sent. This requires an extra variable in the PgSocket struct but I believe that this is the correct way to handle this functionality. Not sure why the windows test is failing though.

etc/pgbouncer.ini
```diff
@@ -66,6 +66,10 @@ listen_port = 6432
 ;; processes that are peered together. When set to 0 pgbouncer peering is disabled
 ;peer_id = 0
 
+;;; Notify client that they are queued after this many seconds
+;;; Disabled when set to 0
+;;;client_queue_notify_seconds = 0
```
JelteF, line 71:
```suggestion
;;;client_queue_notify_seconds = 5
```

src/janitor.c
```diff
@@ -188,7 +189,32 @@ static void per_loop_activate(PgPool *pool)
...
 	statlist_for_each_safe(item, &pool->waiting_client_list, tmp) {
+		usec_t age;
+		PktBuf *buf;
+		bool res;
 		client = container_of(item, PgSocket, head);
+		if (client->query_start == 0) {
+			age = now - client->request_time;
+		} else {
+			age = now - client->query_start;
+		}
```
JelteF, line 200: Based on your recent fix, I think this is using the wrong timestamps for this calculation. Instead of `request_time` or `query_start` this should be using `wait_start` to calculate the age. Because we don't care about the connection age, we care about the connection having to wait more than X seconds.

include/bouncer.h
```diff
@@ -662,12 +662,15 @@ struct PgSocket {
...
 
 	bool wait_for_welcome : 1;	/* client: no server yet in pool, cannot send welcome msg */
+	bool welcome_sent : 1;
 	bool wait_for_user_conn : 1;	/* client: waiting for auth_conn server connection */
 	bool wait_for_user : 1;		/* client: waiting for auth_conn query results */
 	bool wait_for_auth : 1;		/* client: waiting for external auth (PAM) to be completed */
 
 	bool suspended : 1;		/* client/server: if the socket is suspended */
 
+	bool queued_user_notified : 1;	/* client: whether user has been alerted that it is queued */
```
JelteF, line 672: I think this could use a better name. It has nothing to with a "user", which is very specific concept inside pgbouncer. I feel like `sent_wait_nofication` would be decent name, but feel free to propose a better one.

doc/config.md
```diff
@@ -776,6 +776,15 @@ syntax and directives.
 
 Default: empty (use operating system defaults)
 
+### client_queue_notify_seconds
```
JelteF, line 779: I think this name should include `wait` instead of `queue`, that's the terminology we use everywhere else. Maybe allign it with `query_wait_timeout` and call it `query_wait_notify`. I think the `seconds` suffix is also unnecessary. All pgbouncer configs use seconds as the unit.

doc/config.md
```diff
@@ -776,6 +776,15 @@ syntax and directives.
...
+
+The number of seconds that a client will be queued for before
+pgbouncer sends a notification message that they are being
+queued.
```
JelteF, line 783: Clarify the unit here rather than in the setting name.
```suggestion
queued. [seconds]
```

JelteF (changes requested):

doc/config.md
AndrewJackson2020: This should be good now.

doc/config.md
AndrewJackson2020: Understood. Went ahead and replaced all references of `client_queue_notify_seconds` to `query_wait_notify`. Also changed the wording of the documentation a bit since its somewhat redundant mentioning seconds there given that every other setting is in seconds and settings is now referenced in square brackets at the end of the paragraph.

include/bouncer.h
AndrewJackson2020: Changed every reference of `queued_user_notified` to `sent_wait_notification`. That name is as good as any I can think of.

src/janitor.c
AndrewJackson2020: This has the helpful side effect of simplifying the code as well. This has been implemented. Also expanded the test coverage a bit.

JelteF:
Has a merge conflict, and FreeBSD and Windows are failing for some reason.

AndrewJackson2020:
> Has a merge conflict, and FreeBSD and Windows are failing for some reason.

Fixed the merge conflict. 

Was able to isolate the windows fail. Was caused by a test that was using functionality that was unavailable in the windows build (socket password less auth). I marked this test as skip on windows and it seems to work now.

I believe the BSD failure was a flaky test.

Let me know if you see any other issues.

src/objects.c
```diff
@@ -240,6 +240,7 @@ void change_client_state(PgSocket *client, SocketState newstate)
 			newstate = CL_LOGIN;
 	/* fallthrough */
 	case CL_WAITING:
+		client->sent_wait_notification = 0;
```
JelteF, line 243:
```suggestion
		client->sent_wait_notification = false;
```

src/janitor.c
```diff
@@ -191,7 +191,26 @@ static void per_loop_activate(PgPool *pool)
 	sv_tested = statlist_count(&pool->tested_server_list);
 	sv_used = statlist_count(&pool->used_server_list);
 	statlist_for_each_safe(item, &pool->waiting_client_list, tmp) {
+		PktBuf *buf;
+		bool res;
 		client = container_of(item, PgSocket, head);
+
+		if (client->state == CL_WAITING
+		    && client->sent_wait_notification == 0
```
JelteF, line 199:
```suggestion
		    && client->sent_wait_notification
```

AndrewJackson2020: Changed this to `&& !client->sent_wait_notification`. Tests are fixed now.

JelteF (approved):

Issue #1222, opened 2024-12-15
Clients in PgBouncer's waiting queue are not cancelled when ^C is sent via psql


mugunthan7285:
When a client connection is in the waiting queue due to reasons such as default_pool_size limits, max_user_connections, or similar constraints, sending ^C (SIGINT) from psql does not cancel the queued connection. This behavior causes the client to remain stuck in the waiting queue until the server pool has resources to process the connection.

**Expected Behaviour**
When ^C is sent in psql, the corresponding client connection should be immediately cancelled and removed from the waiting queue in PgBouncer. This aligns with user expectations for interruption of pending queries or operations.

**Steps to Reproduce**
Configure PgBouncer with a small default_pool_size or max_user_connections that can be easily exceeded.Open multiple psql connections to PgBouncer until some connections are placed in the waiting queue.While one of the connections is in the waiting queue, press ^C in the psql session.Observe that the connection is not removed from PgBouncer's waiting queue.

AndrewJackson2020:
I agree with this and have had similar issues in the past. I think you can terminate your shell process and that will disconnect the client from pgbouncer, even so ideally control-c should work.

That being said I'm not sure if it is possible to implement this in pgbouncer. There is a part of the code (sbuf_pause) that I believe causes pgbouncer to not listen to clients after they have been queued so control c would not reach pgbouncer until they have been queued. Hopefully someone more knowledgeable will tell me that I'm wrong about this.

mugunthan7285:
Hi @AndrewJackson2020 , Thanks for the response. 

Based on my debugging, I observed that when PgBouncer is in a paused state or when clients are in the waiting queue, cancel requests still reach the **_accept_cancel_request_** method.

In this method, it checks whether the **_main_client_** (for which the cancel request is sent) is linked to a server. If the _main_client_ is not linked, the cancel request client(req) is simply disconnected with the following comment and behavior:
```
/*
* The client is not linked to a server, which means that no query is
* running that can be cancelled. This likely means the query finished by
 * itself before the cancel request arived to pgbouncer.
 */
if (!main_client->link) {
	disconnect_client(req, false, "cancel request for idle client");
	return;
}
```
Cancel requests for waiting clients also fall into this category, as they are not linked to a server yet. It might be beneficial to handle these cases more explicitly. For instance, PgBouncer could receive the data, ignore the data, and reset the iobuf of main_client.

[...]

PR #1265, opened 2025-02-05
Handle query cancellation for waiting clients

mugunthan7285:
This fix enables the cancellation of waiting clients. Previously, if a cancel request arrived for a waiting client, it was ignored, causing the client to hang. With this fix, the behavior now aligns with how cancel requests are handled on a PostgreSQL server

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -1368,28 +1368,27 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 	PgClosePacket close_packet;
 
 
-	if(client->wait_for_cancel){
-		
-		switch(pkt->type){
-			case PqMsg_Query:
-			case PqMsg_FunctionCall:
-			case PqMsg_Sync:
-				client->query_start = 0;
-				client->xact_start  = 0;
-				client->wait_for_cancel = false;
-				break;
-			default:
-				return false;
-				break;
+	if (client->wait_for_cancel) {
+		switch (pkt->type) {
+		case PqMsg_Query:
+		case PqMsg_FunctionCall:
+		case PqMsg_Sync:
+			client->query_start = 0;
+			client->xact_start = 0;
+			client->wait_for_cancel = false;
+			break;
+		default:
+			return false;
+			break;
 		}
-		
+
 		sbuf_prepare_skip(sbuf, pkt->len);
-		
-		if (!send_pooler_error(client,true,NULL,false,"Cancelled waiting query")){
-			disconnect_client(client,false,"Failed to cancel waiting query");
+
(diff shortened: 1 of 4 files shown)


src/client.c
```diff
@@ -1367,6 +1367,32 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
...
+			client->query_start = 0;
+			client->xact_start = 0;
+			client->wait_for_cancel = false;
+			break;
+		default:
+			return false;
+			break;
+		}
+
+		sbuf_prepare_skip(sbuf, pkt->len);
```
JelteF, line 1385: This is not enough in case of the extended protocol. In that case we need to consume all messages until the next `Sync`. You should look at some of the pipeline tests that we have and test that everything is popped from the queue.

JelteF (changes requested):
I think this is a useful change, but the current implementation is not handling some important edge cases afaict.

src/client.c
mugunthan7285: Hi @JelteF , Thanks for the review. I’m making the necessary corrections. Meanwhile, during testing, I had another concern. Since we first send an error response before processing the requests and if it is an explicit transaction the client might initiate rollback(). In that case, I believe we also need to handle it; otherwise, those packets could remain pending, waiting for the server connection again.

While testing this with psycopg, I observed that `conn.rollback()` did not send a `ROLLBACK` command because it checks the current transaction status first. However, if some implementations send a ROLLBACK blindly without performing the transaction status check, it could end up being queued. I’m unsure if my understanding is correct—your insights would be helpful.

JelteF: I think sending the rollback to the actual postgres server is not a huge problem. At that point there's no transaction so a rollback is a no-op. I think a bigger problem is when the next Query packet is actually something that was supposed to be part of the transaction. So for instance Pgbouncer receives messages in this order (while no server is available)

1. Query("BEGIN")
2. Query("INSERT INTO t VALUES (1)")
3. CancelRequest()
4. Query("ROLLBACK")

I'm not sure what postgres behaviour is if a "BEGIN" gets cancelled, but I think that should be investigated before we can merge this. Because I think in the approach suggested so far. We would now throw an error for the BEGIN, but then the INSERT would go through to postgres like normal, i.e. *outside of a transaction*. And then the ROLLBACK would do nothing.

AndrewJackson2020:
Just want to make sure there is a link to the originating issue with this PR.

https://github.com/pgbouncer/pgbouncer/issues/1222

JelteF:
@mugunthan7285 do you intend to continue working on this?

Issue #1215, opened 2024-12-02
ERROR TLS setup failed: failed to load CA: (null)


Erilbeth:
Summary:
When client_tls settings were configured liked below, it gives these errors:

2024-12-02 12:12:56.629 UTC [1] ERROR TLS setup failed: failed to load CA: (null)
2024-12-02 12:12:56.629 UTC [1] FATAL TLS setup failed

```
          client_tls_sslmode = require
          client_tls_key_file = /etc/ssl/certs/tls.key
          client_tls_cert_file = /etc/ssl/certs/tls.crt
```

Cert is created by cert-manager(let's encrypt) and verified that it's working on an nginx ingress.

deployment.yaml

```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: pgbouncer-test
  name: pgbouncer-test
  namespace: default
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app: pgbouncer-test
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      labels:
        app: pgbouncer-test
    spec:
      containers:
      - env:
        - name: PGBOUNCER_PORT
          value: "6432"
        - name: POSTGRESQL_USERNAME
          valueFrom:
            secretKeyRef:
              key: PGUSER
              name: pg-credentials
        - name: POSTGRESQL_PASSWORD
          valueFrom:
            secretKeyRef:
              key: PGPASSWORD
              name: pg-credentials
        - name: POSTGRESQL_HOST
          value: 127.0.0.1
        - name: POSTGRESQL_PORT
          value: "5432"
        image: docker.io/bitnami/pgbouncer:latest
        imagePullPolicy: IfNotPresent
        name: pgbouncer-test
        ports:
        - containerPort: 6432
          protocol: TCP
        resources:
          limits:
            cpu: 500m
            memory: 256Mi
          requests:
            cpu: 100m
            memory: 128Mi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /bitnami/pgbouncer/conf/
          name: pgbouncer-config
        - mountPath: /etc/ssl/certs
          name: ssl-certs
        - mountPath: /var/log/pgbouncer
          name: pgbouncer-log
        - mountPath: /var/run/pgbouncer
[...]

PR #1271, opened 2025-02-10
Allow client_tls_sslmode to be used without client_tls_ca_file

tie:
This change allows `client_tls_sslmode` to be used without `client_tls_ca_file` when set to `require`, `prefer` or `allow`.

Fixes https://github.com/pgbouncer/pgbouncer/issues/1215


JelteF:
Looks reasonable. Can you add a test for this?

JelteF:
The other SSL tests can be found here: https://github.com/pgbouncer/pgbouncer/blob/master/test/test_ssl.py

JelteF:
@tie could you still add a test for this?

PR #1278, opened 2025-02-24
Use devel suffix for development

eulerto:
The current version during development is the latest release version. That's not ideal because it confuses the stable release with a development code.

The new policy states that for a release, we remove the "devel" suffix and the first commit after release is to (a) increment the major and/or minor version and (b) add the "devel" suffix back. It means if we release 1.24.0, the next commit after release should update version from 1.24.0 to 1.25devel.


eulerto:
The downside of this proposal is when we change the MAJOR in MAJOR.MINOR.PATCH. We never did that yet and when we do the "devel" reference won't match the real release. I think that is acceptable because a development version won't be released. (I remember when Postgres used to use MAJOR.MINOR.PATCH and 8.5 was the reference for the next release until the community decided to call it 9.0.)

PR #1281, opened 2025-03-04
Add test_load_balance_hosts.py to 'make dist'

cosgroveb:
This was missed in #736


cosgroveb:
Apologies, opened from wrong fork.

PR #1303, opened 2025-04-07
SSL verify-full mode is broken with multi-host configuration.

### LINKED ISSUE #1300: TLS verification fails if using a multi-host endpoint
In another TLS-related issue we recently uncovered, we have an instance of pgbouncer that connects using the multi-host setup to allow for fast failovers. When we tried to turn TLS from `require` to `verify-full`, we get TLS errors related to the hostname not being in the server certificate.  The error string indicates that it's trying to validate the entire comma-separated hostname instead of validating all of those hosts individually.

```
WARNING TLS handshake error: name `<redacted>.cluster-<redacted>.us-east-2.rds.amazonaws.com,<redacted2>.cluster-<redacted>.us-east-2.rds.amazonaws.com,<redacted3>.<redacted>.us-east-2.rds.amazonaws.com' not present in server certificate
```

It's also just emitted as a WARNING, despite failing to connect. Seems like it should at least be ERROR if not CRITICAL.

emelsimsek:
verify-full requires to compare hostname against the server certificate. Looks like we pass the entire multi-host string as the hostname which causes failure with connection establishment.

Fixes #1300

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -693,6 +693,8 @@ struct PgSocket {
 	PgAddr remote_addr;	/* ip:port for remote endpoint */
 	PgAddr local_addr;	/* ip:port for local endpoint */
 
+	char *host;
+
 	union {
 		struct DNSToken *dns_token;	/* ongoing request */
 		PgDatabase *db;			/* cache db while doing auth query */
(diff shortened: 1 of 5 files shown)


jimbobhickville:
This fixes the bug for us. We were able to enable verify-full with the same configuration that previously failed.

emelsimsek:
> This fixes the bug for us. We were able to enable verify-full with the same configuration that previously failed.

@jimbobhickville  Thank you for the verification.

PR #1305, opened 2025-04-17
Enable Admin Console Access During Safe Shutdown

AndrewJackson2020:
In it's current implementation, during safe shutdowns there is no way to access the admin console. This is problematic if you want to log in, see who is on, alert tenants to restart their apps, etc. Currently the only way to gain visibility about who is connected to your app is via `lsof` and other cli tools but even here you lose visibility as to what database they are connected to, what user, etc.

This PR attempts to provide a way for admin users to continue to access the admin console by keeping the Unix Socket open during the safe shutdown period.

Another approach to accomplish this goal would be to create separate sockets/ports that only admins users can log in through. Happy to consider this as an alternate approach is the method currently implemented does not work.

### DIFF
--- a/src/pooler.c
+++ b/src/pooler.c
@@ -35,6 +35,7 @@ struct ListenSocket {
 };
 
 static STATLIST(sock_list);
+static STATLIST(sock_list_shutdown);
 
 /* hints for getaddrinfo(listen_addr) */
 static const struct addrinfo hints = {
@@ -69,6 +70,29 @@ void cleanup_sockets(void)
 
 	while ((el = statlist_pop(&sock_list)) != NULL) {
 		ls = container_of(el, struct ListenSocket, node);
+		if (pga_is_unix(&ls->addr) && cf_unix_socket_dir[0] != '@') {
+			statlist_append(&sock_list_shutdown, &ls->node);
+			// statlist_remove(&sock_list, &ls->node);
+			continue;
+		}
+		if (event_del(&ls->ev) < 0) {
+			log_warning("cleanup_sockets, event_del: %s", strerror(errno));
+		}
+		if (ls->fd > 0) {
+			safe_close(ls->fd);
+			ls->fd = 0;
+		}
+		statlist_remove(&sock_list, &ls->node);
+		free(ls);
+	}
+}
+
+void cleanup_sockets_unix(void)
+{
+	struct ListenSocket *ls;
+	struct List *el;
+	while ((el = statlist_pop(&sock_list_shutdown)) != NULL) {
+		ls = container_of(el, struct ListenSocket, node);
 		if (event_del(&ls->ev) < 0) {
 			log_warning("cleanup_sockets, event_del: %s", strerror(errno));
 		}
@@ -80,8 +104,9 @@ void cleanup_sockets(void)
 			char buf[sizeof(struct sockaddr_un) + 20];
 			snprintf(buf, sizeof(buf), "%s/.s.PGSQL.%d", cf_unix_socket_dir, cf_listen_port);
 			unlink(buf);
+			statlist_remove(&sock_list_shutdown, &ls->node);
 		}
-		statlist_remove(&sock_list, &ls->node);
+		statlist_remove(&sock_list_shutdown, &ls->node);
 		free(ls);
 	}
 }
@@ -575,6 +600,12 @@ bool for_each_pooler_fd(pooler_cb cbfunc, void *arg)
 	bool ok;
 
 	statlist_for_each(el, &sock_list) {
+		ls = container_of(el, struct ListenSocket, node);
+		ok = cbfunc(arg, ls->fd, &ls->addr);
+		if (!ok)
+			return false;
+	}
+	statlist_for_each(el, &sock_list_shutdown) {
 		ls = container_of(el, struct ListenSocket, node);
 		ok = cbfunc(arg, ls->fd, &ls->addr);
 		if (!ok)
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -5,7 +5,13 @@ import time
 import psycopg
 import pytest
 
-from .utils import HAVE_IPV6_LOCALHOST, PG_MAJOR_VERSION, PKT_BUF_SIZE, WINDOWS
+from .utils import (
+    HAVE_IPV6_LOCALHOST,
+    PG_MAJOR_VERSION,
+    PKT_BUF_SIZE,
+    USE_UNIX_SOCKETS,
+    WINDOWS,
+)
 
 
 def test_connect_query(bouncer):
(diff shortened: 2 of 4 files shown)


test/test_misc.py
```diff
@@ -522,6 +529,29 @@ async def test_already_paused_client_during_wait_for_servers_shutdown(bouncer):
                 await task
 
 
+@pytest.mark.skipif(
+    "not USE_UNIX_SOCKETS", reason="This test does not apply to non unix sockets"
+)
+def test_shutdown_wait_for_clients(bouncer):
+    """
+    Test that after issu
```
JelteF, line 537: Partial comment

test/test_misc.py
```diff
@@ -522,6 +529,29 @@ async def test_already_paused_client_during_wait_for_servers_shutdown(bouncer):
                 await task
 
 
+@pytest.mark.skipif(
+    "not USE_UNIX_SOCKETS", reason="This test does not apply to non unix sockets"
+)
+def test_shutdown_wait_for_clients(bouncer):
```
JelteF, line 535: Test is failing on OSX + FreeBSD

test/test_misc.py
```diff
@@ -522,6 +529,29 @@ async def test_already_paused_client_during_wait_for_servers_shutdown(bouncer):
...
+    """
+    Test that after issu
+    """
+    with bouncer.cur() as cur, bouncer.admin_runner.cur():
+        cur.execute("SELECT 1")
+        bouncer.admin("SHUTDOWN WAIT_FOR_CLIENTS")
+
+        time.sleep(2)
+        bouncer.sql(query=";", host=bouncer.config_dir)
```
JelteF, line 544: I think this should still fail. The only thing that shouldn't fail is connecting to the admin console over the UNIX socket. We should also update the docs for `SHUTDOWN WAIT_FOR_CLIENTS` that you can still connect to the admin database over a unix socket after this.

src/pooler.c
```diff
@@ -68,6 +69,29 @@ void cleanup_sockets(void)
 		return;
 
 	while ((el = statlist_pop(&sock_list)) != NULL) {
+		ls = container_of(el, struct ListenSocket, node);
+		if (pga_is_unix(&ls->addr) && cf_unix_socket_dir[0] != '@') {
```
JelteF, line 73: Why add this check for the `@` in the first character?

src/pooler.c
```diff
@@ -68,6 +69,29 @@ void cleanup_sockets(void)
...
+		if (ls->fd > 0) {
+			safe_close(ls->fd);
+			ls->fd = 0;
+		}
+		statlist_remove(&sock_list, &ls->node);
+		free(ls);
+	}
+}
+
+void cleanup_sockets_unix(void)
```
JelteF, line 90:
```suggestion
void cleanup_unix_sockets(void)
```

src/pooler.c
```diff
@@ -68,6 +69,29 @@ void cleanup_sockets(void)
 		return;
 
 	while ((el = statlist_pop(&sock_list)) != NULL) {
```
JelteF, line 71: Probably good to call this function `cleanup_tcp_sockets` now.

test/test_misc.py
```diff
@@ -522,6 +529,29 @@ async def test_already_paused_client_during_wait_for_servers_shutdown(bouncer):
...
+        with pytest.raises(psycopg.errors.OperationalError):
+            bouncer.sql(query=";", host="127.0.0.1")
```
JelteF, line 546: Let's use the `bouncer.test(...)` method instead. Also for the other places. Maybe in another PR we could replace the `select 1` that that does internally with this empty query.
```suggestion
            bouncer.test(host="127.0.0.1")
```

src/pooler.c
```diff
@@ -35,6 +35,7 @@ struct ListenSocket {
 };
 
 static STATLIST(sock_list);
+static STATLIST(sock_list_shutdown);
```
JelteF, line 38: Why do we need this additional list? It seems nicer to just have a single list and check in `cleanup_unix_sockets` if the socket is a unix socket before trying to close it.

JelteF (commented):
I think the goal of this feature makes sense, but there's some implementation stuff that I don't agree with. See rest of the comments.

AndrewJackson2020:
> I think the goal of this feature makes sense, but there's some implementation stuff that I don't agree with. See rest of the comments.

Thank you for the feedback. Will address these comments in the coming days.

test/test_misc.py
AndrewJackson2020: Went ahead and opened #1306 to change the test() functions to use null queries. Will change the code in this PR to use that instead of the sql() function.

test/test_misc.py
AndrewJackson2020: This comment is now more fully formed.

src/pooler.c
AndrewJackson2020: This was copied/pasted from the original code. I believe in the original code it was to prevent a deletion happening on an abstract unix socket which is not actually a file. Removed if from the `cleanup_tcp_sockets` function,

src/pooler.c
AndrewJackson2020: Renamed the functions to be consistent: `cleanup_tcp_sockets` and `cleanup_unix_sockets`

test/test_misc.py
AndrewJackson2020: These are all changed to use the main test function.

test/test_misc.py
AndrewJackson2020: I modified it so only pgbouncer database can connect to on unix socket after shutdown.

test/test_misc.py
AndrewJackson2020: Issue was that the Linux tests use a different socket directory than BSD and osx. This is fixed now.

src/pooler.c
AndrewJackson2020: Honest answer is that I can't figure out how to consolidate the lists without creating a memory leak which gets flagged in the valgrind tests.

JelteF: I'm pretty sure it's because the `cleanup_tcp_sockets` call (and also the `cleanup_unix_sockets` one), are using `statlist_pop` to iterate over the list. You'd want to use `statlist_for_each_safe` instead and manually remove the TCP entries using `statlist_remove`. Now `cleanup_tcp_sockets` is already removing all unix sockets from the list.

AndrewJackson2020: Thanks for the pointer, will give this a shot.

AndrewJackson2020: Alright I got this working with `statlist_for_each_safe`.

test/test_misc.py
AndrewJackson2020: Realized while I was adding docs that this PR modified the behavior of both `WAIT_FOR_CLIENTS` and `WAIT_FOR_SERVERS`. Added a sentence the the docs for both of those and added a test for `WAIT_FOR_SERVERS`.

src/pooler.c
```diff
@@ -547,7 +568,7 @@ void pooler_setup(void)
 
 		if (!init_done) {
 			/* remove socket on shutdown */
-			atexit(cleanup_sockets);
+			atexit(cleanup_tcp_sockets);
```
JelteF, line 571: I think this should probably continue to clean up both tcp and unix sockets. Although I don't fully understand what the purpose of this cleanup is, afaik once the process exits all sockets will be automatically closed anyway so this seems rather useless. Maybe we should simply remove this?

AndrewJackson2020: I added the atexit call or now.

Personally I'm a bit hesitant to remove 17 year old code that I don't totally understand. Agreed that I'm not sure why sockets would remain open or socket files wouldn't get auto deleted after the process exits.

FWIW the linux tests all pass with this aexit calls removed.

PR #1306, opened 2025-04-18
Change to null query

AndrewJackson2020:
This PR removes the `select 1` in the test functions in favor of a null query. This change is slightly, though admittedly in a very slight way, more performant than `select 1` but also has the added benefit of working in contexts where `select 1` does not work such as in the pgbouncer admin console.


AndrewJackson2020:
> For some reason tests are failing for this branch.

I think this is fixed now.

PR #1307, opened 2025-04-19
Listen port list

AndrewJackson2020:
One common use case of pgbouncer, in my experience is a proxy. There are many times where there are many applications with hardcoded non standard (5432 or 6432) port numbers and admins are forced to set up multiple pgbouncers and slowly find the offending apps and move people over. This is problematic in that it complicates the infrastructure, wastes resources, makes enforcing connection limits harder, etc.

This PR provides an alternative where pgbouncer can listen to multiple ports.

One known outstanding issue with this PR: currently the port number displayed on the admin console is not correct, it just chooses the first port listed in the config.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -724,7 +724,8 @@ static void socket_row(PktBuf *buf, PgSocket *sk, const char *state, bool debug)
 			     sk->login_user_credentials ? sk->login_user_credentials->name : "(nouser)",
 			     sk->pool && !sk->pool->db->peer_id ? sk->pool->db->name : "(nodb)",
 			     replication,
-			     state, r_addr, pga_port(&sk->remote_addr),
+			     (!sk->link && strcmp(state, "active") == 0) ? "idle" : state,
+			     r_addr, pga_port(&sk->remote_addr),
 			     l_addr, pga_port(&sk->local_addr),
 			     sk->connect_time,
 			     sk->request_time,
@@ -1137,6 +1138,7 @@ static bool admin_show_config(PgSocket *admin, const char *arg)
 /* Command: RELOAD */
 static bool admin_cmd_reload(PgSocket *admin, const char *arg)
 {
+	bool ok = true;
 	if (arg && *arg)
 		return syntax_error(admin);
 
@@ -1144,10 +1146,21 @@ static bool admin_cmd_reload(PgSocket *admin, const char *arg)
 		return admin_error(admin, "admin access needed");
 
 	log_info("RELOAD command issued");
-	load_config();
-	if (!sbuf_tls_setup())
+
+	if (!load_config()) {
+		ok = false;
+		log_error("RELOAD Failed, see logs for more details");
+	}
+
+	if (!sbuf_tls_setup()) {
+		ok = false;
 		log_error("TLS configuration could not be reloaded, keeping old configuration");
-	return admin_ready(admin, "RELOAD");
+	}
+
+	if (ok)
+		return admin_ready(admin, "RELOAD");
+	else
+		return send_pooler_error(admin, true, "F0000", false, "RELOAD failed, see logs for additional details");
 }
 
 /* Command: SHUTDOWN */
--- a/src/admin.c
+++ b/src/admin.c
@@ -1851,7 +1851,7 @@ void admin_setup(void)
 	if (!db)
 		die("no memory for admin database");
 
-	// TODO select actual port?
+	/* choose first port to assign to admin database */
 	strport = strlist_pop(listen_port_list);
 	db->port = atoi(strport);
 	strlist_append(listen_port_list, strport);
--- a/src/admin.c
+++ b/src/admin.c
@@ -1844,15 +1844,18 @@ void admin_setup(void)
 	PgGlobalUser *user;
 	PktBuf *msg;
 	int res;
+	char *strport = NULL;
 
 	/* fake database */
 	db = add_database("pgbouncer");
 	if (!db)
 		die("no memory for admin database");
 
-	// TODO select first port
-	// db->port = cf_listen_port;
-	db->port = 6432;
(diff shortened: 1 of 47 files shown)


src/main.c
```diff
@@ -844,14 +845,19 @@ static bool check_old_process_unix(void)
...
+	char *p;
 
 	if (!cf_unix_socket_dir || !*cf_unix_socket_dir || sd_listen_fds(0) > 0)
 		return false;
 
 	memset(&sa_un, 0, len);
 	sa_un.sun_family = domain;
+
+	p = malloc(sizeof(cf_listen_port));
+	safe_strcpy(p, cf_listen_port, sizeof(p));
```
JelteF, line 857: These `sizeof` calls are wrong. They should be using `strlen`. `sizeof` only works in this manner for static arrays (which these pointers are not).

src/pooler.c
```diff
@@ -63,6 +63,8 @@ void cleanup_sockets(void)
 	struct ListenSocket *ls;
 	struct List *el;
 
+	int port;
+
```
JelteF, line 67: let's move this declartion to the `if` block where it's actually used to limit its scope.

src/server.c
```diff
@@ -651,7 +651,7 @@ static bool handle_connect(PgSocket *server)
 	char buf[PGADDR_BUF + 32];
 	bool is_unix = pga_is_unix(&server->remote_addr);
 
-	fill_local_addr(server, sbuf_socket(&server->sbuf), is_unix);
+	fill_local_addr(server, sbuf_socket(&server->sbuf), is_unix, 0);
```
JelteF, line 654: Why are we passing 0 here? Would be good to add a comment on `fill_local_addr` and `fill_remote_addr` in which cases 0 is expected.

JelteF (changes requested):
This indeed seems quite useful. One future-looking thing would be to allow a database entry in the config to specify the transaction pooling port and the session pooling port. That way a single server connection pool could be used for both transaction and session pooling. This would probably be done best by introducing a new `pool_mode` called something like `mixed`.

AndrewJackson2020:
> This indeed seems quite useful. One future-looking thing would be to allow a database entry in the config to specify the transaction pooling port and the session pooling port. That way a single server connection pool could be used for both transaction and session pooling. This would probably be done best by introducing a new `pool_mode` called something like `mixed`.

It would probably have to be 3 ports: transaction, session, and statement. Currently the way I prefer to separate pool mode access is by user (is user_1_session, user_1_transaction, etc.). I understand the point that this is somewhat wasteful in terms of connection sharing though. The thing I don't like about using ports to separate pool mode access is there is no way to restrict access like we can with user (some classes of users are only allowed to use statement mode, etc). Also port numbers are inherently less meaningful to end users than database/usernames in the absence of service discovery.

AndrewJackson2020:
> > This indeed seems quite useful. One future-looking thing would be to allow a database entry in the config to specify the transaction pooling port and the session pooling port. That way a single server connection pool could be used for both transaction and session pooling. This would probably be done best by introducing a new `pool_mode` called something like `mixed`.
> 
> It would probably have to be 3 ports: transaction, session, and statement. Currently the way I prefer to separate pool mode access is by user (is user_1_session, user_1_transaction, etc.). I understand the point that this is somewhat wasteful in terms of connection sharing though. The thing I don't like about using ports to separate pool mode access is there is no way to restrict access like we can with user (some classes of users are only allowed to use statement mode, etc). Also port numbers are inherently less meaningful to end users than database/usernames in the absence of service discovery.

Another much more forward looking thing: would be very nice if the wire protocol allowed a pool_mode parameter to be sent to the server that pgbouncer could use to decide the pooling policy of the connection.

Some adjacent discussion in this link.

https://www.postgresql.org/message-id/flat/20250310230900.01b58a29%40ardentperf.com

src/main.c
AndrewJackson2020: This is now fixed.

src/admin.c
```diff
@@ -1856,14 +1856,19 @@ void admin_setup(void)
...
-
 	/* fake database */
 	db = add_database("pgbouncer");
 	if (!db)
 		die("no memory for admin database");
 
-	db->port = cf_listen_port;
+	/* choose first port to assign to admin database */
+	p = malloc(sizeof(cf_listen_port));
+	safe_strcpy(p, cf_listen_port, sizeof(p));
```
JelteF, line 1868:
```suggestion
	safe_strcpy(p, cf_listen_port, sizeof(cf_listen_port));
```

src/admin.c
```diff
@@ -1856,14 +1856,19 @@ void admin_setup(void)
...
 	int res;
-
 	/* fake database */
 	db = add_database("pgbouncer");
 	if (!db)
 		die("no memory for admin database");
 
-	db->port = cf_listen_port;
+	/* choose first port to assign to admin database */
+	p = malloc(sizeof(cf_listen_port));
```
JelteF, line 1867:
```suggestion
	p = malloc(strlen(cf_listen_port) + 1);
```

JelteF:
@AndrewJackson2020 do you intend to address the feedback here? And there are a few TODOs in the code. What do you want to do about those?

AndrewJackson2020:
> @AndrewJackson2020 do you intend to address the feedback here? And there are a few TODOs in the code. What do you want to do about those?

@JelteF I would like to address the feedback and complete this PR but I recently started a new job (postgres related) and probably won't be able to for a while. Apologies for the radio silence here.

AndrewJackson2020:
Apologies for the radio silence here. Closing this for now because this feature is kind of a pain to test in the presence of takeover mode and I don't have as much of an immediate need for this at my current job. I may reopen a more polished version of this in the future but don't want to clutter pgbouncer's PR list up unnecessarily.

PR #1308, opened 2025-04-22
fix admin_error message for admin_cmd_enable

GuanqunYang193:


AndrewJackson2020:
I'm a bit torn on this change. 

The admin database is always enabled and cannot be disabled so the message "cannot enable Admin db" seems a bit weird to me in that context. On the other hand it also seems weird that the error message would be the same for both admin_command_enable and admin_command_disable.

JelteF (approved):

Issue #1283, opened 2025-03-04
Unstable tests on slow VMs


Update56:
When running pgbouncer tests on slow VMs, there is a chance of test failures.
To find unstable tests, slowed-down virtual machines based on KVM were created.

VM 1:
OS: Alma 9 x64
CPU: 2 cores
RAM: 4 Gb
To emulate the slowdown, the following tags were set in the xml configuration file of the virtual machine before it was booted up:
```
<memory unit='GB'>4</memory>
<vcpu>2</vcpu>
<cputune>
<period>1000000</period>
<quota>40000</quota>
</cputune>
```

VM 2:
OS: Ubuntu 24.04 Arm64(aarch64)
CPU: 2 cores
RAM: 4 Gb
No quotas were set for this VM.

---
The tests were detected as follows: make check was run in a loop with the command `$ for i in $(seq 100000) ; do date; echo "-----Iteration #"$i"-----"; make check || break; done` , after three failures the test was marked xfail.
The following tests were identified, grouped by possible causes (this list is not final):

Asserts in these tests are related to incorrect connection handling
- test_admin.py:115: in test_server_id [log](https://github.com/user-attachments/files/19065357/test_admin.py.115.in.test_server_id.log)
- test_admin.py:74: in test_socket_id [log](https://github.com/user-attachments/files/19065347/test_admin.py.74.in.test_socket_id.log)
- test_ssl.py:405: in test_servers_disconnect_when_changing_tls_config [log](https://github.com/user-attachments/files/19065350/test_ssl.py.405.in.test_servers_disconnect_when_changing_tls_config.log)
- test_limits.py:205: in test_pool_size [log](https://github.com/user-attachments/files/19065349/test_limits.py.205.in.test_pool_size.log)

Asserts in these tests are related to incorrect connection handling with subsequent  errors in assert processing, based on the analysis of the output in the message log
- test_auth.py:53: in test_message [log](https://github.com/user-attachments/files/19065353/test_operations.py.87.in.test_database_restart.log)
- test_auth.py:698: in test_change_server_password_reconnect [log](https://github.com/user-attachments/files/19065362/test_auth.py.698.in.test_change_server_password_reconnect.log)
- test_auth.py:739: in test_change_server_password_server_lifetime [log](https://github.com/user-attachments/files/19065358/test_auth.py.739.in.test_change_server_password_server_lifetime.log)
[...]

PR #1318, opened 2025-05-27
Test 'test_server_id' is updated to support the slow machines

dmitry-lipetsk:
When this test works on the slow machines or we run this test under debugger, PgBouncer may creates TWO connections instead ONE and this test is starting to fail.

This needs to be taken into account.

Moreover, we will make timeout explicitly now and detect the situation when PgBouncer does not create two connection.

It is an indicator of the slow machine, too.

---
When we detect a slow machine we will send warning message.

---
See #1283.

### DIFF
--- a/test/test_admin.py
+++ b/test/test_admin.py
@@ -1,3 +1,4 @@
+import logging
 import threading
 import time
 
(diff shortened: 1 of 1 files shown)


AndrewJackson2020:
Thank you for looking into these tests. I wonder if I may have gone overboard with this and maybe the solution is just to check that the id incremented, does not repeat, etc and not try to predict the exact id.

dmitry-lipetsk:
> Thank you for looking into these tests. I wonder if I may have gone overboard with this and maybe the solution is just to check that the id incremented, does not repeat, etc and not try to predict the exact id.

I think, the such tests must be marked as xfailed or they must use weak checks (warnings, not errors).

dmitry-lipetsk:
Log file with trace of these tests - [pytest-20250529_122130--pid615795.log](https://github.com/user-attachments/files/20502004/pytest-20250529_122130--pid615795.log)

```
    INFO [2025-05-29 12:22:21] --------------------------- [TOTAL INFORMATION]
    INFO [2025-05-29 12:22:21] 
    INFO [2025-05-29 12:22:21]  TOTAL ERROR COUNT  : 0
    INFO [2025-05-29 12:22:21]  TOTAL WARNING COUNT: 4
```

JelteF:
> I wonder if I may have gone overboard with this and maybe the solution is just to check that the id incremented, does not repeat, etc and not try to predict the exact id.

That sounds like the correct solution to this flakiness indeed. Let's just start testing that the ids are unique and incrementing.

PR #1320, opened 2025-05-27
test_ssl::test_servers_disconnect_when_changing_tls_config is corrected to support the slow machines

dmitry-lipetsk:
On the slow machines this test fails because pgbouncer does not able to release a connection in time.

We will do 10 attempts to obtain an empty connection pool.

See #1283.

### DIFF
--- a/test/test_ssl.py
+++ b/test/test_ssl.py
@@ -1,3 +1,4 @@
+import logging
 import subprocess
 import time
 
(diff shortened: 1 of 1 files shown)


test/test_ssl.py
```diff
@@ -396,15 +397,44 @@ def test_servers_disconnect_when_changing_tls_config(bouncer, pg, cert_dir):
...
+                    raise Exception("Connection is not closed in time.")
+                if nAttempt > 1:
+                    time.sleep(0.5)
+
+                cc2 = pg.connection_count(dbname="p0")
+                assert cc2 in [0, 1]
+
+                if cc2 == 0:
+                    if nAttempt == 1:
+                        logging.warning("Connection was closed at the first iteration.")
```
JelteF, line 431: General feedback for all these PRs. I think these warnings are not worth the additional code.

test/test_ssl.py
```diff
@@ -396,15 +397,44 @@ def test_servers_disconnect_when_changing_tls_config(bouncer, pg, cert_dir):
     bouncer.admin("RELOAD")
 
     with bouncer.cur() as cur:
-        assert pg.connection_count(dbname="p0") == 1
+        cc1 = pg.connection_count(dbname="p0")
+        # 1 - OK
+        # 0 - Connection is released
+        assert cc1 in [0, 1]
```
JelteF, line 403: Is this change actually needed? This should not be a valid start state.

JelteF:
Thanks a lot for improving these flaky tests. The additional code used is quite a lot though. I'd rather keep the test code fairly minimal to make it easier to understand what they are doing. So I created #1322 which adds some helper to reduce the boilerplate code necessary to fix all these tests. Could you update your other PRs based on that?

PR #1322, opened 2025-05-27
Handle test flakiness in test_servers_disconnect_when_changing_tls_config

JelteF:
The problem in the test that's solved here as an example was that half a
second wasn't always enough for PgBouncer to close the connection as
expected. This adds a wait_until method, to allow for easy polling until
the required condition is true.

This does a similar fix to #1320, but by introducing this new method
similar test failures can be fixed with a lot less code.

Reported in #1283
Closes #1320

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -132,6 +132,24 @@ def capture(command, *args, stdout=subprocess.PIPE, encoding="utf-8", **kwargs):
     return run(command, *args, stdout=stdout, encoding=encoding, **kwargs).stdout
 
 
+def wait_until(error_message="Did not complete", timeout=5, interval=0.1):
+    """
+    Loop until the timeout is reached. If the timeout is reached, raise an
+    exception with the given error message.
+    """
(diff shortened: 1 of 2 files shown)


test/utils.py
```diff
@@ -132,6 +132,24 @@ def capture(command, *args, stdout=subprocess.PIPE, encoding="utf-8", **kwargs):
...
 
+def wait_until(error_message="Did not complete", timeout=5, interval=0.1):
+    """
+    Loop until the timeout is reached. If the timeout is reached, raise an
+    exception with the given error message.
+    """
+    start = time.time()
+    end = start + timeout
+    last_printed_progress = start
+    while time.time() < end:
```
dmitry-lipetsk, line 143: Will it work correctly when localtime switches between summer and winter time and vice versa? )

dmitry-lipetsk:
> Can you please make a PR? It's much easier to discuss there.
#1324

PR #1337, opened 2025-06-10
Whitespace checks and fixes

petere:
I noticed some files weren't following the .gitattributes whitespace rules, so I fixed those and also added an automatic check for it. (One possible problem is that this check only works in a git checkout. Do we need to support other cases?)

Also, some of that would have been caught by the uncrustify check, but that wasn't run on some files, so I added that as well.


JelteF:
Looks good and nice catch on the missing uncrustify rule

PR #1338, opened 2025-06-12
Cache adhoc scram secrets

### LINKED ISSUE #1335: Cache scram-256 state for a password
There have been multiple reports of scram-256 being much slower than md5 auth. Looking at the code I think this might be fixed by caching (part of) the ScramState for a password that is stored in PgBouncer `auth_file`.

I think we could fairly easily cache the result of `build_adhoc_scram_secret` and `parse_scram_secret` in the `UserCredential` object. That way we don't have to do those operations over and over again for each connection. Especially `build_adhoc_scram_secret` seems rather expensive.  I've done no profiling or benchmarking myself, so I'm not sure about this.

mattheusv:
This PR aims to cache adhoc scram secrets generated from build_adhoc_scram_secret() for cached users. We can have performance improvements by avoid recalculating the scram secrets every time that a cached user open a new connection.

Fixes #1335 

----
This PR is in early draft state, I've noted the following issues using the current approach:

- The PgCredentials struct already have the fields scram_ClientKey[32], scram_ServerKey[32], and  has_scram_keys which is currently used for [SCRAM pass-through ](https://github.com/pgbouncer/pgbouncer/commit/ba1abfe4d94c155a8db22d0b40d513b047c9bf06), so I don't think that we should just set the keys and has_scram_keys to true because we may trigger some code parts that works just for scram pass-through? Maybe another option would be to just include the adhoc_scram_secrets_cached and salt fields and reuse the scram_ClientKey and scram_ServerKey?
- We need to cache the scram_state->salt on PgCrendentials, but since it is a char * and I'm caching using user->salt = strdup(scram_state->salt), the VALGRIND is reporting a memory leak. I think that this is happening because we use the slab API to release the PgCrendential cache entry which just free the pointer and not the chield pointer (in this case the char *salt). Is this correct?

----

I've benchmark the changes using pgbench:
```
==========================> Master

[I] ~/d/pgbouncer  (cache-scram-256-state) ❯❯❯ PGPASSWORD=admin pgbench -h localhost -p 6432 -d postgres -c 1 -j 1 -T 60 -n --connect
pgbench (17.4 (Homebrew), server 18beta1)
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 60 s
number of transactions actually processed: 9347
number of failed transactions: 0 (0.000%)
latency average = 6.420 ms
average connection time = 5.138 ms
tps = 155.763798 (including reconnection times)

[N] ~/d/pgbouncer  ((ed7ecfb9)) ❯❯❯ PGPASSWORD=admin pgbench -h localhost -p 6432 -d postgres -c 2 -j 2 -T 60 -n --connect
pgbench (17.4 (Homebrew), server 18beta1)
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1
query mode: simple
number of clients: 2
number of threads: 2
maximum number of tries: 1
duration: 60 s
number of transactions actually processed: 13806
number of failed transactions: 0 (0.000%)
latency average = 8.693 ms
average connection time = 6.161 ms
tps = 230.070256 (including reconnection times)

==========================> Patch

[I] ~/d/pgbouncer  (cache-scram-256-state) ❯❯❯ PGPASSWORD=admin pgbench -h localhost -p 6432 -d postgres -c 1 -j 1 -T 60 -n --connect
pgbench (17.4 (Homebrew), server 18beta1)
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1
query mode: simple
number of clients: 1
number of threads: 1
maximum number of tries: 1
duration: 60 s
number of transactions actually processed: 12798
number of failed transactions: 0 (0.000%)
latency average = 4.689 ms
average connection time = 3.317 ms
tps = 213.273682 (including reconnection times)


[I] ~/d/pgbouncer  ((ed7ecfb9)) ❯❯❯ PGPASSWORD=admin pgbench -h localhost -p 6432 -d postgres -c 2 -j 2 -T 60 -n --connect
pgbench (17.4 (Homebrew), server 18beta1)
transaction type: <builtin: TPC-B (sort of)>
scaling factor: 1
query mode: simple
number of clients: 2
number of threads: 2
maximum number of tries: 1
duration: 60 s
number of transactions actually processed: 24770
number of failed transactions: 0 (0.000%)
latency average = 4.845 ms
average connection time = 3.490 ms
tps = 412.782169 (including reconnection times)


```

### DIFF
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -20,6 +20,7 @@
  * core structures
  */
 
+#include "common/scram-common.h"
 #include "system.h"
 
 #include <usual/cfparser.h>
@@ -516,6 +517,12 @@ struct PgCredentials {
 	 * settings and connection count tracking.
 	 */
 	PgGlobalUser *global_user;
+
+	int iterations;
+	char *salt;	/* base64-encoded */
+	uint8_t StoredKey[SCRAM_KEY_LEN];
+	uint8_t ServerKey[SCRAM_KEY_LEN];
+	bool adhoc_scram_secrets_cached;
 };
 
 /*
--- a/src/scram.c
+++ b/src/scram.c
@@ -841,7 +841,7 @@ failed:
 	return false;
 }
 
-char *build_server_first_message(ScramState *scram_state, const char *username, const char *stored_secret)
+char *build_server_first_message(ScramState *scram_state, PgCredentials *user, const char *stored_secret)
 {
 	uint8_t raw_nonce[SCRAM_RAW_NONCE_LEN + 1];
 	int encoded_len;
@@ -849,7 +849,7 @@ char *build_server_first_message(ScramState *scram_state, const char *username, 
 	char *result;
 
 	if (!stored_secret) {
-		if (!build_mock_scram_secret(username, scram_state))
+		if (!build_mock_scram_secret(user->name, scram_state))
 			goto failed;
 	} else {
 		switch (get_password_type(stored_secret)) {
@@ -862,8 +862,21 @@ char *build_server_first_message(ScramState *scram_state, const char *username, 
 				goto failed;
 			break;
 		case PASSWORD_TYPE_PLAINTEXT:
-			if (!build_adhoc_scram_secret(stored_secret, scram_state))
-				goto failed;
+			if (user->adhoc_scram_secrets_cached) {
+				scram_state->iterations = user->iterations;
+				scram_state->salt = strdup(user->salt);
+				memcpy(scram_state->StoredKey, user->StoredKey, sizeof(user->StoredKey));
+				memcpy(scram_state->ServerKey, user->ServerKey, sizeof(user->ServerKey));
+			} else {
+				if (!build_adhoc_scram_secret(stored_secret, scram_state))
+					goto failed;
+
+				user->iterations = scram_state->iterations;
+				user->salt = strdup(scram_state->salt);
+				memcpy(user->StoredKey, scram_state->StoredKey, sizeof(scram_state->StoredKey));
+				memcpy(user->ServerKey, scram_state->ServerKey, sizeof(scram_state->ServerKey));
+				user->adhoc_scram_secrets_cached = true;
+			}
 			break;
 		default:
 			/* shouldn't get here */
--- a/include/scram.h
+++ b/include/scram.h
@@ -65,7 +65,7 @@ bool read_client_final_message(PgSocket *client, const uint8_t *raw_input, char 
 			       char **proof_p);
 
 char *build_server_first_message(ScramState *scram_state,
-				 const char *username, const char *stored_secret);
+				 PgCredentials *user, const char *stored_secret);
 
 char *build_server_final_message(ScramState *scram_state);
 
--- a/src/client.c
+++ b/src/client.c
@@ -1127,7 +1127,7 @@ static bool scram_client_first(PgSocket *client, uint32_t datalen, const uint8_t
 		}
 	}
 
-	if (!build_server_first_message(&client->scram_state, user->name, user->mock_auth ? NULL : user->passwd))
+	if (!build_server_first_message(&client->scram_state, user, user->mock_auth ? NULL : user->passwd))
 		goto failed;
 	slog_debug(client, "SCRAM server-first-message = \"%s\"", client->scram_state.server_first_message);
 


mattheusv:
Closing since https://github.com/pgbouncer/pgbouncer/pull/1339 was already merged

JelteF:
Re-opening, since we almost certainly want this change too.

AndrewJackson2020:
@mattheusv Would love to run benchmarks on this PR when you are finished. Please let us know when this is in a good state.

mattheusv:
> @mattheusv Would love to run benchmarks on this PR when you are finished. Please let us know when this is in a good state.

It is working, I just need to fix a memory leak. I think that could be good to start benchmarking to see if the approach that I've used here is worth. What do you think?

include/bouncer.h
```diff
@@ -516,6 +517,12 @@ struct PgCredentials {
 	 * settings and connection count tracking.
 	 */
 	PgGlobalUser *global_user;
+
+	int iterations;
+	char *salt;	/* base64-encoded */
+	uint8_t StoredKey[SCRAM_KEY_LEN];
+	uint8_t ServerKey[SCRAM_KEY_LEN];
```
JelteF, line 524: How is this different from the `scram_ServerKey` field in this struct? Do we need both. If so We need some comments about how they are different.

src/scram.c
```diff
@@ -862,8 +862,21 @@ char *build_server_first_message(ScramState *scram_state, const char *username,
 				goto failed;
 			break;
 		case PASSWORD_TYPE_PLAINTEXT:
-			if (!build_adhoc_scram_secret(stored_secret, scram_state))
-				goto failed;
+			if (user->adhoc_scram_secrets_cached) {
+				scram_state->iterations = user->iterations;
+				scram_state->salt = strdup(user->salt);
+				memcpy(scram_state->StoredKey, user->StoredKey, sizeof(user->StoredKey));
+				memcpy(scram_state->ServerKey, user->ServerKey, sizeof(user->ServerKey));
```
JelteF, line 869: I think we should also do this caching for the `PASSWORD_TYPE_SCRAM_SHA_256` case. So let's move this before the case. So in pseudo code:

```
if cached
    scram_state = cached_scram_state
else
    case PASSWORD_TYPE_SCRAM_SHA_256
	scram_state = parse_scram_secret()
    case PASSWORD_TYPE_PLAINTEXT
	scram_state = build_adhoc_scram_secret()
    cached_scram_state = scram_state
```

src/scram.c
```diff
@@ -862,8 +862,21 @@ char *build_server_first_message(ScramState *scram_state, const char *username,
 				goto failed;
 			break;
 		case PASSWORD_TYPE_PLAINTEXT:
-			if (!build_adhoc_scram_secret(stored_secret, scram_state))
-				goto failed;
+			if (user->adhoc_scram_secrets_cached) {
```
JelteF, line 865: We should have some way to invalidate the cache when a new password is assigned to the user.

mattheusv: Its a good point. Any thoughts on this? Maybe a TTL? Do we have any cache invalidation on PgBouncer code already?

JelteF: There is `unquote_add_authfile_user`, which is called for each user in the reloaded password file. That seems like a reasonable place to do this.

include/bouncer.h
mattheusv: I've changed to reuse the scram_ServerKey and scram_ClientKey and changed the has_scram_keys to use_scram_keys for the scram pass-through feature. The adhoc_scram_secrets_cached controls if scram_ServerKey,  scram_StoredKey and scram_SaltKey should be re-used when building the scram message exchange.

src/scram.c
mattheusv: Fixed

mattheusv:
@JelteF  @AndrewJackson2020 I think that the PR is in a good shape now, the memory leaks are fixed and I've also fixed the initial comments. Can we follow up with the review and the benchmark?

AndrewJackson2020:
Based on my performance testing this seems to have totally fixed any performance issues related to scram.

```console
postgres md5
tps = 1045.835848 (including reconnection times)
postgres scram
tps = 252.256020 (including reconnection times)
pgbouncer md5
tps = 8067.671630 (including reconnection times)
pgbouncer scram encoded
tps = 338.854918 (including reconnection times)
pgbouncer scram plain text
tps = 339.816927 (including reconnection times)
```

JelteF:
> Based on my performance testing this seems to have totally fixed any performance issues related to scram.

That looks very nice (i.e. both taking the same amount of time). Still it would be great to improve general perf more though. 8000 vs 330 is quite a big difference.

src/scram.c
```diff
@@ -841,33 +841,48 @@ static bool build_mock_scram_secret(const char *username, ScramState *scram_stat
...
-		case PASSWORD_TYPE_PLAINTEXT:
-			if (!build_adhoc_scram_secret(stored_secret, scram_state))
-				goto failed;
-			break;
-		default:
-			/* shouldn't get here */
-			goto failed;
+			}
+
+			if (!user->dynamic_passwd) {
```
JelteF, line 879: I think there's probably some possibility of even caching this for dynamic_passwd users (aka `auth_user`), but that can be done in a follow up PR.

AndrewJackson2020:
> > Based on my performance testing this seems to have totally fixed any performance issues related to scram.
> 
> That looks very nice (i.e. both taking the same amount of time). Still it would be great to improve general perf more though. 8000 vs 330 is quite a big difference.

This benchmark is using pgbench with 2 jobs/clients each executing ";" repeatedly. I think that the 8000 vs 330 is just due to decreased performance on the client side.

You see a similar decrease in performance when querying postgres directly as seen in the first 2 benchmarks.

JelteF (commented):
Do we have tests that excersise this caching, if not we should modify some of the existing tests to re-authenticate a few times.

JelteF:
> You see a similar decrease in performance when querying postgres directly as seen in the first 2 benchmarks.

Postgres "only" has a 4 times perf difference, here it's 24 times. I think the most important here is how many connections a single PgBouncer can handle. While running this test is the PgBouncer process maxing out its single core? If not, how many tps can it do until it maxes out?

JelteF:
Rephrasing my question above: How much CPU does the pgbouncer process consume at 300 tps for MD5 vs SCRAM? I think that's what we should care about, trying to get as close to eachother as possible.

AndrewJackson2020:
> > You see a similar decrease in performance when querying postgres directly as seen in the first 2 benchmarks.
> 
> Postgres "only" has a 4 times perf difference, here it's 24 times. I think the most important here is how many connections a single PgBouncer can handle. While running this test is the PgBouncer process maxing out its single core? If not, how many tps can it do until it maxes out?

Let me try to get u those numbers tommorow.

Neustradamus:
@mattheusv: Thanks, good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1
- https://github.com/pgbouncer/pgbouncer/issues/1335
- https://github.com/pgbouncer/pgbouncer/pull/1338

PR #1339, opened 2025-06-13
Make SCRAM Iterations Runtime Configurable

### LINKED ISSUE #1335: Cache scram-256 state for a password

AndrewJackson2020:
This PR makes the number of scram iterations that are used to generate SCRAM keys configurable. This is an attempt to partially resolve #1335 by allowing people to use a lower iteration count to reduce the amount of computation that is needed for PgBouncer to process a client authentication request.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -426,6 +426,11 @@ the error go away.
 
 Default: 200
 
+### scram_iterations
+
+The number of computational iterations to be performed when encrypting a password using SCRAM-SHA-256. A higher number of iterations provides additional protection against brute-force attacks on stored passwords, but makes authentication slower.
+
+Default: 4096
 
 ## Authentication settings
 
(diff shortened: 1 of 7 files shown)


Neustradamus:
@AndrewJackson2020: Thanks, good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1
- https://github.com/pgbouncer/pgbouncer/issues/1335
- https://github.com/pgbouncer/pgbouncer/pull/1338
- https://github.com/pgbouncer/pgbouncer/pull/1339

PR #1347, opened 2025-07-05
ci: Re-enable commented out Debian/Ubuntu tasks, with manual trigger

petere:
Reactivate the Debian/Ubuntu tasks that were commented out in 92b4645, setting them to trigger manually.  This was already done for the other tasks in commit 0738448.  There I wrote, "this does not appear to work for matrix elements inside a task", but this appears to work now (or it worked back then but I didn't do it correctly).

In the re-enabled tasks, raise the Ubuntu version from 20.04 to 22.04, like the change in commit ef611b5, and where it was already 22.04 to now 24.04.

The clang variant is removed.  This is already covered by the FreeBSD task now.

The variant that runs Valgrind against PostgreSQL 9.6 is removed. This was originally meant to handle the non-SCRAM code paths, but those PostgreSQL versions are long obsolete.

Also update the PGVERSIONs of the Debian tasks.  Those were supposed to be the versions that ship with upstream Debian, but in the meantime, the actual "stable" and "oldstable" releases have shifted, so this no longer matched.  To avoid these kinds of silent changes, write out the actual release names instead of "stable" and "oldstable".

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -6,8 +6,9 @@ env:
 task:
   name: Linux (Debian/Ubuntu)
   matrix:
-    # - container:
-    #     image: ubuntu:20.04
+    - container:
+        image: ubuntu:22.04
+      trigger_type: manual
     - container:
         image: ubuntu:22.04
       env:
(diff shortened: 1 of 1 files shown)


petere:
Note:
* The `--disable-evdns` task will currently fail to build because of #1348.
* The ubuntu:24.04 task will currently fail in the tests, reasons unknown, could be a psycopg issue.

JelteF (approved):

PR #1352, opened 2025-07-08
Add support to set ciphers for TLSv1.3

mattheusv:
For TLS v1.3 we need to use a different API to set the ciphers on OpenSSL. This adds new config options to configure these ciphers. The new options are called `client_tls13_ciphers` and `server_tls13_ciphers`. These names aligns well with Postgres its new `ssl_tls13_ciphers` option, which was introduced in PG18.  

This PR was previously open on libusual: https://github.com/libusual/libusual/pull/72

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -243,7 +243,7 @@ static bool admin_set(PgSocket *admin, const char *key, const char *val)
 			if (!buf) {
 				return admin_error(admin, "no mem");
 			}
-			if (strstr(key, "_tls_") != NULL) {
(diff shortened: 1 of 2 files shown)


lib/test/test_tls.c
```diff
@@ -833,6 +835,46 @@ static void test_set_mem(void *z)
 end:    ;
 }
 
+static void test_cipher_tlsv12(void *z)
```
JelteF, line 838: I think it would make sense to also have a Python based test for this functionality. To make sure the pgbouncer integration with this new API works as expected.

JelteF (approved):

PR #1356, opened 2025-07-14
Update SCRAM code to PG18

JelteF:
Since SCRAM support as copied from Postgres 11.5, the SCRAM infrastructure in Postgres has changed quite a bit. This update uses Postgres 18beta3 as base. This was based on the initial work done in #797, but that was then heavily changed to use the new PG APIs everywhere for SCRAM (allowing removal of a bunch of crypto code from libusual).

Obsoletes #797


petere:
The amount of new files this imports seems mad. Why do we need to import a new MD5 implementation as part of this? We should try harder to untangle this and grab only what we need.

eulerto:
@JelteF I'm checking it now.

eulerto:
_This is not a complete review as I'm still reviewing the code..._

Could you reference the exact Postgres commit that this update is based on? You mentioned 18beta3 but it would be better if we have a commit hash for future reference.

I don't know if you are planning to squash the commits into one, if so the second patch has a misleading message (code does not compile as you said in the last paragraph). I prefer that every commit compiles but for the sake of history, the mechanical copy of SCRAM code can be a separate commit. If we refer in the commit message that it doesn't compile but the next commit will use this preparatory work to update the SCRAM code. That's fine.

I would expect that you remove unused header files from the 2nd patch like

```
+//#ifndef FRONTEND
+//#include "postgres.h"
+//#else
+//#include "postgres_fe.h"
+//#endif
```

Maybe you should say in the commit message that these inclusions should be removed for future reference.

Could you remove the unused libusual code in a separate commit? If you're squashing all commits, ignore it.

Is it a requirement?

```
-  flags="$flags -Wno-unused-parameter -Wno-missing-field-initializers"
+  flags="$flags -Wno-unused-parameter -Wno-sign-compare -Wno-missing-field-initializers"
```

It is just a style preference but variable i could be declared inside the for loop in `calculate_client_proof()`.

I tested with different Postgres version and the tests pass from 10 to 18. However, there are some failures for 9.6. The version that introduces SCRAM is Postgres 10. I was curious if the tests were prepared to handle older versions. It seems the answer is no. I will create a separate patch to fix it.

JelteF:
> _This is not a complete review as I'm still reviewing the code..._
> 
> Could you reference the exact Postgres commit that this update is based on? You mentioned 18beta3 but it would be better if we have a commit hash for future reference.

Done, I had used REL_18_BETA3.

> I don't know if you are planning to squash the commits into one

I plan to keep them as separate commits to make it clear in the commit history where code comes from.

> I would expect that you remove unused header files from the 2nd patch like
> 
> ```
> +//#ifndef FRONTEND
> +//#include "postgres.h"
> +//#else
> +//#include "postgres_fe.h"
> +//#endif
> ```

Done, I also removed unneeded `system.h` includes. These things came from your original #797 PR, but I agree it's good to clean up.

> Could you remove the unused libusual code in a separate commit? If you're squashing all commits, ignore it.

Done

> Is it a requirement?
> 
> ```
> -  flags="$flags -Wno-unused-parameter -Wno-missing-field-initializers"
> +  flags="$flags -Wno-unused-parameter -Wno-sign-compare -Wno-missing-field-initializers"
> ```

Yes, some of the Postgres code throws warnings otherwise. It's a rather useless warning anyway.

> It is just a style preference but variable i could be declared inside the for loop in `calculate_client_proof()`.

Agreed, but I copied this from the postgres sources. So I'll leave it like that. Feel free to send a patch to PG to fix it there.

> I tested with different Postgres version and the tests pass from 10 to 18. 

Good to hear!

include/bouncer.h
```diff
@@ -32,6 +32,10 @@
 #include <event2/event.h>
 #include <event2/event_struct.h>
 
+#include "common/postgres_compat.h"
+#include "common/cryptohash.h"
+#include "common/scram-common.h"
```
petere, line 37: This is importing a lot of very specific stuff into a very general header file. We should find a better way.

The header file `postgres_compat.h` is supposed to make PostgreSQL code compilable in PgBouncer, so I don't see what that is doing here.

The other includes can probably be avoided with some judicious use of forward declarations.

petere (changes requested):
As a general matter, I'm not sure we should be importing things from unreleased PostgreSQL versions. (I mean, we can import them, but then we shouldn't release them before upstream has released.) In the past, we imported from stable PostgreSQL releases.

include/bouncer.h
JelteF: Addresesed as best as I could `cryptohash.h` is needed for its enum definition, which I'd rather not forward declare by fully copying it. `scram-common.h` I left out out by hardcoding 32 in all places in this header. `postgres_compat.h` was needed for `cryptohash.h`, but simply defining `uint8` was enough to make this pass.

JelteF:
> As a general matter, I'm not sure we should be importing things from unreleased PostgreSQL versions. 

I definitely agree with this in principle but in this case the changes were so minimal between 17.6 and 18beta3 that it seemed better to me to start using the newest code
[scram-files-17.6-vs-18beta3.diff.gz](https://github.com/user-attachments/files/21977366/scram-files-17.6-vs-18beta3.diff.gz)

petere:
I'm getting compiler warnings from this patch:

```
     CC       src/common/hmac.c
src/common/hmac.c: In function 'pg_hmac_init':
src/common/hmac.c:144:17: warning: comparison of integer expressions of different signedness: 'size_t' {aka 'long unsigned int'} and 'int' [-Wsign-compare]
  144 |         if (len > block_size)
      |                 ^
src/common/hmac.c:181:23: warning: comparison of integer expressions of different signedness: 'int' and 'size_t' {aka 'long unsigned int'} [-Wsign-compare]
  181 |         for (i = 0; i < len; i++)
      |                       ^
     CC       src/common/pgstrcasecmp.c
     CC       src/common/saslprep.c
src/common/saslprep.c: In function 'pg_utf8_string_len':
src/common/saslprep.c:1009:25: warning: comparison of integer expressions of different signedness: 'size_t' {aka 'long unsigned int'} and 'int' [-Wsign-compare]
 1009 |                 if (len < l || !pg_utf8_islegal(p, l))
      |                         ^
src/common/saslprep.c: In function 'pg_saslprep':
src/common/saslprep.c:1078:24: warning: comparison of integer expressions of different signedness: 'int' and 'long unsigned int' [-Wsign-compare]
 1078 |         if (input_size >= MaxAllocSize / sizeof(pg_wchar))
      |                        ^~
     CC       src/common/scram-common.c
     CC       src/common/string.c
     CC       src/common/unicode_norm.c
src/common/unicode_norm.c: In function 'recompose_code':
src/common/unicode_norm.c:286:31: warning: comparison of integer expressions of different signedness: 'int' and 'long unsigned int' [-Wsign-compare]
  286 |                 for (i = 0; i < lengthof(UnicodeDecompMain); i++)
      |                               ^
     CC       src/common/wchar.c
src/common/wchar.c: In function 'pg_utf8_verifystr':
src/common/wchar.c:1904:17: warning: comparison of integer expressions of different signedness: 'int' and 'long unsigned int' [-Wsign-compare]
 1904 |         if (len >= STRIDE_LENGTH)
      |                 ^~
src/common/wchar.c:1906:28: warning: comparison of integer expressions of different signedness: 'int' and 'long unsigned int' [-Wsign-compare]
 1906 |                 while (len >= STRIDE_LENGTH)
      |                            ^~
     CC       lib/usual/aatree.c
```

It's strange that I don't see these on CI anywhere.

JelteF:
> Hmm, disagree. That smells like these old CVS headers that made it impossible to diff files.

I'm also not a fan, so I'll leave it like this.

> I'm getting compiler warnings from this patch:

Running this should fix that (that picks up the new warning ignore I added to the libusual m4 file):

```bash
make clean
./autogen.sh
./configure # add whatever flags you normally use
make
```

lib/m4/usual.m4
```diff
@@ -99,7 +99,7 @@ if test x"$GCC" = xyes; then
   good_CFLAGS="$CFLAGS"
   flags="-Wall -Wextra"
   # turn off noise from Wextra
-  flags="$flags -Wno-unused-parameter -Wno-missing-field-initializers"
+  flags="$flags -Wno-unused-parameter -Wno-sign-compare -Wno-missing-field-initializers"
```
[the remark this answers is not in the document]
JelteF:
```diff
 	input_size = pg_utf8_string_len(input);
-	if (input_size < 0)
-		return SASLPREP_INVALID_UTF8;
```

That definitely introduces a bug. `pg_utf8_string_len` returns `-1` on failure and we'd now miss that. Without looking closely at the code, it's hard to say without careful checking if it introduces more unexpected bugs.

So I'd rather keep PG the as is in this regard, to avoid introducing new bugs (if there's bugs at least we have the same ones as PG). But I agree it's unnecessarily broad to ignore the warning also in PgBouncer code. So, I now only ignores the warning in the files in question. instead I now ignore the warning in these specific files.

Issue #1351, opened 2025-07-07
Why does mkauth.py add a third empty "" field in the auth_file format?


guoard:
Hi,
I was reviewing the `mkauth.py` script used to generate `auth_file` entries and noticed it formats each line with [three quoted fields](https://github.com/pgbouncer/pgbouncer/blob/master/etc/mkauth.py#L32), for example:

```python
lines.append('"%s" "%s" ""\n' % (user, psw))
```

So each line looks like:

```
"user" "password" ""
```

However, the **official PgBouncer documentation** states:

> *“The first field is the user name and the second is either a plain-text, a MD5-hashed password, or a SCRAM secret. PgBouncer ignores the rest of the line.”*

And the example `auth_file`s I’ve seen usually contain only **two fields** per line:

```
"user" "password"
```

Could you please clarify:

* What is the purpose of the **third empty field** added by `mkauth.py`?
* Is this third field required for compatibility or other reasons, or can it be safely omitted?
* Is it just a leftover artifact or for some future-proofing?

Understanding this would help avoid confusion when generating or managing `auth_file` contents.

Thanks for your help!

guoard:
@petere Thanks for the detailed explanation, that makes perfect sense. I've opened a pull request to make the change.

PR #1365, opened 2025-08-04
Remove obsolete third field from auth_file entries in mkauth.py

guoard:
The `mkauth.py` script previously generated auth_file entries with three quoted fields: `"user" "password" ""`.
The third field was a leftover from an old postgresql flat-file format (`pg_pwd`) that included a "valid until" field in addition to the username and password.

This format was originally used to allow pgbouncer to directly read the `pg_pwd` file, maintaining compatibility during a transition period.
However, the `pg_pwd` file has long been removed, and pgbouncer ignores all fields after the second one.

Removing this unused field simplifies the file format, aligns with pgbouncer documentation and common usage, and helps prevent confusion when managing auth_file contents.

Fixes: #1351

### DIFF
--- a/etc/mkauth.py
+++ b/etc/mkauth.py
@@ -29,7 +29,7 @@ for user, psw in curs.fetchall():
     if not psw:
         psw = ""
     psw = psw.replace('"', '""')
-    lines.append('"%s" "%s" ""\n' % (user, psw))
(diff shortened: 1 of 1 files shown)


petere:
Could you add some more detail in the commit message about why it's obsolete, so that the commit log becomes a self-contained description of the change.

guoard:
@petere Done.

PR #1372, opened 2025-08-16
Have pytest report skipped tests by default

petere:
What do you think about having pytest report skipped tests by default at the end of the run, like in the attached patch. I have found this useful to analyze exactly what certain CI tasks test and don't test. So we could enable it on CI only. But I imagine it could also be useful for local development, if you don't realize that there are certain tests that you haven't activated.


JelteF:
Yeah looks nice.

PR #1375, opened 2025-08-19
ci: Enable LDAP on more platforms

petere:
LDAP should be tested on more platforms. Let's look at the test runs ...

- Red Hat: build ok, tests fail
- Alpine: build ok, tests ok
- FreeBSD: build ok, tests fail
- macOS: build fails
- Windows: build fails

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -134,7 +134,7 @@ task:
       - image: alpine:latest
   setup_script:
     - apk update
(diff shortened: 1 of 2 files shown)


JelteF:
The ones that fail the tests, all seem to fail at starting the openldap server. But it doesn't really show in the logs why it fails to start. Probably adding a bit more logging, or at least `set -x` to the script would help debugging.

petere:
Yeah, I think we need to either install the right packages or point the script to the right path. I'll look into it.

Issue #1376, opened 2025-08-25
pgbouncer error: FATAL @src/objects.c:1387 in function disconnect_server(): bad server state: 9


glushakov:
Hi!

we've experiencing an issue with pgbouncer.
We have 3 instances with "so_reuseport = 1" option
2 instance fail in same time with same error, and 1 instance still working normally.

`FATAL @src/objects.c:1052 in function disconnect_client(): bad client state: 0`

after this message in the log it stopped working. pgbouncer version is pgbouncer-1.24.1 , PG 14.18

JelteF:
For reference this means it's trying to disconnect a server in the `SV_JUSTFREE` state. We should probably just make `disconnect_server` a no-op when the server is in that state. 

I'm not entirely sure why a server with that state would get called disconnect_server for though. Maybe the server closed the connection at the wrong moment. Did your Postgres server crash by any chance?

eulerto:
It seems `disconnect_server()` was called twice. The SV_JUSTFREE state is only set in this function. It seems we need a protection somewhere. There are 88 calls of this function. If at least we have a reproducible test case or a stack trace in the logs; unfortunately, we don't have any of them.

JelteF:
Really all logs from somewhat recently before this event would be quite useful.

PR #1382, opened 2025-08-27
Improve fatal messages in disconnect_server and disconnect_client

JelteF:
These functions are called from many places so it's hard to figure out what might have caused an issue with an unexpected state. This helps narrowing that down by including the provided reason in the fatal message.

Related to #1376

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1387,7 +1387,7 @@ void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...
 			unlink_server(server, reason);
 		break;
 	default:
-		fatal("bad server state: %d", server->state);
+		fatal("bad server state: %d, %s", server->state, reason);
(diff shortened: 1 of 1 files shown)


JelteF:
> There is one case (see client_proto()) that disconnect_client() is called with reason equals to NULL.

Oops, yeah fixed now.

Regarding the function name in the fatal log messages. That's already happening, see the original bug report (it even includes the line number): https://github.com/pgbouncer/pgbouncer/issues/1376

That's because the `fatal` definition is like this:
```
/** Log and die.  It also logs source location */
#define fatal(...) do { LOG_CONTEXT_DEF; \
			log_fatal(__FILE__, __LINE__, __func__, false, LOG_CONTEXT, __VA_ARGS__); \
			exit(1); } while (0)
```

PR #1386, opened 2025-09-02
Support Multithread Mode

GuanqunYang193:
Some comments and tests are still unfinished. But overall—please welcome the multithreaded version of PgBouncer! 

### Thread Model
PgBouncer now supports a multithreaded execution model. The number of worker threads is controlled by the thread_number in pgbouncer config.
Each worker maintains its own event base, caches, and pools.
The main thread accepts new connections from TCP/Unix sockets and dispatches file descriptors to workers, ensuring each connection is handled by exactly one worker. This allows worker threads to run lock-free, at the cost of higher memory usage.
Cross-thread interactions remain for:
Admin operations (e.g., killing databases/clients)
* Prepared statements
* User credentials
* DNS resolution
* Statistics
To ensure thread-safety:
Lightweight operations use locks to protect data structures.
Heavy operations (mainly admin tasks) are executed by pausing the target worker thread, ensuring exclusive access without fine-grained locks. This approach increases latency but avoids deadlocks and complex locking.
### Configuration Changes
Multithread mode introduces the following configuration changes:
* *_pool_size: now applies per thread instead of globally. Effective memory usage is approximately original value × number of threads.
thread_number: the number of threads spawned in multithreaded pgbouncer. Default as 0 to use the original single-thread pgbouncer.
Other configuration options remain unchanged.
### Multithread-Specific Behavior
Multithreading introduces a few semantic differences compared to single-thread mode:
* Timekeeping: Each worker maintains its own time_cache instead of a global snapshot.
* Admin commands: SHOW output is per-thread rather than aggregated, alongside a column named “thread_id”, to indicate which thread the resource belongs to.
* Global limits: Limits such as max_client_connection and user-based limits are not strictly enforced across all threads. Connections may slightly exceed the configured limit (up to limit + number of threads).
* Eviction: Under multithread mode, pgbouncer will only evict connections under same thread.
### Unsupported Features
The following features are not supported in multithread mode due to cross-thread complexity:
Takeover
All other features should work as in single-thread mode. Any broken behavior will be treated as a bug.
Remaining Execution Items
Add more thorough unit tests for the new multithreaded execution path
Refactor code to reduce repetitive patterns that are common in single-thread and multithreaded paths.

co-worker @Beihao-Zhou


include/bouncer.h
```diff
@@ -941,3 +947,16 @@ bool load_config(void);
...
+extern pthread_key_t thread_id;
+
+
+extern int client_count;
+extern SpinLock client_count_lock;
+
+extern int total_active_count;
+extern SpinLock total_active_count_lock;
+
+#endif /* BOUNCER_H */
```
JelteF, line 962: Please install editorconfig and save this file again to add the final newline.

include/multithread.h
```diff
@@ -0,0 +1,179 @@
...
+
+ struct event ev_sigquit;
+ struct event ev_sigusr1;
+ struct event ev_sigusr2;
+ struct event ev_sighup;
+#endif
+} SignalEvent;
+
+
+typedef struct WorkdersignalEvents{
```
JelteF, line 57:
```suggestion
typedef struct WorkersignalEvents{
```

src/takeover.c
```diff
@@ -36,6 +37,7 @@
 
 static PgSocket *old_bouncer = NULL;
 
+// Not supported in multithread mode
```
JelteF, line 40: FYI I think that's totally fine. This feature is deprecated, and I think we should probably remove it soon.

src/objects.c
```diff
@@ -1333,9 +1623,9 @@ static void unlink_server(PgSocket *server, const char *reason)
  * The latter is for protocol and communication errors where a normal
  * protocol termination is not possible.
  */
-void disconnect_server(PgSocket *server, bool send_term, const char *reason, ...)
+void disconnect_server(PgSocket *server, bool send_term, int thread_id, const char *reason, ...)
```
JelteF, line 1626: One thing I dislike in the PR is this infection of `thread_id` everywhere througout the API. My understanding is that each `PgSocket` is essentially bound to a thread, because its `SBuf` is bound to a thread. It seems much nicer to have the `thread_id` be extracted from this `SBuf` whenever it's needed when a `PgSocket` is already part of the arguments, instead of then separately passing the `thread_id` again. This will also make the PR much easier to review, because it will have much fewer changed lines of code.

If for some reason the thread_id from the `SBuf` cannot be used for this, it still seems much better to put the `thread_id` of the `PgSocket` in the `PgSocket` definition, so we can still remove these arguments.

JelteF: I think we should definitely try to avoid it at the serven/client level functions too. It seems that all this is only/mostly needed for `disconnect_client` and `disconnect_server`. Another approach would be to have a variant of disconnect_server that takes a thread_id for the places where that's actually needed, but for the normal case it would use the thread id of the server (so have two functions `disconnect_server` and `disconnect_server_on_thread`).

AndrewJackson2020:
> As for whether we should release the multithreaded mode as a separate package, I personally feel that doing so would introduce a significant amount of duplicate code and reduce overall code reuse—it would essentially be like maintaining two separate codebases. That's why we chose not to go with the separate-release approach for multithreaded mode.

Okay this makes sense. I was originally envisioning a configure/build time option. Thank you.

Beihao-Zhou:
> One thing that would be very helpful is a rough overview of the changes that you made in the PR description, as well as the design decisions taken to implement this and why you chose to do it this way. e.g. Questions that I'm wondering about:
> 
> * When do you assign a connection to a thread?
> * Does a connection stay assigned to the same thread forever?
> * How do you handle thread safety in general?
> * Are there exceptions where you do something special to manage thread safety?
> * What were some difficult problems you had to fix?

Hi @JelteF, we updated the PR description to answer these questions as much as possible. Let us know if there are any missed pieces!

JelteF:
Thanks a lot for the overview. These are my thoughts:

> * *_pool_size: now applies per thread instead of globally. Effective memory usage is approximately original value × number of threads.
> 
> * max_db_connections, max_user_connections: limits are enforced per thread.

I think these are problematic, what is stopping us from enforcing these three globally?

> * server_round_robin: balancing is done per thread, not globally. This may affect server_login_retry and failover behavior.

I'm not sure if this is problematic or not, I think it's probably fine.

> * Takeover

This is fine not to support

>     * SIGHUP reload
> 
> 
> If SIGHUP reload is critical, it may be implemented in the future by pausing worker threads.

Reloading on SIGHUP is extremely common. We definitely need this feature.

> * Admin commands: SHOW output is per-thread rather than aggregated.

You mean if you do `SHOW CLIENTS` you only get to see a subset of the clients? That seems very confusing, and like something that should be fixed.

> * Global limits: Limits such as max_client_connection and user-based limits are not strictly enforced across all threads. Connections may slightly exceed the configured limit (up to limit + number of threads).

I think this is probably acceptable, but I'm curious why this is the case.

> * Eviction: Under multithread mode, pgbouncer will only evict connections under same thread.

So what happens if eviction is needed, but there is nothing to be evicted in the same thread?

GuanqunYang193:
> Thanks a lot for the overview. These are my thoughts:
> 
> > * *_pool_size: now applies per thread instead of globally. Effective memory usage is approximately original value × number of threads.
> > * max_db_connections, max_user_connections: limits are enforced per thread.
> 
> I think these are problematic, what is stopping us from enforcing these three globally?
> 
> > * server_round_robin: balancing is done per thread, not globally. This may affect server_login_retry and failover behavior.
> 
> I'm not sure if this is problematic or not, I think it's probably fine.
> 
> > * Takeover
> 
> This is fine not to support
> 
> > ```
> > * SIGHUP reload
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If SIGHUP reload is critical, it may be implemented in the future by pausing worker threads.
> 
> Reloading on SIGHUP is extremely common. We definitely need this feature.
> 
> > * Admin commands: SHOW output is per-thread rather than aggregated.
> 
> You mean if you do `SHOW CLIENTS` you only get to see a subset of the clients? That seems very confusing, and like something that should be fixed.
> 
> > * Global limits: Limits such as max_client_connection and user-based limits are not strictly enforced across all threads. Connections may slightly exceed the configured limit (up to limit + number of threads).
> 
> I think this is probably acceptable, but I'm curious why this is the case.
> 
> > * Eviction: Under multithread mode, pgbouncer will only evict connections under same thread.
> 
> So what happens if eviction is needed, but there is nothing to be evicted in the same thread?

Those are good suggestions. Here are some of our concerns:

* Limits. Currently, limits are checked for all connections. Making limits global in multithreaded mode would introduce lock contention, and the little inaccuracy. This is because the point where PgBouncer performs the check and the point where it finishes adding new connections are not the same. Making the entire check-and-create logic a critical section would hurt performance. We therefore chose to sacrifice some accuracy for better performance.
Our initial thought was that local limits could avoid these problems. However, mixing global and local limits would definitely confuse clients. Thus, we can try to make ALL limits global, consistent with existing behavior.

* Eviction. It is possible (though rare) that all connections are created in other threads, and therefore no connections can be evicted in the current thread (this only happens when limits are global). In our current design, PgBouncer will not launch new connections in such a thread. If the client retries a few times, they may reach a thread that does have connections to evict. This case should remain rare unless users set a small limits. Additionally, server timeouts can free unused server connections, as long as server_idle_timeout is configured to a reasonable value.
My guiding principle is to avoid iterating over all pools across threads for each connections in all threads, since that would require acquiring multiple locks and would still fail to find connection to evict (e.g., the “oldest” connection could be closed and new connection in another thread comes in during the process).
So I think it’s fine not to force an eviction if that rare case happens.

* SIGHUP. While it is common to reload SIGHUP, it’s not clear for us whether reloading configs is widely used. We have not invested in supporting config reload via SIGHUP, since the current config-loading logic is not multithread-friendly. Refactoring would be needed if this becomes a requirement. Please let us know your thought on this.

* SHOW CLIENTS/SERVERS. We added thread IDs to the SHOW results so that users can see which threads the clients were created in. We believe this will help with debugging.

Thanks for your suggestions and please let us know your thoughts on these design choices. We are happy to discuss further or adjust if needed.

petere (changes requested):
First, get the CI builds going. The code currently doesn't compile, and the code formatting is a bit all over the place.

Think about how your patch could be tested on CI.

Avoid unrelated whitespace and formatting changes. (The code formatting checks will probably undo them anyway.)

The code has way too many `if (multithread_mode)`. This should all be pushed down into lower APIs. Stats output, logging output, `SHOW` commands, etc. should show the same thing in both modes.

All of this could make the patch much smaller.

GuanqunYang193:
> @GuanqunYang193 thanks for putting together this work to make pgbouncer multithreaded. It seems very promising, I believe it could enable some very exciting use cases that were not possible with a single threaded pgbouncer.
> 
> Regarding the implementation, the approach that has been taken seems to be to have separate workers handle connections. Did you consider a less disruptive approach where we keep most of the logic as a part of the single event loop but we create a pool of worker threads that handle only specific tasks (e.g cpu intensive operations) and then report back to the main event loop by triggering an event ?

Yes, this is an implementation we have considered, but we believe it would introduce several problems. For example:

1. Events from a single connection would be handled by multiple threads. The execution order of events for one connection must be guaranteed, but it is very difficult to refactor the code to ensure that no two events for the same connection are executed at the same time.

2. This approach would also require locking nearly all data structures in PgBouncer, which could lead to deadlock and contention issues.

We did take some time to explore implementation details, such as different interfaces for thread-safe event bases on different platforms. However, based on what we have now, this implementation seems much more challenging and offers no performance gain. There might be some memory savings since pools would be shared, but we do not feel that is a significant concern for our implementation.

JelteF (changes requested):
I skimmed the code another time.

Overall I agree with Peter: We need to really minimize the places where `multithread_mode` is checked by either pushing the check down into a helper function/macro, or by storing globals always in the Thread object. Otherwise maintaining PgBouncer will become much more timeconsuming and error prone, because of the need to always having to update both branches in the code.

There are currently also no tests for this feature at all. That obviously needs to change. We should have some tests for the connection and pool size limits that they are enforced even when having multiple connections opened concurrently by clients.

The overall design seems to be in the right direction though.

binaryfire:
@GuanqunYang193 Just want to say thank you for working on this. It's going to simplify a lot of people's setups!

GuanqunYang193:
> @GuanqunYang193 Just want to say thank you for working on this. It's going to simplify a lot of people's setups!

I’m glad to see that the community really needs this feature — we’ll do our best to make it available to everyone!

GuanqunYang193:
There are already many flaky tests, and my change may have doubled the number. It took me a long time to determine whether the failures were caused by my changes.

doc/config.md
```diff
@@ -1193,6 +1193,12 @@ This is currently only supported on Linux.
 
 Default: 0
 
+### thread_number
```
AndrewJackson2020: Do we like the name `thread_number` for this parameter? To me it sounds like it is referring to a specific TID instead of the number of threads that will get spawned.

AndrewJackson2020: Another thoughts, even before this PR additional threads would be spawned for auth purposes (LDAP, PAM). Perhaps this name should also specify "woker" thread to distinguish it from "auth" threads

GuanqunYang193: Good idea! How about worker_thread_count?

I’m also working on a refactor to reduce the size of this PR. Please take a look when you have a chance!

AndrewJackson2020: I think worker_thread_count is more intuitive to me.

I'm not an expert on threading in C but happy to contribute whatever I can in terms of review.

AndrewJackson2020: Also one addition question: what is the different between thread_number=0 and thread number =1? Both would be single threaded but does 1 activate additional infra to accommodate multi threading but just with 1 thread? Possibly something to clarify in end user docs here

GuanqunYang193: I think this question actually makes the renaming more meaningful.
With thread_number = 0, PgBouncer behaves exactly as today (single-threaded).
With thread_number = 1, we have one acceptor (main thread) and one worker thread — effectively two threads in total, excluding auth threads.

AndrewJackson2020: Ah, understood, thank you

lib/usual/slab_ts.c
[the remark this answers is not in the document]
GuanqunYang193: We observed that aggressively locking shared data structures leads to deadlocks. As a result, we limit locking to a minimal set of critical data structures.

JelteF:
> On global limits: We already support global user limits and DB limits, but we still feel that pool size should be limited per thread rather than globally, since a global limit could introduce some issues:
> 
>     1. For min_pool_size and reserve_pool_size, how should they work globally? How would we distribute reserved pools across threads?
> 
>     2. Per-thread pool limits provide users with an accurate pool size estimate. A global limit, on the other hand, won’t be as precise as in single-thread mode (though the deviation may not be significant).
> 
> 
> Unlike user and DB limits, the pool is specific to PgBouncer. Adjusting its configuration should only affect PgBouncer’s performance and memory usage, making it more flexible. We also agree that changing the definition of some configurations would break compatibility. To address this, we could introduce new configurations, such as multithread_default_pool_size, which would override default_pool_size in multithreaded mode. Do you think that's a good idea?

Sorry for the super late response. You're making some good points on why this isn't easy to do. The reason why I pushed back on your decision was because I think it's important for users to be able to have some control over the number of connections pgbouncer makes to a server. Otherwise operators have no clue what `max_connections` they should be setting on their server. Currently  the `default_pool_size` is really the only way to control that.

I can think of three ways to do this:
1. Do it as you're saying and simply have completely separate pools per thread, and operators should simply do what they do now except instead of pool_size*num_instances they now also multiply by num_threads. I think the main downside is that the end-goal for me is to eventually set threads to `-1` by default, which would mean choose based on how many cores are in the system. The outgoing limit is then going to change depending on the system you are running on, which is going to be quite confusing for people.
2. Add "yet another" limit that allows configuring a global limit of outgoing connections. I think this will both complicate configuring pgbouncer ever further, and I'd expect it to interact badly with the per thread limits (i.e. you'd probably easily get into a state with unbalanced number of server connections per pool). 
3. Change the PR to not have the pools be split as much as they currently are: How about instead of only having a pool per thread, we'd also have a global pool, which would only contain idle server connections? All threads could then take an idle servers from that pool when they need one. They would also put them back there when they become idle again. Probably after a small delay so they can reuse it for a client they have waiting for a connection without having to take locks.

lib/usual/slab.c
```diff
@@ -64,38 +45,11 @@ static void slab_list_remove(struct Slab *slab)
 }
 
 /* fill struct contents */
-static void init_slab(struct Slab *slab, const char *name, unsigned obj_size,
-		      unsigned align, slab_init_fn init_func,
-		      CxMem *cx)
+static void init_slab_and_store_in_list(struct Slab *slab, const char *name, unsigned obj_size,
```
JelteF: This function is only used in one place. let's inline it. Same for it's thread-safe counterpart.

lib/usual/slab_ts.c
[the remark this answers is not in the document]
JelteF: FWIW: I think it can make sense to have thread safe and non-threadsafe versions of most of our datastructures.

lib/usual/spinlock.c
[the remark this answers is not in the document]
JelteF: Agreed. This PR is already complex enough without having to understand low-level locking implementations for multiple platforms. Let's use plain locks given to use by the underlying platforms. Either pthread mutexes/spinlock or some C11 stuff. Building our own locks seems a bit much for the initial version of this.

If either of those makes windows support for this non-trivial then I'd say let's leave multi-threading pgbouncer on windows for another day. I'm no Windows expert myself, and I don't think many people run pgbouncer in prod on windows. If they do, they can contribute windows threading support later once we have the core design tested on Unixes.

JelteF (commented):
The inline comments are some of my thoughts.

JelteF:
I'm having a hard time finding time to review this PR in depth so far because it requires more focus time than I have available for this at the moment. To at least be able to make some progress (because I would very much like this feature to exist) I asked my friendly LLM to do a review which I've left below. I honestly didn't even have time to fully read that review myself so take its recommendations with a grain of salt, but at least the first couple of "High" issues seemed like serious bugs. I specifically checked if H1 was real, which it is. And it's quite serious, I think we need some protection against bugs like that. Probably some syntactic analysis to check against such silly mistakes. The LLM suggested a Coccinelle rule, but I have no experience with that tool. Probably better to introduce that in a separate PR if that needs significant tooling changes. 

## High

### H1. `find_peer()` leaks the `peer_list` spinlock on every successful lookup

`src/objects.c:699-704`

```c
THREAD_SAFE_STATLIST_EACH(&peer_list, item, {
    peer = container_of(item, PgDatabase, head);
    if (peer->peer_id == peer_id)
        return peer;          /* returns with peer_list.lock still held */
});
```

`THREAD_SAFE_STATLIST_EACH` releases the lock *after* the body, so a `return` inside the
body skips the release. Because the lock is recursive the owning thread doesn't notice,
but `count`/`lock_word` never reset, so **any other worker iterating `peer_list` spins
forever**. `find_database()` immediately below was rewritten with a `found` flag to avoid
exactly this; `find_peer` was missed.

This is a **systemic pattern** — every non-local exit inside `WITH_LOCK(...)` or
`THREAD_SAFE_STATLIST_EACH(...)` leaks the lock. Confirmed occurrences:

| Location | Path |
| --- | --- |
| `find_peer` — `objects.c:702` | common (any successful peer lookup) |
| `add_dynamic_credentials` — `objects.c:624, 631` | `return NULL` on OOM leaks `user_lock` |
| `add_pam_credentials` — `objects.c:655, 662` | same |
| `cleanup_inactive_autodatabases` — `janitor.c:845` | `if (db->db_paused) return;` — also a logic regression (was `continue`) |
| `suspend_pooler` — `pooler.c:494` | `event_del` failure |
| `for_each_pooler_fd` — `pooler.c:665` | takeover callback failure |

### H2. DNS completions run on the main thread but drive worker-owned connection setup

`adns` is pumped **only** on the main thread (`main.c:829` `adns_per_loop`; workers never
call it). So `dns_callback` (`src/objects.c`) executes on the main thread and calls
`connect_server()` → `sbuf_connect()`, which does `event_assign`/`event_add` on
`workers[sbuf->thread_id].base` and mutates the worker-owned server, its pool's
(non-thread-safe) `new_server_list`/`active_server_list`, connection counters, etc.
With no `evthread` support, adding events to another running base is itself undefined,
on top of the state races. Any server host that needs DNS resolution (i.e. not a raw
IP / unix socket) hits this. DNS results need to be marshalled back to the owning
worker (e.g. over its pipe), not applied inline on the main thread.

### H3. Cross-thread cancel forwarding corrupts a foreign thread's slab

`src/objects.c:~2505`, `accept_cancel_request` searches *all* threads' pools for the
matching cancel key, then:

```c
req->pool = pool;                       /* pool may be owned by another thread */
WITH_LOCK(&pool->cancel_req_lock, {
    pause_cancel_request(req);
    launch_new_connection(pool, true);  /* executes on the wrong thread */
});
```

`launch_new_connection(pool)` allocates a server from
`workers[pool->thread_id].server_cache` — a **plain, non-thread-safe `struct Slab`** —
concurrently with that worker's own allocations, appends to that pool's `new_server_list`,
and registers the new socket's events on the other thread's base. `cancel_req_lock`
covers none of that. Can corrupt the slab freelist. The cancel needs to be handed to the
owning worker rather than executed cross-thread.

### H4. Admin commands deadlock (AB-BA) and touch foreign sockets under the wrong lock

Admin commands execute inside a worker callback, so they already hold
`workers[T].thread_lock`. Several then acquire *other* workers' thread_locks in ascending
`FOR_EACH_WORKER_THREAD` order (`admin_show_servers`, `admin_show_sockets`,
`admin_cmd_reconnect`, `admin_cmd_kill`, and `admin_cmd_reload` which grabs *all* of them).
Two admin sessions pinned to different worker threads running `SHOW SERVERS` concurrently
deadlock; `RELOAD` concurrent with any cross-thread admin command deadlocks more readily.
Monitoring tools polling `SHOW` on multiple connections make this realistic.

Related, same file:

- **`SHOW CLIENTS` / `SHOW FDS`** (`admin.c:856/877`, `464/480`) iterate other threads'
  `active_client_list`/`waiting_client_list`/`login_client_list` holding only the
  `pool_list` spinlock — but those inner lists are mutated by the owning thread under
  `thread_lock`, not the pool_list lock. → torn traversal / use-after-free while
  dereferencing `sk->pool->db->name`, `sk->vars`, etc. (`SHOW SERVERS`/`SOCKETS`
  correctly take `lock_worker_thread` — it was simply forgotten here.)
- **`admin_cmd_disable_internal`** (`admin.c:~1548`) sets `admin->sbuf.thread_id = thread_id`
  inside the per-thread loop and never restores it, so the admin socket ends up claiming a
  foreign thread and later `admin_ready()` registers events on the wrong base. It also
  dereferences `db` (`db->admin`, `db->db_disabled`) *before* the `if (db)` NULL check, so
  `DISABLE <nonexistent-db>` crashes.
- **`admin_wait_close_done`** (`admin.c:~2286`) mutates/disconnects other threads' admin
  sockets under each thread's `pool_list` lock instead of its `thread_lock` (the
  `// no lock needed` comment contradicts the code).
- **`KILL_CLIENT`**: `find_client_global` returns the target *after* releasing the owner's
  `thread_lock`; `admin_cmd_kill_client` re-locks and calls `disconnect_client` — TOCTOU
  use-after-free if the owner frees it in between.

### H5. Prepared-statement cache: unlocked refcount and ID counter

`src/prepare.c`

- `ps->use_count += 1` (`:94`, `:114`) and `--server_ps->ps->use_count` (`:199`) run
  **outside** `prepared_statements_lock`, even though `ps` lives in the global
  `prepared_statements` hash shared by all workers. Non-atomic RMW loses updates, and the
  find-then-increment / decrement-then-free sequences aren't a single critical section →
  premature `free()` and use-after-free of a shared `ps`.
- `next_unique_query_id += 1` (`:68`) is incremented with no lock (the call sits *between*
  the two locked blocks of `get_prepared_statement`). Two threads can mint the **same**
  `query_id`, which is used as both the hash key and the server-side statement name →
  wrong-statement routing.

### H6. `refresh_stats` races with worker stat updates

`src/stats.c:368-377` runs on the main thread and writes `pool->older_stats`/
`pool->newer_stats` and reads `pool->stats` holding only the `pool_list` spinlock. But
workers increment `pool->stats.*` under their `thread_lock`. Two different locks for the
same fields ⇒ no mutual exclusion (torn 64-bit reads on 32-bit platforms, lost counters).
`admin_database_stats_internal` does this correctly with `lock_worker_thread`;
`refresh_stats` and `show_stat_totals` (`stats.c:310-317`) don't.

---

## Medium

### M1. `seq` is never incremented in multithread mode → autodb idle detection broken

`janitor.c:890` only does `seq++` when `!multithread_mode`, and `main_thread_full_maint`
never increments it despite the comment saying it should. With `seq` stuck at 0,
`db->active_stamp >= seq - 2` becomes `active_stamp >= UINT_MAX-1` (unsigned wrap) → false
for active autodbs, so **in-use auto-databases get moved to the idle list and reaped**.

### M2. Connection-limit overshoot (TOCTOU)

`launch_new_connection` reads `db->connection_count`/`user->connection_count` in one
`WITH_LOCK` (`objects.c:2176`, `:2193`) and does the matching `++` in a *separate*
`WITH_LOCK` later (`:2227`, `:2229`). Two workers can both observe `< max` and both
increment → `max_db_connections`/`max_user_connections` overshoot by up to
`num_threads - 1`. (`check_db_connection_count`/`check_user_connection_count` in
`client.c` have the same split-region shape.)

### M3. LDAP/PAM polling skipped in multithread mode

`main_loop_once` skips `ldap_poll()` / `pam_poll()` when `multithread_mode`
(`main.c:819-826`), yet `auth_ldap_init()`/`pam_init()` still run. LDAP/PAM auth will
hang. Either drive them per-worker or reject the combination in
`validate_multithread_limits`.

### M4. `client_free` reads freed memory

`objects.c:233` `slab_free(client_cache_, client)` then `:237` reads
`client->contributes_global_client_count`. Currently benign (the flag sits past the bytes
`list_init` clobbers), but a genuine use-after-free — capture the flag into a local before
freeing. If ever read as garbage-true it double-decrements the shared `client_count`.

### M5. `cf_pause_mode` written cross-thread with no lock / barrier / wakeup

`admin_cmd_pause` (`admin.c:1469`), `full_resume` (`:1388`) and `admin_cmd_suspend`
(`:1445`) write other threads' `cf_pause_mode` without `lock_worker_thread`, memory
barrier, or wakeup; workers read it unlocked. Late/missed PAUSE/RESUME visibility.
`SUSPEND` additionally only sets the issuing thread's mode.

### M6. Windows pipe framing not enforced

`handle_request` (`pooler.c:439`) reads `sizeof(ClientRequest)` and only checks `<= 0`;
`pool_accept` treats any `> 0` write as success. Safe on Linux (≤ `PIPE_BUF`, atomic), but
the Windows `evutil_socketpair`/`SOCK_STREAM` path has no atomicity — concurrent writers
(`pool_accept` + `wakeup_worker_thread`) can interleave/short-read, and a partially filled
struct is then used as a real fd.

### M7. Pooler event mutation from worker threads

`suspend_pooler`/`resume_pooler`/`pooler_tune_accept` mutate `pgb_event_base`'s
listen-socket events. If `SUSPEND`/`RESUME` is dispatched to a worker (admin clients are
handled by workers), this is a cross-thread mutation of the main base with no `evthread`.

---

## Low / design notes

### L1. Maintenance runs outside `thread_lock`

`worker_func` (`multithread.c:49-52`) runs `per_loop_maint()`,
`reuse_just_freed_objects()`, `rescue_timers()`, `per_loop_pooler_maint()` **outside** the
worker's `thread_lock`. So `SIGHUP`/`RELOAD` and `PAUSE`, which grab all thread_locks to
fence workers, do not actually exclude these routines — and `config_postprocess`/
`kill_database` free objects into per-thread (non-thread-safe) slabs that a worker's
`per_loop_maint` may be touching concurrently. Decide whether maintenance should also run
inside the wrapper.

### L2. Spinlock design

Correctness in the common path is fine (each worker's own lock is normally uncontended),
but: `SpinLock.count` is overloaded as both the CAS flag and the recursion depth;
`lock_word` (a `pthread_t`) is read non-atomically in `spin_lock_owns` (works on Linux,
fragile elsewhere); and holding a spinlock (spinning with `sched_yield`) across an entire
event callback burns a core for any cross-thread waiter. Also: **recursion is what makes
the H1/H3-class leaks silent** — a non-recursive lock would self-deadlock the leaking
thread at its own next acquire, surfacing the bug locally. Audit which locks truly need
`enable_recursive = true`.

### L3. `put_in_order` split-lock TOCTOU

`objects.c:~487` searches under the list lock but, on the append path, releases and
re-acquires — a TOCTOU on the shared `user_list`/`database_list`/`peer_list`. Currently
masked because callers hold `user_lock` / run at config load, but fragile.

### L4. Torn reads of lock-guarded counters

`SHOW DATABASES`/`USERS`/`LISTS` and `count_db_active` read `db->connection_count`,
`user->connection_count`, per-thread slab counts, etc. without the relevant lock. No
crash, just inconsistent numbers.

### L5. Confusing pointer comparisons

`get_pool`/`evict_connection` compare `pool->db->name == db->name` (pointer comparison of
the embedded array) where they mean `pool->db == db`. Equivalent, but confusing.

### L6. `get_active_client_count()` semantics

Now returns only the current thread's slab count; correctly replaced with
`get_total_active_client_count()` for the `max_client_conn` check, but still used for the
shutdown-drain checks at `janitor.c:969,984` — confirm that's intended to be per-thread.

PR #1390, opened 2025-09-09
ci: Combine some Ubuntu tasks

petere:
We can run --disable-evdns and --without-openssl in one job, because the code related to these two options is independent of each other.


petere:
(I adjusted the branch protection rules to take this changed check into account.)

PR #1395, opened 2025-09-19
Don't close connection with query timeout when no query is in flight

imsurajmishra:
Fixing the bug discussed here :
https://github.com/pgbouncer/pgbouncer/pull/629


JelteF:
You don't seem to actually have committed a fix. Only included a test.

imsurajmishra:
Just checking if there is anything else needs to be done here

JelteF (approved):

PR #1405, opened 2025-10-08
Rename auth_ldap_parameter

eulerto:
The commit cabe794 uses auth_ldap_parameter but this name is conceptually wrong. There isn't a single parameter but multiple ones (ldapserver, ldapport, ...). Postgres refers the LDAP elements as "options" [1] instead of "parameters".  Adopt the same terminology as Postgres. It also removes the "auth" prefix because "ldap" implies that this PgBouncer parameter is for authentication. The bonus is a shorter name. The new parameter name is ldap_options.

[1] https://www.postgresql.org/docs/current/auth-ldap.html


petere:
But I see that we use "LDAP options" in the PostgreSQL documentation and messages, so it seems it's ok, and it's probably better to be consistent. So let's leave it as is now.

PR #1409, opened 2025-10-17
[postgres-proxy] log send buffer stalls

ericli-oai:
Log whenever sends stall. We'll use this to see the effects of tuning send buffers.


ericli-oai:
hmm, meant this for my fork.

PR #1410, opened 2025-10-28
LDAP documentation cleanup

petere:
Clean up the documentation around the new LDAP authentication feature a bit.  Some of the text was poorly worded.

Also change one error message to be more consistent with a recent PostgreSQL change (PostgreSQL commit f68d85bf692).

PR #1411, opened 2025-10-28
Rename ldap_options to auth_ldap_options

petere:
The argument in commit 1a95d51 that the "auth" prefix is unnecessary is flawed.  There are also other ways in which LDAP could be used in a PostgreSQL stack, and they would use different LDAP options.  Explicit is better.  Also, PgBouncer groups all related settings under common prefixes, and they sort close together, so it seems good to maintain that.


eulerto (approved):
LGTM.

petere:
Rebased over #1410.

PR #1412, opened 2025-11-09
Add stat field to track client connections

AndrewJackson2020:
This PR adds the ability to track client connection counts from the admin console. 

While pgbouncer allows clients much greater performance in workloads that make fast connections/disconnects I have seen extreme instances where this clients will overuse this performance and cause issues in other areas of the network stack. For example a client will run an ETL that loops through a list of data to insert in a multiprocessed manner, create and close a connection for each row to be inserted. While the performance looks a lot better using pgbouncer than native postgres, it would still be better to find instances of clients doing this and be able to nudge them onto better practices (COPY from STDIN in this instance). It also exposes the clients to nondeteriministic network stack errors from the OS/libc, at least from what I have seen.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -186,6 +186,9 @@ total_xact_count
 total_query_count
 :   Total number of SQL commands pooled by **pgbouncer**.
 
+total_client_connect_count
+:   Total times that a client connected to a pool
+
 total_server_assignment_count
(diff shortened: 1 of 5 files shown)


AndrewJackson2020:
Closing in favor of https://github.com/pgbouncer/pgbouncer/pull/1457

PR #1417, opened 2025-11-12
C11 enablement

petere:
This is a series of patches to enable C11 mode by default and clean away some workarounds for older versions.

This mirrors similar work I'm doing on PostgreSQL.

---
(This can of course wait until after a 1.25.1 release, if that is needed.)


JelteF (approved):

lib/usual/talloc.c
```diff
@@ -90,7 +87,7 @@ struct TLimit {
  */
 
 static void log_to_stderr(const char *message);
-static void do_abort(const char *fmt, ...) _NORETURN;
+_NORETURN static void do_abort(const char *fmt, ...);
```
JelteF, line 90: If this is the only place we use this, how about we change to the C11 casing instead of keeping our own define?

petere: I tried to explained this in the commit message:

> We keep the _NORETURN macro that is currently in use.  Using the
    standard <stdnoreturn.h> header and noreturn notation has the
    potential for conflicts with third-party headers that use some variant
    of `__attribute__((noreturn))`.  So for now we keep this until we have a
    better idea.  Maybe converge on pg_noreturn that is used in
    PostgreSQL.

PR #1434, opened 2025-12-02
Make warnings in tests behave as errors

JelteF:
In PR #1264 a few tests were not correctly awaiting all coroutines. This
caused warnings in the tests (and I think also some flakiness). This
fixes that by sprinkling await in a few more places. To avoid this
happening again in the future this also starts treating warnings in
tests as errors. If later we want to ignore some warnings we can
explicitly do that.


petere:
I'm not qualified to review this code in detail, but cleaning up these warnings is highly welcome.

PR #1441, opened 2025-12-08
Change types in `SHOW USERS` admin command

AndrewJackson2020:
In the `SHOW USERS` command, `pool_size` and `reserve_pool_size` are both sent to the client as strings instead of integers. There may be a reason for this that I'm not seeing but it seems like an oversight as these fields also show up in `SHOW DATABASES` and `SHOW PEERS` as integers. This probably constitutes a breaking change because there is probably a lot of downstream code that depends on these being strings and not integers. Maybe that precludes merging this, just figure I would raise a PR just in case.

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -603,8 +603,6 @@ static bool admin_show_users(PgSocket *admin, const char *arg)
 	struct List *item;
 	PktBuf *buf = pktbuf_dynamic(256);
 	struct CfValue cv;
-	char pool_size_str[12] = "";
-	char res_pool_size_str[12] = "";
 	const char *pool_mode_str;
 
 	if (!buf) {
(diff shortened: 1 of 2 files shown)


JelteF:
I agree there's a backwards compatibility concern here. What's the benefit that we're gaining from doing that? i.e. what do you care that we send them as integers?

AndrewJackson2020:
> I agree there's a backwards compatibility concern here. What's the benefit that we're gaining from doing that? i.e. what do you care that we send them as integers?

I think that I was writing some code for monitoring that interfaced with pgbouncer and ran into issues before I realized that these fields are sent as strings. Figured I would send in a PR as I'm sure many other people have ran into this issue.

If the backwards compatibility concerns override any benefit this changes provide please feel free to close this PR.

PR #1310, opened 2025-04-30
Increase the maximum size of passwords to 65535

davecramer:
Cloud services use very long passwords and this value is in line with libpq

PR #1446, opened 2025-12-10
Correctly handle large and partial server login packets

DAlperin:
This issue was discovered when there was an intermediate proxy like so: pgbouncer -> other proxy -> upstream db. If for whatever reason `other proxy` split a login packet into multiple parts, we would bail the connection. This fix is modeled after the fix in 
https://github.com/pgbouncer/pgbouncer/pull/1058.

If the packet is simply incomplete we yield the sbuf back and wait for the next event loop tick. If it's bigger than the sbuf_len it uses the existing packet buffering logic.


JelteF:
I need to look closer at this, but it sounds like this would fix the issue reported here, i.e. with a large password causing the startup packet being larger than the pktbuf: https://github.com/pgbouncer/pgbouncer/pull/1310

JelteF (approved):

PR #1449, opened 2025-12-12
Implement user and database level query_wait_timeout

AndrewJackson2020:
This PR expands the query_wait_timeout to be configurable at the user level and database level.

The goal here is to allow some users to have a longer and more dangerous query_wait_timeout to allow them a usage pattern of "blast pgbouncer with hundreds of simultaneous queries" and allow it to throttle queries to postgres without causing contention on the machine. This can allow pgbouncer to take the role of an asyc queue or producer/consumer architecture.

Also this PR adds unit tests for the existing global query_wait_timeout functionality.


doc/config.md
```diff
@@ -1342,6 +1342,15 @@ they are logged but ignored otherwise.
...
 the default `pool_mode` is used.
 
+### query_wait_timeout
+
+Maximum time queries are allowed to spend waiting for execution.
+0 disables. -1 means this value is not set. [seconds]
+
+See description of the global `query_wait_timeout` setting for additional detail.
+
+If not set, the database or default `query_wait_timeout` is used.
```
JelteF:
```suggestion
If not set, the user or default `query_wait_timeout` is used.
```

src/janitor.c
```diff
@@ -405,9 +405,11 @@ static void pool_client_maint(PgPool *pool)
 	struct List *item, *tmp;
 	usec_t now = get_cached_time();
 	PgSocket *client;
+	PgDatabase *db;
```
JelteF: Nit: let's define this in the smallest scope that we need it in.

JelteF (approved):
Overall this looks good. Some small comments.

src/janitor.c
AndrewJackson2020: Agreed. The scope has been reduced with the latest commit. 

Tried declaring the variable within the `statlist_for_each_safe` block but if failed the valgrind test. Moved it to within the `if` statement 2 lines up and now its fine.

Is there something about declarding PgDatabase within `statlist_for_each_safe` that is memory unsafe?

Also its failing FreeBSD though I suspect that is a flaky test.

Thank you for the review.

Issue #1354, opened 2025-07-12
Native prometheus exporter


shivampaw:
Following on from #289 i wanted to revive this as 5 years on k8s is used everywhere. 

The solutions out there right now have issues such as on terminating a pgbouncer instance, they can’t scrape stats because pgbouncer stops accepting connections. 

Also, none can currently provide histogram support, but by doing it natively, we can do that. And therefore provide a lot of percentile and more useful metrics.

AndrewJackson2020:
Wrt access during shutdown I believe that #1305 being merged should help with this.

Issue #289, opened 2018-04-25
Native Prometheus exporter support


stanhu:
We're looking at instrumenting pgbouncer with Prometheus instead of relying on third-party exporters. Would you accept a PR to add native Prometheus support?

SuperQ:
For context, this would involve adding a micro embedded http server that would allow access to the data from `show stats`, and similar.

The format itself is a simple read-only text[0] representation of the data.

There are two current exporters, both have issues that we would like to eliminate by having native instrumentation.

https://github.com/spreaker/prometheus-pgbouncer-exporter 
https://github.com/larseen/pgbouncer_exporter/

[0]: https://prometheus.io/docs/instrumenting/exposition_formats/

vthriller:
> both have issues that we would like to eliminate by having native instrumentation

I'm curious: what issues are you talking about here?

SuperQ:
The `spreaker` exporter in Python, which we would prefer to not need for our packaging sanity.

The `larseen` exporter is better, but has a number of naming/metrics best practices problems.  Since I posted the comment, we created a [fork](https://github.com/stanhu/pgbouncer_exporter/) with a number of fixes.  I will try and get this sent upstream.

The main reason we would like native metrics support is it simplifies deploys by eliminating the need for a 3rd party metrics tool.

petere:
I don't see us adding an HTTP server into PgBouncer.

PJMODOS:
Not to mention I don't think we want to integrate into one specific monitoring solution (unless it's couple lines of code, which HTTP server definitely isn't). I think the time is better spent pushing forwards projects like https://github.com/influxdata/telegraf/pull/3918 TBH.

eakraly:
How about https://github.com/launchdarkly/c-client? Now should be pretty straightforward.

halcyondude:
@PJMODOS given what's happened in the past 18 months since this issue was closed (k8s adoption, OpenMetrics, widespread adoption of Prometheus, CNCF, Cortex, ..., etc), IMHO this should be revisited.

We would greatly prefer a native integration vs. relying on an exporter, perhaps feature flags could allow for supporting OpenMetrics in a selective "opt-in" way.

SuperQ:
There's a new [Prometheus C library](https://github.com/digitalocean/prometheus-client-c) that might be worth looking at.

MSDehghan:
Any updates on this?

PR #1462, opened 2026-01-25
Connections Balancing

o-alexandre-felipe:
# Improves existing load balancing strategy.
This pull requests implements load balancing based on the number of active connections per host.

## Rationale
Consider the following (extreme) example. Server A has 10 connections of which 1 is idle. Server B has 10 connections of which 10 is idle. A user decides to run `SELECT * FORM ALL TABLES ORDER BY some json field` because it is easy, and that uses all the memory, all the CPU, starts to use temporary file. If the the user was assigned to server B we are fine, if the user was assigned to server A that will degrade the experience of 9 other users.

Now on a more subtle scenario. We know that each session will compete for resources with the existing sessions. So if a new user is opening a session to minimize the potential impact of a performance degradation we should connect it to the server that has the least amount of sessions.

## Change

The current load balancing strategy consists in starting new connections in a round robin manner. So that the number of open connection to a database is fixed. In a production environment during spikes the connections will be open, and after that the clients will be always assigned to any idle connection.

What this pull request does is to make sure that when incoming requests will be assigned to a server that has the least amount of active connections.

## Constraints

Different databases can have either identical or disjoint pools, e.g. `{db1: [a], db2: [b]}`, `{db1: [a,b], db2: [a,b], db3: [c,d]}` are all valid, but `{db1: [a], db2: [a,b]}`, is invalid because `a` appears in two distinct host pools. A pool cannot have references to the same host e.g. `{db1: [a,a,b]}`.

## Algorithm

A structure `PgHost` was introduced to hold details about each host, and track its active connections. A structure `PgHostPool` was introduced to keep track activity on all hosts of a database. 

A hash array of hosts is maintained in order to match hosts by name. Multiple references of to the same host will be internally represented to pointers to a single object. If a pool has more than one host it will use a bucket array to track hosts by number of connections.

### Server connection
Connecting to a pool with `n` hosts involves
 * `n` host lookups
 * Check duplicates `O(n)` using reference counters.
 * Pool validation also runs in `O(n)` by storing at most one pool pointer.
 * Potentially allocating a hash element.
Nothing serious.

### Enter `SV_IDLE`
Append the socket to the server connection list this required one struct List to be added to `PgSocket`.

### Enter or leave `SV_ACTIVE`.

Increment/decrement the active connections counter and, if more than one host is in the bool, move the host from one bucket to another one `list_del` and one `list_append`.

### When finding server

Normally with only one database we can get a with `first(first(pool->buckets[pol->min_active]) -> idle_server_list)`, where first is just abstraction of get first element in a list. So, average case is `O(1)` and the worst case involves (1) checking all buckets difference between the maximum and minimum active connections per host, very small when load is balanced; (2) checking all hosts (2) checking all idle connections. Not great but still limited.

The performance could be improved adding shortcuts between long gaps. i.e. there are many servers with low load but no usable idle connections (e.g. different users).

### Additional memory

With `sizeof(int) = 4`, and `sizeof(void*) = 8`,

| Struct | Size | Added | Usage |
|--------|------|-------|-------|
| `PgHost` | - | 112 | One per distinct host:port |
| `PgHostPool` | - | 56 | One per distinct host list |
| `PgSocket` | 760 | +16 | One per connection |
| `PgPool` | 720 | +8 | One per (user, database) pair |

Additionally each host keeps an array bucket for faster lookup of server with least connections, adding `(HostPool idle connections) * 16 / (pool size)`

Tests:
[ ] test_load_unload - Verify round-robin distributes connections evenly
[ ] test_failover_uses_first_host - Verify failover mode uses first host
[ ] test_single_port_for_single_host - Verify basic single host/port works
[ ] test_multi_host_single_port - Verify single port replicated to multiple hosts (requires USE_SUDO=1)
[ ] test_connection_reuse_prefers_least_loaded - Verify idle connections reused from least-loaded host
[ ] test_port_count_mismatch_error - Verify config error on mismatched host/port counts
[ ] Existing test_load_balance_hosts_* tests pass


o-alexandre-felipe:
@petere @eulerto @JelteF 
Looking into the automated checks I see.
> Not enough compute credits to prioritize tasks!

This is the first of a series of pull requests that I intend to open.
- Fine control: of load balancing level. Per database, per user, per credentials, global, or none, where none will be the backward compatible.
- Next it would be nice to have a more modern configuration format, e.g. JSON or YAML.
- Metrics: On a production environment it is important to have metrics, so I plan to add a prometheus metrics server.

AndrewJackson2020:
> @petere @eulerto @JelteF Looking into the automated checks I see.
> 
> > Not enough compute credits to prioritize tasks!
> 
> This is the first of a series of pull requests that I intend to open.
> 
>     * Fine control: of load balancing level. Per database, per user, per credentials, global, or none, where none will be the backward compatible.
> 
>     * Next it would be nice to have a more modern configuration format, e.g. JSON or YAML.
> 
>     * Metrics: On a production environment it is important to have metrics, so I plan to add a prometheus metrics server.


Id be curious what the maintainers think of these proposals. 

For JSON/YAML are you going to introduce any build or runtime dependencies? Currently the config basically just tracks what postgres offers with postgresql.conf/hg_hba.conf which is fine but gets a bit awkward when you are trying to do anything involving nesting. One example of this is the user/database level config, especially as we more options are added. If pgbouncer stays with the postgres style config what would probably happen is more config files like how postgres has the user mapping file.

With regard to prometheus pgbouncer already exposes the admin console which is readable via third party prometheus exporters like [0]. This also is how postgres handles this. That being said this does seem to be an often requested feature but has been rejected before [1] [2].
 
[0] https://github.com/prometheus-community/pgbouncer_exporter. 
[1] https://github.com/pgbouncer/pgbouncer/issues/1354
[2] https://github.com/pgbouncer/pgbouncer/issues/289

docker/Dockerfile
```diff
@@ -0,0 +1,118 @@
+FROM ubuntu:22.04
```
AndrewJackson2020, line 1: What are these new docker assets used for? If this is separate from the goal of "Connections Balancing" I would think that this should be in a separate PR.

o-alexandre-felipe: Thank you for your time @AndrewJackson2020 ,

The only reason was for testing.
I can separate if this is the only a blocker for the PR, but I would prefer to do so at the end of the review, if we decide that it is necessary to split.

Thank you for pointing out pgbouncer_exporter

o-alexandre-felipe: Created a side car runningg [dockerized pgbouncer-exporter](https://hub.docker.com/r/prometheuscommunity/pgbouncer-exporter), and is already running in our development environment.

As easy as it should be.

AndrewJackson2020: IMO PR's should be as targeted as possible. The maintainers of this project are very busy and might not be able to review your work for a while. When they do you are going to have a lot more success with a handful of smaller PR's than one giant PR. It also makes it easier for them to review.

If it were me I would also split out the multi port functionality into its own PR as that would be a much smaller PR and would provide benefit outside of "Connections Balancing".

o-alexandre-felipe: @AndrewJackson2020 thank you for your recommendation. I opened a pull request with a simpler implementation (that was possible only because now I understand the codebase better).

o-alexandre-felipe: #1464 Comments are welcome.

PR #1463, opened 2026-01-28
Fix SCRAM auth failure when using forced user with plaintext password

### LINKED ISSUE #1461: SCRAM authentication fails when using forced user with different plaintext password
## Description

When a database is configured with a forced user and plaintext password (`user=X password=Y`), and a client authenticates using SCRAM-SHA-256 with a different password, backend authentication fails.

## Configuration

```ini
[databases]
mydb = host=localhost user=backend_user password=secret123
```

```
# userlist.txt
"client_user" "SCRAM-SHA-256$4096:salt$StoredKey:ServerKey"
```

## Reproduction

1. Client connects as `client_user` with SCRAM authentication (succeeds)
2. PgBouncer connects to backend as `backend_user` (fails: "password authentication failed")

Direct connection to PostgreSQL with the same password works.

## Cause

After client SCRAM auth, PgBouncer saves the client's SCRAM keys to `pool->user_credentials`. With a forced user, this points to `forced_user_credentials`, corrupting it with the client's keys and setting `use_scram_keys = true`.

During backend auth, PgBouncer sees `use_scram_keys = true` and uses the cached (wrong) keys instead of deriving from the plaintext password.

## Disclaimer

The root cause analysis was in part done with the help of AI agents.

l2dy:
Fix #1461

When a database has a forced user configured with a plaintext password, the pool credentials point to forced_user_credentials. The code that saves SCRAM keys after client authentication was incorrectly saving the client's keys to the pool credentials, corrupting the forced user's credentials and causing backend authentication to fail.

Fix by skipping SCRAM key saving only when:
1. There is a forced user, AND
2. The forced user has a non-empty password, AND
3. That password is plaintext (not a SCRAM/MD5 hash)

This preserves the existing behavior for:
- Non-forced user configurations (keys always saved)
- Forced user with no password (keys saved for pass-through auth)
- Forced user with hashed password (keys saved)

## Disclaimer

The solution is implemented by Claude Code and only tested against a limited number of scenarios.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -1345,8 +1345,15 @@ static bool handle_client_startup(PgSocket *client, PktHdr *pkt)
 				if (!mbuf_get_bytes(&pkt->data, length, &data))
 					return false;
 				if (scram_client_final(client, length, data)) {
-					/* save SCRAM keys for user */
-					if (!client->scram_state.adhoc && !client->db->fake) {
+					/*
+					 * Save SCRAM keys for user, but not when forced user
+					 * has plaintext password (would corrupt credentials
+					 * and cause backend auth to fail).
+					 */
+					if (!client->scram_state.adhoc && !client->db->fake &&
+					    !(client->db->forced_user_credentials &&
+					      client->db->forced_user_credentials->passwd[0] &&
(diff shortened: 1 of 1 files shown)


eulerto:
The PR #1612 fixed it so I'm going to withdraw this one.

PR #1470, opened 2026-02-19
Fix server connection stuck after COPY error due to late CopyDone

fvannee:
### Problem                                                                                                                                                                            
                                                                                                                                                                                         
  In transaction pooling mode, server connections can become permanently stuck in "active" state after a COPY FROM STDIN fails server-side (e.g. due to a constraint violation). Once    
  stuck, the connection is never returned to the pool, effectively reducing pool capacity. With enough occurrences, the entire pool becomes exhausted and all clients queue indefinitely.
   Killing the stuck postgres backends is the only recovery.

We observed this bug in production and managed to reproduce it in dev.

  ### How to reproduce

  1. Configure pgbouncer in transaction pooling mode (any pool_size, pool_size=1 is sufficient)
  2. In a client, run COPY FROM STDIN (binary) inside an explicit transaction (`BEGIN; COPY ...; COMMIT;`)
  3. The batch must be large enough that the client is still sending CopyData messages when the server encounters an error (e.g. a unique constraint violation early in the batch)
  4. Under moderate concurrency, the server connection gets stuck within seconds

  The race window depends on batch size: larger batches mean more data in flight after the server errors, increasing the likelihood. With 5000 rows of ~300 bytes each (~1.5MB total),
  reproduction is near-immediate.

  ### Root cause

  When the server hits an error during COPY, the following protocol exchange occurs:

  1. Server sends `ErrorResponse` — pgbouncer sets `copy_mode=false`, calls `clear_outstanding_requests_until("cf")` which finds no CopyDone/CopyFail in the queue yet (only the original
   Query entry, which gets cleared), leaving `outstanding_requests=0`
  2. Server sends `ReadyForQuery` (state `'E'`, aborted transaction) — nothing to pop, `query_failed` reset
  3. Client's late `CopyDone` arrives (the client was still sending data when the server errored) — unconditionally added to `outstanding_requests`, bumping it to 1

  This CopyDone is never popped from the queue: postgres ignores the late CopyDone (it already left COPY mode) and sends no response for it. The phantom entry in `outstanding_requests`
  permanently prevents `release_server()` from returning this connection to the pool.

  Each subsequent COPY error on the same connection adds another orphan entry, and the count only grows.

  ### Fix

  In `handle_client_work()`, only track CopyDone/CopyFail as outstanding requests when the linked server is still in `copy_mode`. If the server has already exited copy mode (because it
  sent ErrorResponse), the CopyDone is still forwarded to postgres (which ignores it), but no entry is added to `outstanding_requests`.

This PR is the simpler fix that only fixes COPY within an explicit BEGIN/END transaction. My preference would be to merge https://github.com/pgbouncer/pgbouncer/pull/1471 instead which fixes it both inside and outside the transaction.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -1458,7 +1458,21 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 	/* copy end markers */
 	case PqMsg_CopyDone:
 	case PqMsg_CopyFail:
-		track_outstanding = true;
+		/*
+		 * Only track as outstanding if the server is still in copy_mode.
+		 *
+		 * When a COPY fails server-side (e.g. constraint violation), the
+		 * server sends ErrorResponse which sets copy_mode=false and clears
+		 * outstanding requests up to 'c'/'f'. If the client's CopyDone
+		 * arrives after this, the server will have already sent
+		 * ReadyForQuery. Postgres ignores the late CopyDone, so no
+		 * server response will arrive for it. Tracking it would create
+		 * an entry in outstanding_requests that never gets popped,
+		 * permanently preventing release_server() from returning this
+		 * connection to the pool.
+		 */
+		if (client->link && client->link->copy_mode)
+			track_outstanding = true;
 		break;
 
 	/*


JelteF:
Closing in favor of #1471

Issue #57, opened 2015-06-11
bulk COPY may result in CLOSE_WAIT socket state


sezal:
CopyData (packet type 'd') expected to end with CopyDone (packet type 'c'), but in some cases (when any error occurred) CopyData seems to be processed incompletely.

See extra/unprocessed pkt='d' packet is coming after pkt 'E'

```
2015-06-11 16:05:38.748 31090 NOISE safe_send(10, 2048) = 2048
2015-06-11 16:05:38.748 31090 NOISE resync: done=2048, parse=2048, recv=2048
2015-06-11 16:05:38.748 31090 NOISE safe_recv(9, 2048) = 2048
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=217
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=216
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=217
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=219
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=217
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=215
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=217
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=216
2015-06-11 16:05:38.748 31090 NOISE C-0x2654040: ts_traffic/ts_test@127.0.0.1:53215 pkt='d' len=219
2015-06-11 16:05:38.749 31090 NOISE safe_send(10, 2048) = 2048
2015-06-11 16:05:38.749 31090 NOISE resync: done=2048, parse=2048, recv=2048
2015-06-11 16:05:38.749 31090 NOISE safe_recv(9, 2048) = 2048
[...]

PR #1471, opened 2026-02-19
Fix late COPY messages causing pool exhaustion in transaction mode

fvannee:
This is a more extensive fix that can be applied instead of https://github.com/pgbouncer/pgbouncer/pull/1470
The main difference is that this also allows COPY to work fine *outside* of an explicit transaction (BEGIN/END) (eg. see some background that this currently doesn't work in https://github.com/pgbouncer/pgbouncer/issues/57)
This is another issue we hit a while ago, which was mitigated by explicitly wrapping the client code in BEGIN/END, but I think it makes more sense if it's just properly supported by pgbouncer.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1320,7 +1320,6 @@ static void unlink_server(PgSocket *server, const char *reason)
 
 	client = server->link;
 
-	client->copy_mode = false;
 	client->link = NULL;
 	server->link = NULL;
 	/*
--- a/src/objects.c
+++ b/src/objects.c
@@ -1214,6 +1214,7 @@ bool release_server(PgSocket *server)
 	case SV_BEING_CANCELED:
 	case SV_ACTIVE:
 		if (server->link) {
+			server->link->copy_mode = false;
 			server->link->link = NULL;
 			server->link = NULL;
 		}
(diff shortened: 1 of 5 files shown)


src/objects.c
```diff
@@ -1319,6 +1320,7 @@ static void unlink_server(PgSocket *server, const char *reason)
 
 	client = server->link;
 
+	client->copy_mode = false;
```
JelteF: Why is this one needed? It seems strange that we'd have to change the copy_mode if we're going to disconnect it right after.

fvannee: Yeah you're right it's not strictly necessary. Mostly defensive to clear state here. I'm happy to remove it.

fvannee: Removed it

fvannee: Would you have any other thoughts on this or the other PR, @JelteF? Is it a contribution you'd be willing to accept?

JelteF (approved):

fvannee:
Hi @JelteF , I noticed you marked this with the "Next release" label, but it wasn't shipped with last week's release. Hereby a gentle ping as I hope it won't be forgotten :)

Issue #1482, opened 2026-03-20
RELOAD with invalid config file may not fully preserve the previous active configuration


fheikens:
### Summary

The documentation for `RELOAD` (and `SIGHUP`) describes the success path clearly, but does not specify what guarantees apply when the config file is invalid and the reload fails. Reading the code suggests the "failed reload preserves old config" guarantee may not hold in all cases.

### What seems unclear

Looking at `load_config()` in `src/main.c` (around line 459) and the config parser in `lib/usual/cfparser.c`:

1. **Incremental writes during parse**: `cf_load_file` applies config values to global variables one at a time via `cf_set()` as each line is parsed. When a section header is encountered, `fill_defaults()` first resets all reloadable parameters in that section to their compile-time defaults, then the file's key=value pairs overwrite those defaults line by line. If a parse error occurs mid-section, parameters appearing before the error will have new (or file-specified) values, while parameters after the error will have been reset to **defaults** — not the previously running values.

2. **Peer dead-flag asymmetry**: Before loading, both `set_dbs_dead(true)` and `set_peers_dead(true)` are called (lines 470–471). On failure, `set_dbs_dead(false)` is called to undo the marking (line 486), but `set_peers_dead(false)` is not. Since `config_postprocess()` runs unconditionally after loading (line 517) and kills all entries with `db_dead == true`, a failed reload appears to kill all peer connections.

3. **No config preservation test**: The existing `test_reload_error` in `test/test_admin.py` verifies that `RELOAD` returns a `ConfigFileError` when the config file is invalid, but does not check whether the previous configuration values are actually preserved (e.g., via `SHOW CONFIG` before and after).

### Why this matters operationally

Operators frequently rely on `RELOAD` / `SIGHUP` being safe to run in production — the expectation is that a typo or bad merge in the config file should be caught, an error should be reported, and the running configuration should remain fully intact. If a failed reload can partially update the active config (or kill peers), it creates a subtle failure mode that's hard to diagnose: the process keeps running but with an unexpected mix of parameter values.

### Evidence

- `src/main.c:466–488` — `load_config()` failure path
- `lib/usual/cfparser.c:349–377` — `fill_defaults()` resets reloadable params to compile-time defaults before applying file values
- `lib/usual/cfparser.c:288–336` — `cf_set()` writes to global config variables immediately during parse
- `src/janitor.c:957–977` — `config_postprocess()` kills peers still marked dead
[...]

PR #1483, opened 2026-03-20
Add regression test for config preservation on failed RELOAD (#1482)

fheikens:
## Summary

Adds an xfail regression test that demonstrates the behavior described
in #1482: a failed `RELOAD` (due to an invalid config file) does not
fully preserve the previous active configuration.

The test starts PgBouncer with known non-default values, writes a config
file with a valid parameter change followed by an invalid value that
triggers a parse error, issues `RELOAD`, and checks `SHOW CONFIG`
afterward. It verifies three cases:

- a reloadable parameter changed before the error line (currently gets
  the new value instead of being preserved)
- a reloadable parameter absent from the bad config file (currently gets
  reset to its compile-time default instead of being preserved)
- the parameter whose invalid value caused the error (currently gets
  reset to its compile-time default instead of being preserved)

The test is marked `xfail` and will begin passing once the underlying
issue in `load_config()` / `cf_load_file()` is addressed.

## Test plan

- `test_reload_error` still passes (no regression in existing coverage)
- `test_reload_error_preserves_config` xfails on the expected assertion
- `--runxfail` confirms failure is `default_pool_size changed from 5 to 42`
- `black --check` and `isort --check` pass

Refs #1482

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -884,6 +884,13 @@ connection is next released (according to the pooling mode), and new
 server connections will immediately use the updated connection
 parameters.
 
+If the configuration file cannot be parsed (for example because of a
+syntax error or an invalid parameter value), the reload fails and an
+error is reported.  In that case the previously active configuration is
+preserved in full: reloadable parameters keep their running values, and
+existing database and peer connections are left untouched.  A failed
+reload never leaves PgBouncer with a partially applied configuration.
+
 #### WAIT_CLOSE [db]
 
 Wait until all server connections, either of the specified database or
--- a/src/main.c
+++ b/src/main.c
@@ -460,10 +460,92 @@ static bool requires_auth_file(int auth_type)
 	return auth_type >= AUTH_TYPE_TRUST;
 }
 
+/*
+ * Snapshot and restore of the reloadable [pgbouncer] parameters.
+ *
+ * cf_load_file() applies values to the live configuration as it parses,
+ * and resets reloadable parameters to their compile-time defaults when it
+ * enters the [pgbouncer] section. If the file contains a parse error part
+ * way through, the live configuration is left in a mix of new values, old
+ * values, and compile-time defaults. To keep a failed reload from changing
+ * the running configuration, snapshot the reloadable parameters before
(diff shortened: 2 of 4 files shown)


fheikens:
Updated: this is now a complete fix for #1482 (config + peer preservation), not just the xfail regression test. Rebased onto current `master` (v1.25.2). Details and the open design question are in #1482.

fheikens:
Heads-up on the red CI: the failures are pre-existing flaky tests, not from this change.

I pushed twice with byte-identical file contents (only the commit timestamp differed to re-trigger CI), and the failures moved between platforms each run:

- Run 1: Windows failed `test_server_id`, plus xdist worker crashes in `test_notify_queue_negative` and `test_auth_dbname_with_auto_database`; FreeBSD and everything else passed.
- Run 2: Windows passed cleanly; FreeBSD failed `test_prepared_statement_pipeline_stress`.

macOS, the Valgrind Linux job, and the Linux configs pass consistently in both runs. None of these tests exercise the config-reload path this PR touches, and the two tests added here (`test_reload_error_preserves_config`, `test_failed_reload_preserves_peers`) pass everywhere they run. The `test_notify_queue*` failures also emit pre-existing `coroutine ... was never awaited` warnings, independent of this change.

Since the same tree produces different failures on different platforms, this looks like flakiness in the suite rather than anything introduced here. Happy to rebase or re-run if a maintainer can kick the flaky jobs.

fheikens:
Rebased onto current `master` (was ~35 commits behind and conflicting). The only conflict was in `test/test_admin.py` where `master` and this branch each added a new test function at the same spot — resolved by keeping both (`test_show_user` and `test_reload_error_preserves_config`); the `src/main.c` fix merged cleanly.

Validated locally against current `master` (Debian, gcc, PgBouncer 1.25.2 build):
- builds + links clean;
- `test_reload_error_preserves_config` (config preservation), `test_failed_reload_preserves_peers` (peer preservation), the existing `test_reload_error` (no regression), and `test_show_user` all pass.

No behavioural change from the previous version of the PR — just un-stale and conflict-free for review.

PR #1488, opened 2026-04-08
Add notify message on client connection

AndrewJackson2020:
This PR adds a welcome message that displays as a NOTICE message after the client connects to pgbouncer. This is embedded as apart of the welcome message. The use case here is that pgbouncer is a mostly transparent connection pooler but there are enough exceptions to that rule that it makes sense to let clients know when they connect that this is not in fact a postgres direct connection.

This is based on neons welcome message[0]. This works because the postgres wire protocol allows NOTICE messages to be sent during start-up[1], after AuthenticationOk but before ReadyForQuery.

[0] https://neon.com/blog/how-psql-pg-neon-tech-works
[1] https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-START-UP

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -817,6 +817,14 @@ A value of 0 disables this notification message.
 
 Default: 5
 
+### login_welcome_message
+
+Welcome notify message that is sent to the client after a login
+is successful. Can be used to ensure that clients undertand that
+they are connecting to pgbouncer instead of postgres directly.
+
+Default: empty (no welcome message sent)
+
 ## TLS settings
 
 If the contents of any of the cert or key files are changed without
--- a/doc/config.md
+++ b/doc/config.md
@@ -817,7 +817,7 @@ A value of 0 disables this notification message.
 
 Default: 5
 
-### login_welcome_message
+### login_notify_message
 
 Welcome notify message that is sent to the client after a login
 is successful. Can be used to ensure that clients undertand that
--- a/src/proto.c
+++ b/src/proto.c
@@ -304,10 +304,10 @@ bool welcome_client(PgSocket *client)
 
 	pktbuf_write_BackendKeyData(msg, client->cancel_key);
 
-	if (!(strcmp(cf_login_welcome_message, "") == 0)) {
+	if (!(strcmp(cf_login_notify_message, "") == 0)) {
 		pktbuf_write_Notice(
 			msg,
-			cf_login_welcome_message
+			cf_login_notify_message
 			);
 	}
 
--- a/src/proto.c
+++ b/src/proto.c
@@ -304,7 +304,7 @@ bool welcome_client(PgSocket *client)
 
 	pktbuf_write_BackendKeyData(msg, client->cancel_key);
 
-	if (!strcmpeq(cf_login_welcome_message, "")){
+	if (!strcmpeq(cf_login_welcome_message, "")) {
 		pktbuf_write_Notice(
 			msg,
 			cf_login_welcome_message
--- a/src/proto.c
+++ b/src/proto.c
@@ -304,7 +304,7 @@ bool welcome_client(PgSocket *client)
 
 	pktbuf_write_BackendKeyData(msg, client->cancel_key);
 
-	if (!strcmpeq(cf_login_welcome_message, "")) {
+	if (!(strcmp(cf_login_welcome_message, "") == 0)) {
 		pktbuf_write_Notice(
 			msg,
 			cf_login_welcome_message
--- a/src/proto.c
+++ b/src/proto.c
@@ -304,6 +304,13 @@ bool welcome_client(PgSocket *client)
 
 	pktbuf_write_BackendKeyData(msg, client->cancel_key);
 
+	if (!strcmpeq(cf_login_welcome_message, "")){
+		pktbuf_write_Notice(
+			msg,
+			cf_login_welcome_message
+			);
+	}
+
 	/* finish */
 	pktbuf_write_ReadyForQuery(msg);
 	if (msg->failed) {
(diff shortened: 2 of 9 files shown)


dvarrazzo:
TBH there are better ways than a NOTICE, that's for human beans.

The right way to do it would be a [parameter status](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQPARAMETERSTATUS).

AndrewJackson2020:
> TBH there are better ways than a NOTICE, that's for human beans.
> 
> The right way to do it would be a [parameter status](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQPARAMETERSTATUS).

So this PR is actually meant to provide a human readable message on login that is disabled by default.

JelteF:
I think we should have a ParameterStatus message too. Probably `pgbouncer.version` and then containing the version.

AndrewJackson2020:
> I think we should have a ParameterStatus message too. Probably `pgbouncer.version` and then containing the version.

you want that in this PR or in a different one?

JelteF:
Either would be fine to me. Separate one would be nice from a "small commits" perspective. But if that then causes a merge conflict I don't think it's worth it.

test/test_misc.py
```diff
@@ -17,6 +17,67 @@
...
+    admin_users = pgbouncer
+    auth_file = {bouncer.auth_path}
+    listen_port = {bouncer.port}
+    logfile = {bouncer.log_path}
+    pool_mode = session
+    """
+
+    with bouncer.run_with_config(config):
+
+        ret = bouncer.psql(
```
JelteF, line 43: Instead of using psql I think it makes more sense to use the psycopg notice handler. We already hook into that if I'm not mistaken to log the notices.

AndrewJackson2020: Unfortunately because of an existing limitation in psycopg, I dont think this is currently possible. I originally tried doing it the way you described but found that you can only set the notice handler after the connection has been made. I raised an issue with psycopg [0]. Also I documented this in the docstring just to be sure.

I could possibly experiment with if its possible with the async interface, maybe its possible with that.

[0] https://github.com/psycopg/psycopg/issues/1291

AndrewJackson2020:
> Either would be fine to me. Separate one would be nice from a "small commits" perspective. But if that then causes a merge conflict I don't think it's worth it.

Ill open a separate one. Might have to wait until after the commitfest though.

src/proto.c
```diff
@@ -304,6 +304,13 @@ bool welcome_client(PgSocket *client)
 
 	pktbuf_write_BackendKeyData(msg, client->cancel_key);
 
+	if (!strcmpeq(cf_login_welcome_message, "")) {
```
petere: What new hell is this? Can we stick with strcmp() please?

doc/config.md
```diff
@@ -817,6 +817,14 @@ A value of 0 disables this notification message.
 
 Default: 5
 
+### login_welcome_message
```
petere: Maybe this should be named something like "login_notify_message" instead? That would make the mechanics clearer. The word "welcome" is just filler, doesn't mean anything. The message is not required to be welcoming. ;-)

src/proto.c
AndrewJackson2020: Done.

Just for context `strcmpeq` is apart of the `libusual` library. The comment above its definition explains its intended purpose. I have used it in previous PR's. Definitely seems like `strcmp` is more widely used than `strcmpeq` in pgbouncer but it does have a presence.
 
```console
 > grep --after-context=10 --before-context=3 strcmpeq lib/usual/string.c
/*
 * Same as strcmp, but handles NULLs. If both sides are NULL, returns "true".
 */
bool strcmpeq(const char *str_left, const char *str_right)
{
        if (str_left == NULL && str_right == NULL)
                return true;

        if (str_left == NULL || str_right == NULL)
                return false;

}
```

```console
> rg 'strcmp\(' | wc -l
163

> rg 'strcmpeq\(' | wc -l
22
```

doc/config.md
AndrewJackson2020: True. I have some coworkers who would like nothing more than to use this feature to berate users into using sqlserver instead of postgres.

name changed. 

```
find . -type f -exec sed -i 's/login_welcome_message/login_notify_message/g' {} \+
```

AndrewJackson2020:
> I think we should have a ParameterStatus message too. Probably `pgbouncer.version` and then containing the version.

Is this roughly what you had in mind?

https://github.com/pgbouncer/pgbouncer/pull/1490

src/proto.c
JelteF: FWIW, the original spelling of this was `strings_equal` and it was introduced in 0ba795f0, to avoid introducing more of this kind of horror (and remove the existing ones):

```c
		} else if (!!host != !!db->host
			   || (host && strcmp(host, db->host) != 0)) {
```

Then it was moved to libusual as `strcmpeq` in https://github.com/libusual/libusual/pull/57

doc/config.md
```diff
@@ -817,6 +817,14 @@ A value of 0 disables this notification message.
 
 Default: 5
 
+### login_notify_message
+
+Welcome notify message that is sent to the client after a login
+is successful. Can be used to ensure that clients undertand that
```
JelteF, line 823:
```suggestion
is successful. Can be used to ensure that interactive clients understand that
```

test/test_misc.py
JelteF: oh that's annoying... Then I think the current approach seems fine.

JelteF (approved):

PR #1490, opened 2026-04-10
Send PgBouncer specific ParameterStatus messages to client

### LINKED ISSUE #653: Detecting pgbouncer mode at connection
Hello,

psycopg 3 can use prepared statements, and when a statement is prepared its name is cached in the connection object. This is likely not compatible with pgbouncer used at transaction level.

It would be good to detect the presence of pgbouncer and its transaction level at connection time, and eventually to disable the prepared statements behaviour

Is it possible, ideally via [`PQparameterStatus`](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQPARAMETERSTATUS)?

[edit: a bunch of "prepared transactions" errors fixed]

AndrewJackson2020:
Allows clients/drivers to change behaviour or disable features when connecting to PgBouncer. 

Related #1488
Fixes #653

### DIFF
--- a/src/admin.c
+++ b/src/admin.c
@@ -1912,7 +1912,7 @@ void admin_setup(void)
 
 	pktbuf_write_ParameterStatus(msg, "pgbouncer.pool_mode", "statement");
 	pktbuf_write_ParameterStatus(msg, "pgbouncer.version", PACKAGE_VERSION);
-	pktbuf_write_ParameterStatus(msg, "pgbouncer.max_prepared_statements", max_prepared_statements);
+	pktbuf_write_ParameterStatus(msg, "pgbouncer.max_prepared_statements", "0");
 
 	if (msg->failed)
 		die("admin welcome failed");
--- a/src/admin.c
+++ b/src/admin.c
@@ -1910,6 +1910,10 @@ void admin_setup(void)
 	pktbuf_write_ParameterStatus(msg, "standard_conforming_strings", "on");
 	pktbuf_write_ParameterStatus(msg, "is_superuser", "on");
 
+	pktbuf_write_ParameterStatus(msg, "pgbouncer.pool_mode", "statement");
+	pktbuf_write_ParameterStatus(msg, "pgbouncer.version", PACKAGE_VERSION);
+	pktbuf_write_ParameterStatus(msg, "pgbouncer.max_prepared_statements", max_prepared_statements);
+
 	if (msg->failed)
 		die("admin welcome failed");
 
--- a/src/admin.c
+++ b/src/admin.c
@@ -1909,12 +1909,9 @@ void admin_setup(void)
 	pktbuf_write_ParameterStatus(msg, "standard_conforming_strings", "on");
 	pktbuf_write_ParameterStatus(msg, "is_superuser", "on");
 
-	pktbuf_write_ParameterStatus(msg, "pgbouncer.max_prepared_statements", "0");
-
 	if (msg->failed)
 		die("admin welcome failed");
 
-	pool->welcome_msg = msg;
 	pool->welcome_msg_ready = true;
 
 	msg = pktbuf_dynamic(128);
--- a/src/admin.c
+++ b/src/admin.c
@@ -1898,10 +1898,9 @@ void admin_setup(void)
 	}
 
 	/* prepare welcome */
-	msg = pktbuf_dynamic(128);
+	msg = pool->welcome_msg;
 	if (!msg)
 		die("out of memory");
-	pktbuf_write_AuthenticationOk(msg);
 	pktbuf_write_ParameterStatus(msg, "server_version", PACKAGE_VERSION "/bouncer");
 	pktbuf_write_ParameterStatus(msg, "client_encoding", "UTF8");
 	pktbuf_write_ParameterStatus(msg, "server_encoding", "UTF8");
@@ -1910,8 +1909,6 @@ void admin_setup(void)
 	pktbuf_write_ParameterStatus(msg, "standard_conforming_strings", "on");
 	pktbuf_write_ParameterStatus(msg, "is_superuser", "on");
 
-	pktbuf_write_ParameterStatus(msg, "pgbouncer.pool_mode", "statement");
-	pktbuf_write_ParameterStatus(msg, "pgbouncer.version", PACKAGE_VERSION);
 	pktbuf_write_ParameterStatus(msg, "pgbouncer.max_prepared_statements", "0");
 
 	if (msg->failed)
--- a/src/main.c
+++ b/src/main.c
@@ -39,6 +39,8 @@
 #include <sys/resource.h>
 #endif
 
+#include <math.h>
+
 /*
  * Default number of iterations when generating secret.  Should be at least
  * 4096 per RFC 7677.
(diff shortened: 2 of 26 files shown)


dvarrazzo:
Hello,

what would be the use case of overriding the server version of the database? I think that this would add more confusion in the client-server communication and I really fail to see the usefulness.

You can add extra information and declare `pgbouncer_version` and `pgbouncer_pool_mode`: the latter would be pretty useful. No need to obfuscate a valid value.

See #653 for old and stale ideas. The more you hide the fact that pgbouncer is in the loop the worse the experience for the end user and the expectation of clients misbehaving.

AndrewJackson2020:
> Hello,
> 
> what would be the use case of overriding the server version of the database? I think that this would add more confusion in the client-server communication and I really fail to see the usefulness.
> 
> You can add extra information and declare `pgbouncer_version` and `pgbouncer_pool_mode`: the latter would be pretty useful. No need to obfuscate a valid value.
> 
> See #653 for old and stale ideas. The more you hide the fact that pgbouncer is in the loop the worse the experience for the end user and the expectation of clients misbehaving.

Yeah okay I think I missed the point of Jelte's comment over in  https://github.com/pgbouncer/pgbouncer/pull/1488. Converted this into a draft, will update it with what you  are saying above (which is what I believe his intent was). Thank you for the review.

AndrewJackson2020:
@dvarrazzo updated this PR more along the lines that you described. Let me know if this looks good to you.

dvarrazzo:
It looks useful. Ideally it would be useful to send information about what could reasonably affect the conversation between a client and the server. For example Postgres [sends info](https://www.postgresql.org/docs/current/libpq-status.html#LIBPQ-PQPARAMETERSTATUS) about date and time format, string escaping rules.

The version is good to have. I don't know precisely what to do with it because I don't have extensive PgBouncer experience (for example prepared statements weren't supported until a few versions ago). Knowing in which mode pgbouncer is operating might also be useful maybe? For example if we know we are in `pool_mode = statement` then we may start in autocommit mode and throw an error if a customer tried to run an explicit transaction.

I would be happy to receive suggestions about what behaviour should change in psycopg according to specific PgBouncer settings.

src/proto.c
```diff
@@ -231,6 +231,8 @@ bool add_welcome_parameter(PgPool *pool, const char *key, const char *val)
 	if (msg->write_pos == 0)
 		pktbuf_write_AuthenticationOk(msg);
 
+	pktbuf_write_ParameterStatus(msg, "pgbouncer_version", PACKAGE_VERSION);
```
JelteF: Let's use postgres extension style scoping:

```suggestion
	pktbuf_write_ParameterStatus(msg, "pgbouncer.version", PACKAGE_VERSION);
```

JelteF: Agreed with @dvarrazzo let's add a `pgbouncer.pool_mode` too.

JelteF: Let's also add `pgbouncer.max_prepared_statements`. Together with `pgbouncer.pool_mode`, that should be enough for clients to detect whether prepared statements are supported.

AndrewJackson2020: Okay so I went ahead and added `pgbouncer.max_prepared_statements` and `pgbouncer.pool_mode`. `pgbouncer.max_prepared_statements` was straightforward but `pgbouncer.pool_mode` has one complication. I'm currently using the function `probably_wrong_pool_pool_mode` which is probably wrong, doesn't take into account replication connections which are always session.  I had issues trying to use `connection_pool_mode`. I think that the `set_replication` needs to be set before we can use `connection_pool_mode`.  

I also added these parameters to the admin dashboard just to be consistent and wrote tests for that. Also expanded the tests so that all pool modes are verified. 

For max_prepared_statements we probably want to store this as a string once globally as opposed to reconverting it to a string every time a parameter is sent? Also probably I shouldn't assume the size of int (for string conversion) and use limits?

AndrewJackson2020: Went ahead and changed max_prepared_statements to be a global variable that is reused. I think I need to test that it is resilient to a reload though.

Also implemented connection_pool_mode function usage. Need to figure out if it's possible to test that part.

petere:
Should this be a draft?

AndrewJackson2020:
> Should this be a draft?

Apologies, in the future I will:
1. Draft PRs if I am implementing feedback and convert back to "ready for review" only when they are actually ready for review and all feedback has been addressed to the best of my ability
2. I will dev on a different branch and squash before merging with the PR branch so there is less noise/clutter on the PR

src/proto.c
```diff
@@ -338,6 +339,20 @@ bool add_welcome_parameter(PgPool *pool, const char *key, const char *val)
 	if (msg->write_pos == 0)
 		pktbuf_write_AuthenticationOk(msg);
 
+	pktbuf_write_ParameterStatus(msg, "pgbouncer.version", PACKAGE_VERSION);
```
JelteF: Seems like a bit of a weird function to add this. The fact that we write the AuthenticationOk message here is pretty weird too. Seems better not to continue bolting onto that strange design. And the current code seems wrong too, since these new pktbuf_write calls will happen N times: once for each welcome parameter this function is called for (since they are not part of the `write_pos == 0` block)

AndrewJackson2020: I took a stab at fixing the architecture of the `welcome_msg` initialization with the last commit. Basically now instead of getting initialized in `add_welcome_parameter` guarded with conditional logic to check if it exists beforehand, it now is encompassed in its own function `new_welcome_msg`. `new_welcome_msg` is called in `new_pool` as apart of the pool initialization process and `tag_pool_dirty` because the pool's cached `welcome_msg` gets invalidated in this function.

Let me know if this is in the direction of what you had in mind. Also happy to split this off into a separate PR if this is out of scope for this one.

src/admin.c
```diff
@@ -1910,6 +1909,8 @@ void admin_setup(void)
 	pktbuf_write_ParameterStatus(msg, "standard_conforming_strings", "on");
 	pktbuf_write_ParameterStatus(msg, "is_superuser", "on");
 
+	pktbuf_write_ParameterStatus(msg, "pgbouncer.max_prepared_statements", "0");
+
```
JelteF: this line seems unnecessary here.

The `pool->welcome_msg = msg;` line a few lines below also seems unneeded.

src/main.c
```diff
@@ -493,6 +517,16 @@ bool load_config(void)
 		ok = false;
 	}
 
+	if (strcmp(cf_max_prepared_statements, "0") == 0) {
+		max_prepared_statements = 0;
+	} else {
+		max_prepared_statements = atoi(cf_max_prepared_statements);
+		if (max_prepared_statements == 0) {
+			die("invalid max_prepared_statements value in config");
```
JelteF: `die` is quite extreme in case of a typo in the config on reload

src/main.c
```diff
@@ -493,6 +517,16 @@ bool load_config(void)
 		ok = false;
 	}
 
+	if (strcmp(cf_max_prepared_statements, "0") == 0) {
+		max_prepared_statements = 0;
+	} else {
+		max_prepared_statements = atoi(cf_max_prepared_statements);
```
JelteF: Let's use `strtonum` for this parsing instead of `atoi`

AndrewJackson2020: This is fixed now.

src/main.c
AndrewJackson2020: `die` has been removed

src/admin.c
AndrewJackson2020: These 2 lines have been removed.

PR #1492, opened 2026-04-11
Correctly propagate system errors for SSL. Fixes #956

### LINKED ISSUE #956: Failed to load certificate file does not correctly propagate error
When pgbouncer fails to load a certificate file specified in client_tls_cert_file the underlying error is not propagated. This makes debugging extremely difficult for end users.

Example log line:
```
Oct 03 21:35:17 pgbouncer pgbouncer[9652]: TLS setup failed: failed to load certificate file "/root/cert.crt": (null)
```

3Davydov:
In OpenSSL 3.0.0 and later, ERR_reason_error_string function will return NULL if given error code relates to the system error. Replace this function with ERR_error_string in order to propagate strings describing system errors too.


JelteF:
Can you include a test for this too? Other than that, this seems like a good change

3Davydov:
Hi, that you for looking into this!
I have added a simple test that ensures that we can see messages generated by system errors. Only single possible error is tested. I guess it is OK since we just need to verify specific code path.

BTW, in the issue I have wrote that we can make output of such system errors more compact. That is, get rid of this string : `error:80000002:system library::` . Please, let me know if it's worth the effort.

3Davydov:
@JelteF , Hi!

Can you please remove the `need-tests` label? I'm afraid that it can confuse somebody, since tests are already added.
(IIUC, I cannot remove this label by myself)

JelteF (approved):

JelteF:
thank you

PR #1496, opened 2026-05-04
Implement Multi Port Database Connection String

AndrewJackson2020:
Currently pgbouncer supports multi host connection strings for load balancing and fail over which is nice. Unfortunately pgbouncer makes the assumption that each of these hosts has a postgres server listening on the same port. No flexibility is allowed to have different ports listening on different hosts.

This PR attempts to add the ability to add a list of comma separated ports if multiple ports are specified. 

Validations are added to ensure that number of ports must match number of hosts, each element of the csv must be integers, etc.

Potential issues:
- breaking change to `show databases`. Port is now a string data type to accommodate a comma separated list of ports
- ports are stored as a comma separated list, the port is extracted and converted to int when used. This was the same approach taken with host list but may not be appropriate here.
-  I am specifically disallowing this functionality in peering


petere:
This appears to be in active development. Should this be a Draft PR, or a PR at all?

AndrewJackson2020:
> This appears to be in active development. Should this be a Draft PR, or a PR at all?

Apologies, I won't dev against PR branches in the future. Will convert to draft.

PR #1503, opened 2026-05-09
Fix several issues with ldap/pam auth workers. Fixes #799

### LINKED ISSUE #799: Valgrind detects memory leak when pam is enabled
```
==45571== VALGRIND-ERROR-BEGIN
==45571== 304 bytes in 1 blocks are possibly lost in loss record 3 of 3
==45571==    at 0x483AB65: calloc (vg_replace_malloc.c:760)
==45571==    by 0x4012D16: allocate_dtv (dl-tls.c:343)
==45571==    by 0x4012D16: _dl_allocate_tls (dl-tls.c:589)
==45571==    by 0x4D01B81: allocate_stack (allocatestack.c:622)
==45571==    by 0x4D01B81: pthread_create@@GLIBC_2.2.5 (pthread_create.c:660)
==45571==    by 0x1263E2: pam_init (pam.c:129)
==45571==    by 0x1148C0: main (main.c:1011)
==45571== 
==45571== VALGRIND-ERROR-END
```

It seems to be because we don't clean up the thread at exit and rely on OS behaviour.

Found by @eulerto when testing #792

3Davydov:
Please, read about these changes in my first comment in the [issue 799](https://github.com/pgbouncer/pgbouncer/issues/799)


AndrewJackson2020:
It might make sense to split this into 2 different PR's. There is no test coverage over PAM but there is for LDAP. Might be easier/safer to merge LDAP than PAM because of this.

Also might make sense too add LDAP/PAM to Valgrind testing in CI with this change (assuming no other issues crop up after).

3Davydov:
Hi,
Thank you for looking into this!

> It might make sense to split this into 2 different PR's

Changes for LDAP and PAM workers are completely the same. I don't think that having two PR's will be handy here, since every fix in one of the PR's will require the same fix in the second PR. I'd rather add tests for PAM worker here, if you don't mind.

> Also might make sense too add LDAP/PAM to Valgrind testing in CI with this change (assuming no other issues crop up after).

Hm, okay. I'll try to implement it.

3Davydov:
I have added a test for PAM and also included pam&ldap to the test with valgrind. Please, see the latest commit in the branch.

P.S.
As far as I understand, we cannot write a normal test for PAM, since it is depending on the system (config files, users) which requires a superuser privileges to operate on. Please, correct me if I am wrong

PR #1504, opened 2026-05-11
Fix SCRAM pass-through with stored SCRAM secrets after reconnect

kathia-barahona:
The fix in bf06ce1f62a66e1c5b7b89cd9fe873370c1859f9 (https://github.com/pgbouncer/pgbouncer/pull/1432) unconditionally sets state->adhoc = true
when the cached path is hit. This is correct for plaintext passwords
(whose adhoc SCRAM secrets must not be saved for pass-through), but
incorrect for stored SCRAM secrets. When the password in userlist.txt
is a real SCRAM hash, the ClientKey/ServerKey extracted from the client
authentication are valid for server pass-through and must be saved.

Because the adhoc_scram_secrets_cached flag is shared by both plaintext
and SCRAM hash passwords, the second client connection after a server
reconnect incorrectly marks real SCRAM credentials as adhoc, preventing
key saving and causing server login to fail with "password is SCRAM
secret but client authentication did not provide SCRAM keys".

Fix by only setting state->adhoc when the stored password is actually
plaintext. Also add a regression test.

Related to https://github.com/pgbouncer/pgbouncer/pull/1432, https://github.com/pgbouncer/pgbouncer/pull/1338

### DIFF
--- a/src/scram.c
+++ b/src/scram.c
@@ -919,7 +919,8 @@ char *build_server_first_message(ScramState *state, PgCredentials *user, const c
 			goto failed;
 	} else {
 		if (user->adhoc_scram_secrets_cached) {
-			state->adhoc = true;
+			if (get_password_type(stored_secret) == PASSWORD_TYPE_PLAINTEXT)
+				state->adhoc = true;
 			state->iterations = user->scram_Iiterations;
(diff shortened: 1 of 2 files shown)


JelteF:
Thank you!

PR #1507, opened 2026-05-11
Openssl3 eof noise

### LINKED ISSUE #1506: Noisy TLS warnings on unexpected EOF
OpenSSL 3 makes a choice to be noisy when the client doesn't gracefully shut down a connection. (This was apparently a choice made before with 1.1.1e, but then rolled back with 1.1.1f... see https://github.com/php/php-src/issues/8369).

This affects pgBouncer by putting warnings like:

```WARNING tls_sbufio_recv: read failed: error:0A000126:SSL routines::unexpected eof while reading```

into log files when clients abruptly hang up. pgBouncer has code to deal with this situation, but it needs to be revisited for OpenSSL 3. There is a config file option to get openssl itself to go back to less noisy behavior, but only starting in 3.2.

benchub:
Fix #1506 by correctly working with openssl 3's new flags to know
that a client had an unexpected EOF, and that we don't really need
to log anything about it. This maintains pgBouncer's previous
behavior as seen on openssl <3.

This pull request adds a test, showing the problematic behavior, a fix
for it, and a successful test pass after the fix is in place.


benchub:
I see the formatting failed on ./Makefile, but I didn't change that file, so I'm uncertain what to do about the failure.

petere:
> I see the formatting failed on ./Makefile, but I didn't change that file, so I'm uncertain what to do about the failure.

The GitHub view of this is misleading. You can get a better view on Cirrus directly: https://cirrus-ci.com/task/5928529105256448

The failure is in the file `lib/usual/tls/tls.c`.

benchub:
Thanks @petere, that was the grip I needed to make headway. Looks good to me now.

PR #1508, opened 2026-05-12
crypto: Prevent optimization of sensitive data erasure

ProjectMutilation:
Replace memset(lanebuf, 0, ...) with explicit_bzero to prevent compiler from
optimizing(Dead Store Elimination) away the zeroing of cryptographic data on stack.
Fixes potential information leak where sensitive data could remain on stack after function exit.


3Davydov:
Hi,

Thank you for the patch! A few comments :
1. An ability to call non-optimized version of `memset` can be useful not only within the `kessak.c` file. I suggest adding declaration of this function to the one of the frequently-used headers.
2. Please, see the `explicit_bzero.c` file in the PostgreSQL source code. It solves the same problem in more reliable way. I think that we can implement that same logic within this patch.

petere:
PgBouncer already has all the portability scaffolding that you can use `explicit_bzero()` without further ceremony. So just call it directly.

3Davydov:
Oops, I didn't notice that this was already implemented in the source code. For sure, in this case it is enough to call existing `explicit_bzero` function.

ProjectMutilation:
> PgBouncer already has all the portability scaffolding that you can use `explicit_bzero()` without further ceremony. So just call it directly.

Thanks, replaced

PR #1510, opened 2026-05-15
Fix errors pointed out by static analyzer

3Davydov:
Hi,

A static analyzer found several errors :
1. Potential null pointer dereference (fixed in the first commit)
2. Potential memory leak (fixed in the second commit)

Please, see fixes in this pull request.

***
Comments on the implementation :
`varcache.c` already contains example of out-of-memory handling (see `apply_var` function). I decided to do the same, i.e. just `die`.

### DIFF
--- a/src/varcache.c
+++ b/src/varcache.c
@@ -71,8 +71,10 @@ static void init_var_lookup_from_config(const char *cf_track_extra_parameters, i
 		HASH_FIND_STR(lookup_map, var_name, lookup);
 
 		/* If the var name is already on the hash map, do not update its idx */
-		if (lookup != NULL)
+		if (lookup != NULL) {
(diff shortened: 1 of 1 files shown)


hlinnaka (approved):
Looks good to me

>  Comments on the implementation :
varcache.c already contains example of out-of-memory handling (see apply_var function). I decided to do the same, i.e. just die.

Yeah, that seems fine here. This code runs at pgbouncer startup. If we supported reloading `track_extra_params`, we'd maybe need to work a little harder.

It's actually worse for `apply_var()`, that's executed while pgbouncer is up and running, and dying on out of memory would not be so nice. But that's a pre-existing issue.

JelteF (approved):

PR #1512, opened 2026-05-30
Replace CirrusCI with Github Actions

JelteF:
Cirrus CI has shut down. This starts using Github Actions as our CI instead. It loses FreeBSD CI for now. We could add that back later if we really want, but it would require us using nested virtualization. 

This also fixes a few flaky tests that came up often while I was testing this.


.github/workflows/pgbouncer-ci.yml
```diff
@@ -0,0 +1,401 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+  # Cirrus marked a number of tasks `trigger_type: manual`. Those are mapped
+  # onto the workflow_dispatch event so they can be started by hand from the
+  # Actions UI ("Run workflow") instead of running on every push.
+  workflow_dispatch:
```
petere, line 12: This didn't work for me. I didn't see a "Run workflow" button in my fork. Maybe something is misconfigured?

.github/workflows/pgbouncer-ci.yml
```diff
@@ -0,0 +1,401 @@
+name: CI
```
petere, line 1: Please rename the file to something like `pgbouncer-ci.yml` so it is less likely to confuse in downstream forks.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+          ./configure --prefix="$HOME/install2" --enable-werror --without-cares $configure_args
+          make -j4
+          make -j4 install
+      - name: Upload tarball
+        if: ${{ matrix.dist == 'yes' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: pgbouncer-tarball
+          path: pgbouncer-*.tar.gz
+          if-no-files-found: ignore
```
petere: This seems wrong. If the dist step produces no matching files then it's probably broken and should fail?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+      - name: Upload config.log
+        if: failure()
```
petere: In Cirrus we had this to always upload. I have found this useful. We should keep it that way.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+    name: Determine Linux matrix
+    runs-on: ubuntu-latest
+    outputs:
+      linux_matrix: ${{ steps.set.outputs.linux_matrix }}
+    env:
+      EVENT_NAME: ${{ github.event_name }}
+    steps:
+      - id: set
+        run: |
+          core='{"name":"cares-pam-ldap","runner":"ubuntu-22.04","pgversion":"16","configure_args":"--with-cares --with-pam --with-ldap","cflags":"","enable_valgrind":"","scan_build":"","dist":"yes"},{"name":"no-evdns-no-openssl","runner":"ubuntu-22.04","pgversion":"16","configure_args":"--disable-evdns --without-openssl","cflags":"","enable_valgrind":"","scan_build":"","dist":""},{"name":"valgrind","runner":"ubuntu-22.04","pgversion":"17","configure_args":"","cflags":"-O0 -g","enable_valgrind":"yes","scan_build":"","dist":""}'
```
petere: The valgrind job previously used PG 18.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+          name: config-log-macos
+          path: config.log
+          if-no-files-found: ignore
+
+  windows:
+    name: Windows
+    runs-on: windows-2022
+    env:
+      HAVE_IPV6_LOCALHOST: "yes"
+      # Use the MSYS2 installation preinstalled at C:\msys64 on the runner
```
petere: This comment doesn't explain why, or why that is relevant or interesting.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+      # rather than a third-party setup action. MSYSTEM picks the MINGW64
```
petere: The comment about MSYSTEM is content-free.

.github/workflows/pgbouncer-ci.yml
```diff
@@ -0,0 +1,401 @@
...
+  windows:
+    name: Windows
+    runs-on: windows-2022
+    env:
+      HAVE_IPV6_LOCALHOST: "yes"
+      # Use the MSYS2 installation preinstalled at C:\msys64 on the runner
+      # rather than a third-party setup action. MSYSTEM picks the MINGW64
+      # toolchain and CHERE_INVOKING keeps the login shell in the checkout dir.
+      MSYSTEM: MINGW64
+      CHERE_INVOKING: "1"
```
petere, line 359: Move the comment sentence about `CHERE_INVOKING` to before this line.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+      - uses: actions/checkout@v4
+      - name: Install pandoc
+        shell: pwsh
+        run: choco install -y --no-progress pandoc
+      - name: Configure hosts file
+        shell: pwsh
+        run: Add-Content c:\Windows\System32\Drivers\etc\hosts "127.0.0.1 localhost"
+      - name: Install MSYS2 packages
+        run: |
+          pacman -S --noconfirm --needed base-devel mingw-w64-x86_64-gcc mingw-w64-x86_64-libevent mingw-w64-x86_64-openssl mingw-w64-x86_64-postgresql mingw-w64-x86_64-python mingw-w64-x86_64-python-pip autoconf automake libtool pkg-config zip
```
petere: should use `MINGW_PACKAGE_PREFIX`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+          windres pgbouncer.exe
+      - name: Build zip
+        run: make -j4 zip
+      - name: Upload zip
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: pgbouncer-windows-zip
+          path: pgbouncer-*.zip
+          if-no-files-found: ignore
```
petere: dubious, see earlier

.github/workflows/ci.yml
```diff
@@ -0,0 +1,401 @@
...
+  group: ${{ github.workflow }}-${{ github.ref }}
+  # Never cancel in-progress runs on master, so every commit is fully tested.
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
+
+env:
+  DEBIAN_FRONTEND: noninteractive
+  LANG: C
+
+jobs:
+  format-check:
```
petere: Is there a reason this job is listed first now?

README.md
```diff
@@ -1,6 +1,8 @@
 PgBouncer
 =========
 
+[![CI](https://github.com/pgbouncer/pgbouncer/actions/workflows/ci.yml/badge.svg)](https://github.com/pgbouncer/pgbouncer/actions/workflows/ci.yml)
```
petere: This looks like an unrelated change. (It doesn't replace the Cirrus badge.)

AFAICT, this would show the status of the master branch even in a different branch or a fork. That seems confusing.

petere (changes requested):
I have some comments on some lesser details, but mainly I didn't get the workflow_dispatch thing working, so I couldn't check how most of the jobs run.

JelteF:
Thanks for the review. I think I addressed all of it. I also did some more testing on my own fork and got all the builds green there. I'm going to merge this now, so we at least have something. Feel free to do follow up work/improvements. 

The problem with workflow dispatch not working is because the job did not exist on the master branch... So merging that solves it. Similarly PR events also don't work without the job existing in this repo. To make it easier I also added a "ci-full" label that you can add to a PR to run all the CI jobs.

PR #1514, opened 2026-06-09
mkauth.py: interpret "-" as stdout

dawagner:
If "-" is passed as first (DSTFN) argument, print the result on stdout instead of trying to update an existing file.

Interpreting "-" as stdout is a common convention. I'm using this patched version of mkauth.py in a deployment pipeline to read the hashes from a postgres cluster on a given host, store them in memory and then write them on another host.


JelteF (approved):

dawagner:
Thanks, @JelteF :)

PR #1516, opened 2026-06-18
Implement support for a bounded accept loop

achanda:
The current behavior accepts connections until the accept queue is empty. Under high connection load, that can monopolize the event loop. This new setting lets operators trade faster backlog draining for better event-loop fairness.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -1087,6 +1087,14 @@ Default: 10
 
 ## Low-level network settings
 
+### accept_batch
+
+Maximum number of client connections to accept from a single listening socket in
+one event loop iteration. 0 means no limit, so PgBouncer accepts connections
+until the accept queue is empty.
+
+Default: 0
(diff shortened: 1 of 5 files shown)


JelteF:
What do you intend to set this configuration to? It might make sense to actually change the default behavior so all users can benefit from a better default.

achanda:
I was initially thinking a value around 100 might make sense. But after running more experiments today, I’m no longer convinced this setting helps.

The test setup was: keep a steady background load of 100 existing connections repeatedly running `SELECT 1`, then inject a burst of 1k new client connections. I measured the tail latency of the background workload during the connection storm while varying `accept_batch` across `0`, `250`, `500`, and `1000`.

I did not see a consistent improvement for any nonzero value. That makes me suspect this setting may not matter in practice, at least for pgbench workload.

Issue #1530, opened 2026-07-20
Server connection can hang over TLS when a large ParameterStatus crosses the pkt_buf boundary


joncwong:
Hi all,

I think there's a lost-wakeup bug on the TLS read path: a connection can stall indefinitely when a packet the proto handler needs to see in full straddles the [`pkt_buf`](https://github.com/pgbouncer/pgbouncer/blob/13a344f2625381296fc02e29b986a11be9c6b983/src/main.c#L316) boundary, because decrypted bytes left sitting inside OpenSSL never get drained before PgBouncer goes back to waiting on the socket.

Conditions to hit it:

- TLS-based connection
- a query response larger than `pkt_buf` (default 4096) that arrives as a single TLS record
- a `ParameterStatus` message ends up split across that buffer boundary

That `ParameterStatus` typically comes from a `SET application_name`. The server echoes the value back at a size proportional to it, so a longish value pushes the message past 64 bytes. pgbouncer's handler assumes this won't happen: [`load_parameter()`](https://github.com/pgbouncer/pgbouncer/blob/13a344f2625381296fc02e29b986a11be9c6b983/src/server.c#L35-L38) notes that `SBUF_SMALL_PKT` "must be larger than max param pkt".

Mechanism, as far as I can tell:
[...]

PR #1531, opened 2026-07-20
Fix stalled TLS connection when a packet straddles pkt_buf

joncwong:
## Overview
Addresses https://github.com/pgbouncer/pgbouncer/issues/1530.

When a packet the proto handler needs to read whole straddles the pkt_buf boundary on a TLS connection, sbuf_process_pending() takes the need_more_data path and sbuf_main_loop() returns to wait for a read event. But SSL_read() has already pulled the entire TLS record out of the kernel into OpenSSL's own buffer, so the rest of that packet is decrypted and sitting there while the kernel socket is empty. With level-triggered polling the read event never fires again, and the connection stalls until unrelated traffic happens to wake the socket.

This is reachable via ParameterStatus: a SET application_name with a long value produces a message larger than SBUF_SMALL_PKT, which load_parameter() must see in full.

Add an sbufio_pending op (SSL_pending() for TLS, 0 for raw sockets) and, on the need_more_data path, re-read instead of waiting when the TLS layer still has buffered decrypted data. No change for non-TLS sockets.

## Testing 
I couldn't fit this into the existing `test_ssl.py` suite. Those tests run pgbouncer against a real Postgres, and the bug depends on TLS record framing: Postgres over loopback flushes the `ParameterStatus` in its own TLS record, so it never straddles the `pkt_buf` boundary and the stall doesn't happen. I believe triggering it deterministically needs something that emits the response as a single TLS record like a mock TLS server in place of Postgres, or a re-framing TLS proxy in front of it.

I verified it locally with a small mock TLS server (a crafted response with an oversized `ParameterStatus` sitting across the boundary). It successfully reproduces it on master and passes with this change. Happy to contribute that as a test case if desired.

### DIFF
--- a/include/sbuf.h
+++ b/include/sbuf.h
@@ -60,6 +60,8 @@ struct SBufIO {
 	ssize_t (*sbufio_recv)(SBuf *sbuf, void *buf, size_t len);
 	ssize_t (*sbufio_send)(SBuf *sbuf, const void *data, size_t len);
 	int (*sbufio_close)(SBuf *sbuf);
+	/* buffered data the poller cannot see (e.g. decrypted by TLS) */
+	size_t (*sbufio_pending)(SBuf *sbuf);
 };
 
 /*
@@ -168,6 +170,11 @@ static inline ssize_t sbuf_op_recv(SBuf *sbuf, void *buf, size_t len)
 static inline ssize_t sbuf_op_send(SBuf *sbuf, const void *buf, size_t len)
 {
 	return sbuf->ops->sbufio_send(sbuf, buf, len);
+}
+
+static inline size_t sbuf_op_pending(SBuf *sbuf)
+{
+	return sbuf->ops->sbufio_pending(sbuf);
 }
 
 static inline int sbuf_op_close(SBuf *sbuf)
--- a/lib/usual/tls/tls.c
+++ b/lib/usual/tls/tls.c
@@ -695,6 +695,13 @@ out:
 	return (rv);
 }
 
+size_t tls_pending(struct tls *ctx)
+{
+	if (ctx->ssl_conn == NULL)
+		return 0;
+	return SSL_pending(ctx->ssl_conn);
+}
+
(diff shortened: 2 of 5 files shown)


decatimz (approved):

KIRTIKUMARKK21 (approved):

AdzerKI:
Field report in support of this patch, since it has had no feedback for a month: we hit the stall it fixes in a real deployment and have been running this patch since mid-August.

**Setup.** CloudNativePG's PgBouncer pooler in transaction mode, `max_prepared_statements = 100`, TLS between pooler and Postgres, JDBC clients (Hibernate) using server-side prepared statements. PgBouncer 1.25.1, now 1.25.2.

**Symptom.** During bulk catalog writes a server connection stalls permanently: the server waits for a client command, the client waits for a response, and the socket queues on both sides are empty. It never recovers on its own — the connection is lost for the lifetime of the pooler process.

**Why this patch matters for that.** With a non-zero prepared-statement limit the pooler parses extended-protocol packets in full, so a packet that arrives split across reads has to be reassembled. When TLS holds the remainder in its own buffer, the socket read event does not fire again, so the remainder is never collected and the reassembly never completes. Draining TLS pending data — what this patch does — is what makes the wakeup happen.

We run it together with #1532; the two address the two halves of the same stall (being woken to read the rest, and reassembling a named Parse that arrived incomplete). With both applied the workload that reproduced the stall for us has run without a single occurrence since.

Alternatives we tried before patching, in case they are useful as evidence that this is not a configuration problem:

- turning off server-side prepared statements in the driver — loses the plan cache, and the pooler's prepared-statement limit still has to be non-zero for the drivers we use;
- a larger packet buffer — only moves the boundary at which the split happens;
- plaintext between pooler and Postgres — sidesteps the TLS half, but puts the data path outside the encrypted namespace, which is not acceptable for us.

Happy to supply more detail from our side if it helps move this along.

JelteF:
Thank you! I got claude to create a test for it.

JelteF (approved):

PR #1532, opened 2026-07-21
fix: PS stall — always use CB_WANT_COMPLETE_PACKET for incomplete named Parse

KIRTIKUMARKK21:
## Problem

When transaction pooling is enabled with `max_prepared_statements > 0`, a client that pipelines `Parse/Bind/Execute/Sync` bursts can permanently stall a server backend in `active/ClientRead` state.

**Trigger condition:** A named Parse packet whose size fits within `pkt_buf` but arrives incomplete (i.e. `incomplete_pkt(pkt)` is true but `pkt->len <= cf_sbuf_len`). This is common on any connection where TCP segments don't align with message boundaries — TLS, high-latency links, large prepared statement texts.

### Stall mechanism

1. `inspect_parse_packet()` fires on the **partial** packet. The statement name sits in the first few bytes, so inspection succeeds and the packet is identified as a named Parse.
2. The previous code returned `false` (inline retry) since `pkt->len <= cf_sbuf_len`, telling sbuf to retry once more data arrives.
3. On retry, `handle_parse_command()` runs on the now-complete packet and writes `RA_FAKE` into `outstanding_requests`.
4. The transaction ends; the server backend is released to the pool. The next `Bind` in the pipeline is assigned to a **different** backend (transaction pooling).
5. `ensure_statement_is_prepared_on_server()` injects `RA_SKIP(Parse)` on the new backend **behind** the `RA_FAKE` written in step 3, producing queue: `[RA_FAKE, RA_SKIP, …]`.
6. `pop_outstanding_request()` refuses to pop an `RA_FAKE` head. `RA_SKIP` is never consumed. The `Sync` response is never forwarded. The backend is permanently pinned with `outstanding_requests > 0` — a deadlock until the process is killed.

Note: the bug is independent of `pkt_buf` size — it triggers whenever a named Parse fits *within* the buffer but arrives split across reads. Counterintuitively, setting `pkt_buf` *smaller* than the Parse packet avoids it by forcing the `CB_WANT_COMPLETE_PACKET` path, but this is not a real fix.

## Fix

Remove the `pkt->len > cf_sbuf_len` branch and unconditionally use the `CB_WANT_COMPLETE_PACKET` dynamic reassembly path for all `PS_HANDLE_FULL_PACKET` incomplete packets.

This ensures the entire packet is buffered before any queue entry is written, so `inspect_parse_packet()` and `handle_parse_command()` always operate on a consistent, complete packet. The `RA_FAKE`/`RA_SKIP` ordering corruption cannot occur.

The `CB_WANT_COMPLETE_PACKET` path is already used for packets larger than `pkt_buf` with no issues; this change extends it to the remaining case.

## Related

PR #1531 (`joncwong:fix-tls-parameterstatus-stall`) addresses a separate but related stall on TLS connections where `SSL_pending` data is not visible to the event poller. Both bugs involve the `need_more_data` / incomplete packet path and can coexist.

### DIFF
--- a/src/client.c
+++ b/src/client.c
@@ -1545,26 +1545,31 @@ static bool handle_client_work(PgSocket *client, PktHdr *pkt)
 	}
 
 	if (ps_action == PS_HANDLE_FULL_PACKET && incomplete_pkt(pkt)) {
-		if (pkt->len > (unsigned) cf_sbuf_len) {
-			/*
-			 * We need to handle the complete packet, but it is too
-			 * large to fit into our sbuf buffer size (determined
-			 * by the pkt_buf config). So now we need to fetch the
-			 * whole packet using our dynamically sized packet
-			 * buffering logic.
-			 */
-			client->packet_cb_state.flag = CB_WANT_COMPLETE_PACKET;
-			sbuf_prepare_fetch(sbuf, pkt->len);
-			return true;
-		} else {
-			/*
-			 * We need to handle the complete packet, but it fits
-			 * in our sbuf buffer, so we can simply return false to
-			 * indicate to sbuf to retry once it has received more
-			 * data
-			 */
-			return false;
-		}
+		/*
+		 * Always use the dynamic reassembly path for named prepared
+		 * statement packets that haven't been fully received yet,
+		 * regardless of whether they fit in pkt_buf.
+		 *
+		 * The previous code returned false (inline retry) when
+		 * pkt->len <= cf_sbuf_len, relying on sbuf to call us again
+		 * once the full packet arrived. However, this causes a
+		 * prepared-statement accounting corruption: inspect_parse_packet()
+		 * succeeds on the partial packet (statement name fits in the
+		 * initial bytes), then the inline retry fires handle_parse_command()
+		 * which adds RA_FAKE to outstanding_requests. When the next Bind
+		 * in the pipeline arrives on a freshly-assigned server backend,
+		 * handle_bind_command() injects RA_SKIP(Parse) behind that RA_FAKE.
+		 * pop_outstanding_request() refuses to pop an RA_FAKE head, so
+		 * RA_SKIP is never consumed, leaving the server permanently pinned
+		 * with outstanding_requests > 0 (active/ClientRead stall).
+		 *
+		 * The reassembly path (CB_WANT_COMPLETE_PACKET) buffers the entire
+		 * packet before any queue entries are written, preventing the
+		 * RA_FAKE/RA_SKIP ordering corruption entirely.
+		 */
+		client->packet_cb_state.flag = CB_WANT_COMPLETE_PACKET;
+		sbuf_prepare_fetch(sbuf, pkt->len);
+		return true;
 	}
 
 	if (ps_action == PS_INSPECT_FAILED) {


JelteF:
It sounds like it should be possible to create a regression test for this

JelteF:
I think this is solving the problem you're describing in the wrong way. The fact that this RA_FAKE message survives so long sounds like it's the bug that needs fixing instead. I don't fully understand how that happens though, because you're leaving out some key details here:

> 4. The transaction ends; the server backend is released to the pool. The next Bind in the pipeline is assigned to a different backend (transaction pooling).

How does the transaction end here? Using a `Sync` message? Because a `Sync` message should consume the RA_FAKE entry.

AdzerKI:
We run into this stall in production and carry this patch, so let me offer what I can toward @JelteF's two points — the reproduction he asked for, and the `RA_FAKE` lifetime question. Reading 1.25.2:

**What consumes a fake entry.** Not a client `Sync` directly. Fake entries are drained by the loop at the end of *server response* handling in `server.c`, which pops leading `RA_FAKE` entries and calls `queue_fake_response()` for each. A `Sync` only consumes them transitively: the server answers `ReadyForQuery`, that response is processed, and the loop runs. `add_outstanding_request()` short-circuits and answers a fake immediately **only** when the queue is empty at that moment; otherwise the entry is appended and waits for that drain. So a fake survives exactly as long as no further server response is processed — which is what happens once the server has nothing left to answer.

**This seems to contradict step 4 of the description.** `release_server()` refuses to release a server whose `outstanding_requests` is non-empty — it calls `disconnect_server(server, true, "client disconnected with queries in progress")` instead, and the client-disconnect path does the same with "client disconnected with query in progress". So a queued `RA_FAKE` should not be able to ride a pooled backend to a different client through either path; the backend would be closed. Whatever keeps the entry alive is therefore not "released to the pool with the entry queued", and narrowing that down looks like the thing that would settle whether the fix belongs here or at the point where the entry is created.

**Reproduction shape**, for the regression test:

- transaction pooling, `max_prepared_statements` non-zero;
- client pipelines `Parse`(named)/`Bind`/`Execute`/`Sync` bursts;
- the named `Parse` packet must arrive **split across reads** while still fitting in `pkt_buf` — i.e. `incomplete_pkt(pkt)` true with `pkt->len <= cf_sbuf_len`.

The split is what a test has to force deliberately; in the field TLS provides it for free, which is why we see it constantly and a plaintext test setup does not. Ours is CloudNativePG's PgBouncer with TLS to Postgres and JDBC clients using server-side prepared statements; bulk catalog writes reproduce it reliably. The observable end state is a server backend stuck in `active`/`ClientRead` with both socket queues empty and no recovery.

For completeness: we apply this together with #1531, since with TLS the remainder of the split packet also needs the read wakeup that one restores — either alone was not enough for us.

JelteF:
I think the underlying issue probably got fixed at the right place by #1555 

Can you check if this change is still needed for you with that PR applied.

AdzerKI:
@JelteF #1555 matches what we saw, and it answers the question I was stuck on above.

I could not explain how a fake entry survives long enough to reach another client, because both release paths refuse a server with a non-empty `outstanding_requests`. #1555's first fault is the missing piece: no release is involved. A skipped response leaves `RA_FAKE` at the head because fakes were flushed only after forwarded packets, so the queue is misaligned in place for the rest of that server connection — and the poisoned connection is what later reaches another client. That is the same symptom we carry the patch for, fixed at the layer where it belongs.

I will rebuild from master, re-run the workload that produced the stall for us, and report the result here. If it holds, we drop our build and this PR is not needed on our account.

AdzerKI:
Rebuilt the pooler from master at c0d4162 (so #1555 and #1531 are in, this patch is not) and re-ran the workload that stalled for us: bulk catalog writes in transaction mode with max_prepared_statements=100 — an external API catalog sync of one partner, a larger one of another, and a full rebuild of the stand catalog, 51 listings deleted and recreated end to end with snapshot republication across eight services.

No stall. cl_waiting and maxwait stayed at zero and no backend sat in idle in transaction; before, one item would leave a transaction open with the server waiting on ClientRead and the next item blocked on the same key until the role's transaction timeout.

So #1555 covers it for us — this PR is not needed on our account and we have dropped the patch from our build.

JelteF:
Thank you for checking. Closing this then.

PR #1534, opened 2026-07-24
Add target_session_attrs support

cosgroveb:
Replaces #1179.

pgbouncer stores `in_hot_standby` and `default_transaction_read_only` on each server connection. It checks them after authentication and after running the configured `connect_query`, before making the connection available to clients. This separation of backend selection from client session state is the big improvement over #1179.

`default_transaction_read_only` also has a client-side role. #1573 adds it to `VarCache` by default so pgbouncer can replay the client's value onto another backend.

`prefer-standby` is the only unsupported libpq mode.

PostgreSQL 14 began reporting `in_hot_standby` and `default_transaction_read_only` in `ParameterStatus`. On older servers, only `any` matches.

Backend rejection during startup requires cleanup around `disconnect_server()`. pgbouncer keeps a linked replication client waiting for another host, discards the rejected backend's startup parameters, and publishes parameters only after accepting a backend. It preserves an earlier connection error for client reporting.

Rejecting a backend counts as a failed connection attempt for host selection. With multiple hosts and `load_balance_hosts=disable`, pgbouncer tries the next host in configured order. If no candidate matches, later clients receive the cached failure during `server_login_retry`.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -1399,6 +1399,41 @@ fast retries when multiple hosts are available.
 
 Default: `round-robin`
 
+### target_session_attrs
+
+Controls which PostgreSQL servers are accepted for new server connections.
+PgBouncer checks status values reported by PostgreSQL after authentication and
+the configured `connect_query`, before making the connection available to
+clients.
+
+any
+:   Accept any server. This is the default.
+
+read-write
+:   Accept a server only when both `in_hot_standby` and
+    `default_transaction_read_only` are `off`.
+
+read-only
+:   Accept a server when either `in_hot_standby` or
+    `default_transaction_read_only` is `on`.
+
+primary
+:   Accept a server only when `in_hot_standby` is `off`. A primary whose
+    `default_transaction_read_only` setting is `on` still matches `primary`.
+
+standby
+:   Accept a server only when `in_hot_standby` is `on`.
+
+Values other than `any` require the status parameters reported by PostgreSQL
+14 and later. If PostgreSQL does not report enough information to evaluate the
+requested value, PgBouncer rejects the server connection. `prefer-standby` is
+not supported.
+
+The check occurs when a server connection is established. PgBouncer does not
+re-evaluate an admitted connection if its reported status later changes.
+
+Default: `any`
+
 ### max_db_connections
 
 Configure a database-wide maximum of server connections (i.e. all pools within
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,9 +85,9 @@ filterwarnings = [
 extend-exclude = ["lib"]
 
 [tool.ruff.lint]
-# Ruff's default rules (a subset of pycodestyle E + pyflakes F); "I" adds
-# import sorting, replacing isort.
-extend-select = ["I"]
+# Keep the rule set explicit so Ruff releases cannot change lint scope. E4, E7,
+# E9, and F were Ruff's default before 0.16; I replaces isort.
+select = ["E4", "E7", "E9", "F", "I"]
 # F401 (unused import) and F541 (f-string without placeholders) were ignored
 # under the old flake8 config and the test suite still relies on that.
 ignore = ["F401", "F541"]
--- a/src/objects.c
+++ b/src/objects.c
@@ -2426,6 +2426,8 @@ bool use_server_socket(int fd, PgAddr *addr,
 	server->suspended = true;
 	server->pool = pool;
 	server->login_user_credentials = credentials;
+	/* Takeover does not transfer target-session observations. */
+	server->close_needed = db->target_session_attrs != TARGET_SESSION_ANY;
 	server->connect_time = server->request_time = get_cached_time();
 	server->query_start = 0;
 	statlist_init(&server->canceling_clients, "canceling_clients");
--- a/src/server.c
+++ b/src/server.c
@@ -27,6 +27,15 @@
 
 #define ERRCODE_CANNOT_CONNECT_NOW "57P03"
 
+static enum TargetSessionAttrValue parse_target_session_attr(const char *value)
+{
+	if (strcmp(value, "off") == 0)
+		return TARGET_SESSION_ATTR_OFF;
+	if (strcmp(value, "on") == 0)
+		return TARGET_SESSION_ATTR_ON;
+	return TARGET_SESSION_ATTR_UNKNOWN;
+}
+
 static bool load_parameter(PgSocket *server, PktHdr *pkt, bool startup)
 {
 	const char *key, *val;
@@ -44,6 +53,10 @@ static bool load_parameter(PgSocket *server, PktHdr *pkt, bool startup)
 	if (!mbuf_get_string(&pkt->data, &val))
 		goto failed;
 	slog_debug(server, "S: param: %s = %s", key, val);
+	if (strcmp(key, "in_hot_standby") == 0)
+		server->in_hot_standby = parse_target_session_attr(val);
+	else if (strcmp(key, "default_transaction_read_only") == 0)
+		server->default_transaction_read_only = parse_target_session_attr(val);
 
 	varcache_set(&server->vars, key, val);
 
--- a/src/server.c
+++ b/src/server.c
@@ -60,7 +60,7 @@ static bool load_parameter(PgSocket *server, PktHdr *pkt, bool startup)
 
 	varcache_set(&server->vars, key, val);
 
-	if (client) {
+	if (client && !(startup && server->replication)) {
 		slog_debug(client, "setting client var: %s='%s'", key, val);
 		varcache_set(&client->vars, key, val);
 	}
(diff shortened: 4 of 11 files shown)


cosgroveb:
I'm going to take a stab at a fresh re-implementation of #1205 stacked onto this to show that this can be a kind of [pre-factoring](https://testing.googleblog.com/2026/07/prefactoring-clear-way-for-your-new.html).

saolof:
I like the fact that this rewrite basically just does what I would expect it to do before looking at the code and that it is simpler.

This does assume that your promotion strategy on the postgres HA side fences the old primary and kicks out the backend connections to force them to reconnect right? That's quite common but just needs to be documented properly, which seems to be the case here.

cosgroveb:
@saolof correct. My HA use-case does assume fencing on the primary side. Previously in my environment that was managed by Pacemaker and now Aurora. I agree that the pgbouncer pieces are documented accordingly. 

It would be nice, perhaps, for there to exist a recipe for the HA use-case but that doesn't belong to pgbouncer proper.

cosgroveb:
@saolof @ya-mouse @JelteF [this is a draft](https://github.com/cosgroveb/pgbouncer/pull/4) of what an alternate version of #1205 could look like if this PR is the direction we choose for e.g. `target-session-attrs`: https://github.com/cosgroveb/pgbouncer/pull/4

Edit: view https://github.com/cosgroveb/pgbouncer/pull/4 instead.

src/server.c
```diff
@@ -27,6 +27,15 @@
 
 #define ERRCODE_CANNOT_CONNECT_NOW "57P03"
 
+static enum TargetSessionAttrValue parse_target_session_attr(const char *value)
```
JelteF: Let's make this more generic and make it `parse_bool_option`. Similar for the enum name, and the enum variants.

test/test.ini
```diff
@@ -44,6 +44,26 @@ hostlist_good_first = port=6666 host=127.0.0.1,127.0.0.3 dbname=p0 user=bouncer
 hostlist_bad_first = port=6666 host=unresolvable-hostname,127.0.0.1 dbname=p0 user=bouncer load_balance_hosts=disable
 load_balance_hosts_update = port=6666 host=127.0.0.1,127.0.0.3 dbname=p0 user=bouncer load_balance_hosts=disable
 
+; for testing target_session_attrs
```
JelteF: This `test.ini` file has become pretty unwieldy. I think adding 1-2 items per new feature to it probably still make sense. I think either we should use `bouncer.run_with_config(...)` for these tests or create a dedicated ini file for these (and start from that ini file instead). Or ofcourse if most of these new entries are unnecessary or could be combined that's also an option.

src/server.c
cosgroveb: good call fd186a199e3b2a77424b2f61a8ea0699a3f028c1

test/test.ini
cosgroveb: Yup. I think [this is cleaner](https://github.com/pgbouncer/pgbouncer/pull/1534/changes/6b12271927dc1c9b12832d28470e87c3bdf513e0). Happy to adjust as-needed.

hlinnaka:
I think looking at `default_transaction_read_only` is based on some kind of confusion. PostgreSQL doesn't change `default_transaction_read_only` based on whether the server can do writes or not. The client can change it at any time, even on a standby server.  So it seems pointless to make decisions on which server to connect to based on that. Now `transaction_read_only` is a different story, maybe the idea was to do something with that?

doc/config.md
```diff
@@ -1399,6 +1399,41 @@ fast retries when multiple hosts are available.
...
+:   Accept a server only when `in_hot_standby` is `off`. A primary whose
+    `default_transaction_read_only` setting is `on` still matches `primary`.
+
+standby
+:   Accept a server only when `in_hot_standby` is `on`.
+
+Values other than `any` require the status parameters reported by PostgreSQL
+14 and later. If PostgreSQL does not report enough information to evaluate the
+requested value, PgBouncer rejects the server connection. `prefer-standby` is
+not supported.
```
hlinnaka, line 1429: This is the only reference to `prefer-standby`. Some kind of leftover?

pyproject.toml
```diff
@@ -85,9 +85,9 @@ filterwarnings = [
 extend-exclude = ["lib"]
 
 [tool.ruff.lint]
-# Ruff's default rules (a subset of pycodestyle E + pyflakes F); "I" adds
-# import sorting, replacing isort.
-extend-select = ["I"]
+# Keep the rule set explicit so Ruff releases cannot change lint scope. E4, E7,
+# E9, and F were Ruff's default before 0.16; I replaces isort.
+select = ["E4", "E7", "E9", "F", "I"]
```
JelteF: This shouldn't be needed anymore after merging with main

JelteF:
> So it seems pointless to make decisions on which server to connect to based on that. Now `transaction_read_only` is a different story, maybe the idea was to do something with that?

I kindof agree with you in theory, but this feature mirrors a libpq feature that uses `default_transaction_read_only` this way: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS

Also `transaction_read_only` isn't `GUC_REPORT` so we cannot use it here.

pyproject.toml
cosgroveb: master still uses extend-select = ["I"](https://github.com/pgbouncer/pgbouncer/blob/a341bf2c9e6c03895e58ae226e7f8c9d0ac61b55/pyproject.toml#L87-L93), so I think this is still needed. Can you double check?

doc/config.md
cosgroveb: Nope. This  documentation was requested in the earlier PR's review (https://github.com/pgbouncer/pgbouncer/pull/1179#discussion_r1881001502)

cosgroveb:
> > So it seems pointless to make decisions on which server to connect to based on that. Now `transaction_read_only` is a different story, maybe the idea was to do something with that?
> 
> I kindof agree with you in theory, but this feature mirrors a libpq feature that uses `default_transaction_read_only` this way: https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNECT-TARGET-SESSION-ATTRS
> 
> Also `transaction_read_only` isn't `GUC_REPORT` so we cannot use it here.

Yeah @hlinnaka `default_transaction_read_only` tracked in the `VarCache` because clients can set it, while `target_session_attrs` reads a separate per-server value from `ParameterStatus`, following the earlier discussions in [1179](https://github.com/pgbouncer/pgbouncer/pull/1179#discussion_r1881016656) and [1205](https://github.com/pgbouncer/pgbouncer/pull/1205#discussion_r1881036221).

pyproject.toml
JelteF: The linting issues were addressed by: https://github.com/pgbouncer/pgbouncer/pull/1538

doc/config.md
hlinnaka: A-ha, I missed that this is modeled after libpq's `target_session_attrs` option! It would be good to mention that explicitly, the "`prefer-standby` is not supported" sentence makes much more sense when you're comparing with libpq.

cosgroveb: good point: 84568cc

pyproject.toml
cosgroveb: OK cool dropping b9d287d from the PR

test/test.ini
```diff
@@ -95,5 +96,5 @@ tcp_socket_buffer = 4096
 
 ;track_extra_parameters is case sensitive.
 ;bgbouncer will only track the parameters with GUC_REPORT flag set in Postgres
-track_extra_parameters = search_path, intervalstyle
```
JelteF, line 98: This isn't needed anymore

test/test.ini
```diff
@@ -44,6 +44,7 @@ hostlist_good_first = port=6666 host=127.0.0.1,127.0.0.3 dbname=p0 user=bouncer
 hostlist_bad_first = port=6666 host=unresolvable-hostname,127.0.0.1 dbname=p0 user=bouncer load_balance_hosts=disable
 load_balance_hosts_update = port=6666 host=127.0.0.1,127.0.0.3 dbname=p0 user=bouncer load_balance_hosts=disable
 
+
```
JelteF: unnecessary addition

src/objects.c
```diff
@@ -2427,6 +2427,8 @@ bool use_server_socket(int fd, PgAddr *addr,
 	server->suspended = true;
 	server->pool = pool;
 	server->login_user_credentials = credentials;
+	/* Takeover does not transfer target-session observations. */
+	server->close_needed = db->target_session_attrs != TARGET_SESSION_ANY;
 	server->connect_time = server->request_time = get_cached_time();
```
JelteF: this now causes a trivial merge conflict because I removed all this code in: https://github.com/pgbouncer/pgbouncer/pull/1581

src/server.c
```diff
@@ -44,10 +53,14 @@ static bool load_parameter(PgSocket *server, PktHdr *pkt, bool startup)
...
 	slog_debug(server, "S: param: %s = %s", key, val);
+	if (strcmp(key, "in_hot_standby") == 0)
+		server->in_hot_standby = parse_bool_option(val);
+	else if (strcmp(key, "default_transaction_read_only") == 0)
+		server->default_transaction_read_only = parse_bool_option(val);
 
 	varcache_set(&server->vars, key, val);
 
-	if (client) {
+	if (client && !(startup && server->replication)) {
```
JelteF, line 64: This could use a code comment. what are we doing here?

cosgroveb: Replication clients are linked before their backend is accepted, so this defers copying startup values into the client until after `target_session_attrs` accepts that backend.

42bd6b11833d1b99bfddbcb711998d4e06754ef8

test/test.ini
cosgroveb: 53759a3

test/test.ini
cosgroveb: f1db804

src/objects.c
cosgroveb: Rebased and cleaned up dead fixtures 0cdf6ada622281af91f18e5a4f03c9b036d45a16

test/test.ini
cosgroveb: i mean 9c2e49cdb8126f91444555115eab6248af0e0f61

cosgroveb:
Pushed a rebase to aid in reviewability

PR #1543, opened 2026-08-13
Keep prepared statements registered after a pipelined reset

### LINKED ISSUE #1535: Pipeline `DEALLOCATE ALL` removes a later named prepared-statement mapping
**PgBouncer version**

`v1.25.2` (commit `13a344f`)

**Description**

With `max_prepared_statements > 0`, PgBouncer can remove a named prepared statement mapping created later in the same pipeline segment as `DEALLOCATE ALL`. The same happens with `DISCARD ALL`. The issue affects `transaction` and `statement` pool modes.

When the later named `Parse` is read, `handle_parse_command()` registers the client-visible name immediately [`src/prepare.c`](https://github.com/pgbouncer/pgbouncer/blob/13a344f2625381296fc02e29b986a11be9c6b983/src/prepare.c#L340-L344):

```c
client_ps = create_client_prepared_statement(pp.name, ps);
HASH_ADD_STR(client->client_prepared_statements, stmt_name, client_ps);
```

The reset is cleaned up later, when the backend sends `CommandComplete`. The following branch matches the reset tag and clears both caches [`src/server.c`](https://github.com/pgbouncer/pgbouncer/blob/13a344f2625381296fc02e29b986a11be9c6b983/src/server.c#L500-L519):

```c
case PqMsg_CommandComplete:
	/* unrelated COPY handling omitted */
	if (is_prepared_statements_enabled(server)
	    && (pkt->len == 1 + 4 + 15 || pkt->len == 1 + 4 + 12)) {
		const char *tag;
		if (mbuf_get_string(&pkt->data, &tag)) {
			if (strcmp(tag, "DEALLOCATE ALL") == 0 ||
			    strcmp(tag, "DISCARD ALL") == 0) {
				free_server_prepared_statements(server);
				if (client)
					free_client_prepared_statements(client);
			}
		} else {
			return false;
		}
	}
	pop_outstanding_request(server, (char[]) {PqMsg_Execute, '\0'}, &ignore_packet);
	break;
```

Because pipeline input can deliver the later `Parse` before the reset's `CommandComplete`, the cleanup removes the newly registered mapping even though the prepared statement still exists on the PostgreSQL backend.

**Reproduce:**

1. Start PostgreSQL 16 and PgBouncer in `transaction` or `statement` pool mode.
2. Run the following with the direct and PgBouncer DSNs supplied by the environment:

```python
import psycopg


def run(label: str, dsn: str, reset_sql: str = "DEALLOCATE ALL") -> None:
    try:
        with psycopg.connect(dsn, autocommit=True) as conn:
            with conn.pipeline() as pipeline:
                reset = conn.cursor()
                cur = conn.cursor()

                reset.execute(reset_sql)
                cur.execute("SELECT 1", prepare=True)
                pipeline.sync()
                print(f"{label}: first={cur.fetchall()!r}")

                cur.execute("SELECT 1", prepare=True)
                pipeline.sync()
                print(f"{label}: second={cur.fetchall()!r}")
    except Exception as exc:
        print(f"{label}: {type(exc).__name__}: {exc}")


run("direct", DIRECT_DSN)
run("pgbouncer", PGBOUNCER_DSN)
```

3. Observe the direct run passes; the PgBouncer run fails at the second `pipeline.sync()`.

4. Replace `DEALLOCATE ALL` with `DISCARD ALL` to reproduce the same result.

**Expected behavior:** The second execution should return `1` and the connection should remain usable, as it does with direct PostgreSQL.

**Actual behavior:** Direct prints `first=[(1,)]` and `second=[(1,)]`. PgBouncer prints `first=[(1,)]`; the second `pipeline.sync()` raises `OperationalError` after PgBouncer closes the frontend connection.

**Proposed fix:** Do not let reset cleanup remove client mappings created after the reset command in the same pipeline segment. The client mapping and backend prepared-statement cache should remain consistent after cleanup.

m0g3r:
Fixes #1535.

## Problem

With `max_prepared_statements > 0` in transaction or statement pooling, PgBouncer registers a
client-visible prepared statement name eagerly, when it reads the `Parse` message
(`handle_parse_command()` in `src/prepare.c`). Cleanup for `DEALLOCATE ALL` and `DISCARD ALL`
happens late, when the backend's `CommandComplete` for the reset arrives (`src/server.c`), and it
cleared both statement caches wholesale.

Pipelined input can deliver the later `Parse` before that `CommandComplete` comes back. The
blanket cleanup then removed a mapping for a statement the server really had prepared, because the
server executes the reset first and the `Parse` afterwards. The client received a failure for a
statement it had just prepared successfully, and the connection was closed.

## Solution

Give both statement caches an ordering, and let the reset remove only what predates it.

- A per-socket `prepared_statement_seq` counter stamps each client and server statement as it is
  registered.
- `OutstandingRequest` records both counters when the request is queued.
- Handling the reset's `CommandComplete` frees only statements stamped at or before the values
  recorded on that request, via the new `free_client_prepared_statements_upto()` and
  `free_server_prepared_statements_upto()`.

Anything prepared later in the same pipeline keeps its mapping, which matches what actually exists
on the server.

Both caches needed this. Fixing only the client side moved the failure from a closed connection to
`prepared statement "PGBOUNCER_1" already exists`: the server cache had also been cleared, so
PgBouncer re-sent a `Parse` for a statement that was still present on the backend.

## Testing

`test_reset_all_does_not_remove_later_prepared_statement` is added to `test/test_prepared.py`,
parametrised over `DEALLOCATE ALL` and `DISCARD ALL`. It issues the reset and a named `Parse` in
one pipeline segment, then executes the statement. On master both parameters fail with
`psycopg.OperationalError: consuming input failed: server closed the connection unexpectedly`;
with this change both pass.

- `pytest test/test_prepared.py` — 27 passed, 1 skipped. This includes the existing
  `test_deallocate_all` and `test_discard_all`, which assert that a reset *does* invalidate
  statements prepared before it. Those still pass, so the reset keeps its intended effect.
- Wider run — 214 passed.
- `dev/format.sh check` and `dev/format.sh lint` — both clean.

I could not run `test/test_ssl.py` or `test/test_load_balance_hosts.py` locally: they fail on this
machine for want of TLS fixtures, and they fail identically on an unmodified checkout of `master`
(17 failed there versus 8 passed, the same set). Nothing in this change touches code those tests
reach, but I would rather say so than imply a clean full-suite run.

## AI assistance

Claude Code was used to reproduce the issue, write the change and the test, and run the
verification above. The complete diff was reviewed before submission.

### DIFF
--- a/src/prepare.c
+++ b/src/prepare.c
@@ -210,6 +210,7 @@ void unregister_prepared_statement(PgSocket *server, uint64_t query_id)
  */
 bool add_prepared_statement(PgSocket *server, PgServerPreparedStatement *server_ps)
 {
+	server_ps->seq = ++server->prepared_statement_seq;
 	HASH_ADD_UINT64(server->server_prepared_statements, query_id, server_ps);
 	if (uthash_alloc_failed) {
 		uthash_alloc_failed = false;
@@ -341,6 +342,7 @@ bool handle_parse_command(PgSocket *client, PktHdr *pkt)
 	client_ps = create_client_prepared_statement(pp.name, ps);
 	if (client_ps == NULL)
 		goto oom;
+	client_ps->seq = ++client->prepared_statement_seq;
 	HASH_ADD_STR(client->client_prepared_statements, stmt_name, client_ps);
 	if (uthash_alloc_failed) {
 		uthash_alloc_failed = false;
@@ -693,6 +695,32 @@ bool handle_close_statement_command(PgSocket *client, PktHdr *pkt, PgClosePacket
 }
 
 /*
+ * Frees the prepared statements that the client registered at or before the
+ * given sequence number, leaving any it registered later in place.
+ *
+ * A DEALLOCATE ALL or DISCARD ALL is cleaned up when the server reports its
+ * CommandComplete, but pipelined input means the client can already have sent
+ * a Parse for a new statement by then. That statement really does exist on the
+ * server, so dropping its mapping would make PgBouncer reject a statement the
+ * client just prepared successfully.
+ */
+void free_client_prepared_statements_upto(PgSocket *client, uint64_t seq)
+{
+	PgClientPreparedStatement *client_ps, *tmp;
+
+	HASH_ITER(hh, client->client_prepared_statements, client_ps, tmp) {
+		if (client_ps->seq > seq)
+			continue;
+		HASH_DEL(client->client_prepared_statements, client_ps);
+		if (--client_ps->ps->use_count == 0) {
+			HASH_DEL(prepared_statements, client_ps->ps);
+			free(client_ps->ps);
+		}
+		free(client_ps);
+	}
+}
+
+/*
  * Frees all the prepared statements that are cached on the client.
  */
 void free_client_prepared_statements(PgSocket *client)
@@ -710,6 +738,23 @@ void free_client_prepared_statements(PgSocket *client)
 
 	free(client->client_prepared_statements);
 	client->client_prepared_statements = NULL;
+}
+
+/*
+ * Frees the statements prepared on the server at or before the given sequence
+ * number. A reset the server has just run only removes what it had prepared
+ * beforehand; anything Parsed later in the same pipeline still exists there.
+ */
+void free_server_prepared_statements_upto(PgSocket *server, uint64_t seq)
+{
+	struct PgServerPreparedStatement *current, *tmp_s;
+
+	HASH_ITER(hh, server->server_prepared_statements, current, tmp_s) {
+		if (current->seq > seq)
+			continue;
+		HASH_DEL(server->server_prepared_statements, current);
+		free_server_prepared_statement(current);
+	}
 }
 
 /*
--- a/include/bouncer.h
+++ b/include/bouncer.h
@@ -673,6 +673,20 @@ typedef struct OutstandingRequest {
 	PgServerPreparedStatement *server_ps;
 
 	uint64_t server_ps_query_id;
+
+	/*
+	 * Value of the client its prepared_statement_seq counter when this
+	 * request was queued. A reset only invalidates client statements that
+	 * were registered at or before this point; anything the client
+	 * registered later in the same pipeline still exists on the server.
+	 */
+	uint64_t client_ps_seq;
+
+	/*
+	 * Same idea for the server its statement cache: what the server had
+	 * prepared when this request was queued.
+	 */
+	uint64_t server_ps_seq;
 } OutstandingRequest;
 
 enum ReplicationType {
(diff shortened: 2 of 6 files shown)


src/prepare.c
cosgroveb, line 356: ~In this implementation, the condition here would result in pgbouncer and Postgres disagreeing. Consider a pipeline with a `DEALLOCATE ALL` immediately followed by a `PREPARE` that re-prepares a statement that was previously prepared~

Update: I was wrong above. Yeah this weirdness already existed. I'll think about this more and may open a separate issue.

```sql
-- Already prepared earlier
PREPARE "PGBOUNCER_1" AS SELECT 1;

-- client sends these without waiting between them:
DEALLOCATE ALL;
PREPARE "PGBOUNCER_1" AS SELECT 1;
EXECUTE "PGBOUNCER_1";
```

Now that we are not unconditionally clearing prepared statements in `server.c` at this point Pgbouncer does not know that it must not omit the `PREPARE` (it has, basically, a stale cache). So this is an analogy by way of SQL of what Postgres functionally sees:

```sql
DEALLOCATE ALL;
-- PREPARE was omitted
EXECUTE "PGBOUNCER_1"; -- ERROR: statement no longer exists
```

Funny. This is exactly what [the comment block](https://github.com/pgbouncer/pgbouncer/pull/1543/changes#diff-1abc5651133d108c0c420d9411925373c711133e7748d9e4f4c97d5fb543fdd9L500-L506) before the new partial state clearing warns about:

```c
		/*
		 * Clean up prepared statements if needed if the client sent a
		 * DEALLOCATE ALL or a DISCARD ALL query. Not doing so would
		 * confuse our prepared statement handling, because we would
		 * expect certain queries to be prepared at the server that are
		 * not.
		 */
```

m0g3r: Thanks for digging into it.

Noted on the update. If you do open a separate issue for the pipelined `DEALLOCATE ALL` + re-`PREPARE` case, I'm happy to take it — I already have a setup here that drives pipelined resets against a real backend, so I can run that exact sequence against unmodified master and report what it does either way.

And if you want anything changed in this PR in the meantime, just say.

JelteF:
Thanks for the contribution. Without some context when a user actually runs into this in a practical setting (see [my comment on the issue](https://github.com/pgbouncer/pgbouncer/issues/1535#issuecomment-5298723579) I'm not sure it's worth complicating the prepared statement logic even further to fix this. Especially since this doesn't fix all cases afaict: I believe that if you recreate a prepared statement with the same statement name before and after the `DEALLOCATE ALL`, you'll get an error that it already exists. I don't think that's solvable easily.

JelteF:
Closing this, see #1535

Issue #318, opened 2018-08-10
Feature idea: Support for GSS/Kerberos


ktosiek:
I have a PostgreSQL cluster in a Kerberos (Active Directory) environment. I'd like to use pgbouncer in that environment while keeping the Kerberos authentication.
I think this could be achieved by:
* implementing the gss authentication method
* using Constrained Delegation / Kerberos S4UProxy to pass the client credentials to the PostgreSQL cluster. This would be transparent to clients.

Problems I can see:
* GSS API is synchronous, and I don't think there are any guarantees about thread safety. Are there any components like this already used? Would I need to keep GSS calls on a separate thread?

ktosiek:
I've started working on this, but there's an architectural decision I'm not sure about: what to do with the "borrowed" user's credential?

Just storing it on the PgPool won't work - I need a fresh credential to authenticate to the server, but I can't just blindly put it there each time a client connects because I might need to use the same client credential in multiple calls to GSS. Also, I can't copy it to PgSocket - there's no way to copy the token I'm getting from GSS. Maybe some kind of garbage collection might work here? I have a scheme in mind, but if there's anything similar already in pgbouncer I'd love a hint.

petere:
I can't comment on the technical details, but GSS/Kerberos support in PgBouncer would be welcome.

Godwottery:
I second the fact that it would be welcome. I would be able to help with testing.

Godwottery:
Without fully understanding the question: I would like pgBouncer to pass along the authentication information that the user has, so that the user can authenticate via pgBouncer using their domain credentials. I would like to have only a single source of user-info, which is the AD-server, and then have everything authenticate against this.  I think that would be constrained/unconstrained delegation.

aiobofh:
@ktosiek how did your work with this progress?

My thinking is that every connected user should have a more permanent connection to the database through the bouncer (a pool of connections), which is set-up on the first GSSAPI login to the pgbouncer, until it times out or something. And the users should in turn be able to connect and disconnect as they please to the pgbouncer from their clients. Right? I'm very curious on your progress and findings! :D

ktosiek:
@aiobofh I've stopped working on this for now.

Your thinking is correct for first connection to the backend db, but pgpool might need to connect as the same user without getting a new connection on the front-end.

I think I was too attached to the idea of always using a client-provided token, when in fact we could just use S4U2Self when what we have expires.

aiobofh:
[...]

PR #1545, opened 2026-08-14
GSSAPI (Kerberos) authentication and encryption for client and backend connections

julesroussel3:
Adds GSSAPI/Kerberos support to PgBouncer: authentication (acceptor and initiator) and transport encryption (`gssencmode=require` / `gssencmode=prefer`), enabling fully Kerberos-secured connection pooling without passwords.

### Authentication

**Client (acceptor)**: pgbouncer validates Kerberos service tickets using `gss_accept_sec_context()` and maps authenticated principals to local usernames via `gss_localname()`.

**Backend (initiator)**: pgbouncer authenticates to postgres using the process's existing TGT from the credential cache (KCM, FILE:, etc.), with an optional keytab override for environments without automatic credential management.

### Encryption

**Client (acceptor)**: pgbouncer accepts `GSSENCRequest` from libpq clients and performs the GSS encryption handshake as the acceptor, establishing a `gss_wrap()`/`gss_unwrap()` encrypted channel.

**Backend (initiator)**: pgbouncer sends `GSSENCRequest` to postgres backends and performs the GSS encryption handshake as the initiator. When the backend responds `'N'`, pgbouncer falls back to SSL or plain; under `server_gssencmode = prefer` it also skips GSS when no initiator credentials are available.

**Transport layer**: GSS encryption uses the same `SBufIO` vtable abstraction as TLS. After the handshake, `gss_sbufio_ops` replaces `raw_sbufio_ops`, routing all I/O through `gss_wrap()`/`gss_unwrap()` transparently. A dedicated `SBUF_EV_GSS_READY` event signals handshake completion.

**Identity extraction**: When GSS encryption is active and the configured method is GSSAPI, the client's Kerberos identity is already established by the encryption handshake. No separate `AUTH_REQ_GSS` exchange is needed; the principal is extracted via `gss_inquire_context()` and mapped via `gss_localname()`. Any other configured method still runs over the encrypted channel.

### Differences from PR #1209

This is a parallel implementation to #1209. Both address the same feature request (#318). The key architectural differences:

**Event-driven, not threaded.** `gss_accept_sec_context()` is a local cryptographic operation with no network I/O, in the same class as verifying a SCRAM proof. PostgreSQL itself runs it inline. Threading adds locking, request queues, and thread lifecycle complexity for an operation that does not block.

**Multi-round `GSS_S_CONTINUE_NEEDED` support.** The GSSAPI spec requires support for multi-round context establishment. Each round is a separate sbuf event; GSS context state persists on the PgSocket struct. PR #1209's author noted this as an unresolved concern: "I also am not sure how this loop type flow could be implemented in the pgbouncer async framework." The event-driven approach handles it naturally.

**`gss_localname()` for principal-to-username mapping.** Rather than implementing PostgreSQL's `include_realm` and `krb_realm` HBA options (application-level ad-hoc realm stripping), this delegates mapping to `auth_to_local` rules in `krb5.conf` via `gss_localname()`. This is the site-policy-respecting approach: the same rules apply to all Kerberos services on the system. PostgreSQL added `gss_localname()` support in v12 as the preferred alternative. MIT Kerberos >= 1.7 (2009) is universally available on systems that can build with GSSAPI. `include_realm` is redundant with `auth_to_local` and is ignored with a warning, so administrators can copy GSSAPI HBA rules from postgres without modification; the restrictive `krb_realm` and `map=` options are rejected (fail closed) rather than silently weakened, so a line never grants more access than written.

**Credential-cache-default initiator.** Backend connections use the process's existing TGT from the default credential cache (KCM, FILE:, etc.), like any Kerberos client. An explicit keytab override (`auth_gssapi_client_keytab`) is available for environments without automatic credential management, but it is not the default.

**Transport encryption.** This implementation adds full GSSAPI transport encryption support with `client_gssencmode` and `server_gssencmode` configuration, matching PostgreSQL's `gssencmode` semantics. The encryption context requests `GSS_C_MUTUAL_FLAG | GSS_C_REPLAY_FLAG | GSS_C_SEQUENCE_FLAG | GSS_C_CONF_FLAG | GSS_C_INTEG_FLAG`, matching postgres's `fe-secure-gssapi.c`. This differs intentionally from the auth-only context which uses only `GSS_C_MUTUAL_FLAG`: the encryption context needs replay and sequence protection since it carries application data. When `client_gssencmode = disable` (the default), no encryption buffers are allocated and there is zero per-connection overhead.

### New configuration parameters

- `auth_gssapi_keytab`: path to the acceptor keytab (required)
- `auth_gssapi_client_keytab`: optional initiator keytab override
- `auth_gssapi_service_name`: backend SPN service name (default: `postgres`)
- `client_gssencmode`: GSS encryption mode for client connections (`disable`, `allow`, `require`; default: `disable`)
- `server_gssencmode`: GSS encryption mode for backend connections (`disable`, `prefer`, `require`; default: `disable`)

The asymmetry reflects negotiation role: a server responds to what the client requests (`disable`/`allow`/`require`), while a client controls how it initiates (`disable`/`prefer`/`require`).

### Testing

A test suite (`test/test_auth_gss.py`, 26 cases) covers authentication, transport encryption, backend-encryption fallback, downgrade/injection refusal, and error paths, run against a local MIT KDC (`test/setup_kdc.sh`). A GitHub Actions job builds with `--with-gssapi`, sets up the KDC, and runs it; the tests auto-skip when GSSAPI is not built or the KDC is not running.

### References

- Issue #318: Feature idea: Support for GSS/Kerberos
- PR #1209: GSS Auth Implementation (parallel effort)
- PR #441: Add Kerberos token based GSSAPI authentication (closed)
- PR #743: Add GSS/KRB5 Encryption support (closed)

### DIFF
--- a/.github/workflows/pgbouncer-ci.yml
+++ b/.github/workflows/pgbouncer-ci.yml
@@ -67,6 +67,12 @@ jobs:
             cflags: "-O0 -g"
             enable_valgrind: "yes"
             meson_args: "-Dcares=enabled -Dldap=disabled -Dpam=disabled"
+          # GSSAPI/Kerberos auth + encryption, exercised against a local MIT KDC
+          # set up by test/setup_kdc.sh (gated on setup_kdc below).
+          - name: gssapi
+            build_system: meson
+            meson_args: "-Dgssapi=enabled"
+            setup_kdc: "yes"
     env:
       PGVERSION: ${{ matrix.pgversion || '16' }}
       BUILD_SYSTEM: ${{ matrix.build_system }}
@@ -79,6 +85,7 @@ jobs:
       CFLAGS_OVERRIDE: ${{ matrix.cflags }}
       ENABLE_VALGRIND: ${{ matrix.enable_valgrind }}
       DIST: ${{ matrix.dist }}
+      SETUP_KDC: ${{ matrix.setup_kdc }}
     steps: &linux_steps
       - uses: actions/checkout@v4
       - name: Install dependencies
@@ -91,6 +98,7 @@ jobs:
           pkgs="autoconf automake ca-certificates cpio git iptables ldap-utils libc-ares-dev libevent-dev libldap-dev libpam0g-dev libssl-dev libsystemd-dev libtool make meson ninja-build pandoc pkg-config postgresql-$PGVERSION python3 python3-pip python3-venv slapd socat sudo"
           if [ "$ENABLE_VALGRIND" = "yes" ]; then pkgs="$pkgs valgrind"; fi
           if [ "$USE_SCAN_BUILD" = "yes" ]; then pkgs="$pkgs clang clang-tools"; fi
+          if [ "$SETUP_KDC" = "yes" ]; then pkgs="$pkgs krb5-kdc krb5-admin-server krb5-user libkrb5-dev"; fi
           # $pkgs is an intentionally word-split package list.
           # shellcheck disable=SC2086
           sudo apt-get -y --no-install-recommends install $pkgs
@@ -112,6 +120,9 @@ jobs:
           [ -n "$CFLAGS_OVERRIDE" ] && export CFLAGS="$CFLAGS_OVERRIDE"
           [ "${USE_SCAN_BUILD:-}" = "yes" ] && export SCANBUILD="scan-build"
           sh ci/run.sh build "$BUILD_SYSTEM"
+      - name: Set up Kerberos KDC
+        if: ${{ env.SETUP_KDC == 'yes' }}
+        run: sudo bash test/setup_kdc.sh
       - name: Test
         run: |
           set -e
--- a/doc/gssapi-auth.md
+++ b/doc/gssapi-auth.md
@@ -1,0 +1,429 @@
+# GSSAPI (Kerberos) Authentication in PgBouncer
+
+## Overview
+
+PgBouncer supports GSSAPI/Kerberos authentication in both directions:
+
+- **Client → pgbouncer (acceptor)**: connecting clients present Kerberos service
+  tickets. pgbouncer validates the ticket using its keytab, maps the
+  authenticated principal to a local username via `gss_localname()` and the
+  site's `auth_to_local` rules in `krb5.conf`, and verifies the result matches
+  the username the client claimed in the startup packet.
+
+- **pgbouncer → postgres (initiator)**: when the backend requests GSSAPI
+  authentication, pgbouncer authenticates using its own pool service-account
+  identity. By default it uses whatever TGT already exists in the process
+  credential cache (KCM, FILE:, or any other standard ccache). No client
+  credentials are ever forwarded.
+
+No passwords are used anywhere.
+
+## Trust model and authorization boundary
+
+pgbouncer is the **authentication boundary**. It verifies each client's
+Kerberos identity and logs the authenticated principal. Postgres trusts the
+pool service account (e.g., `pgbouncer@EXAMPLE.COM`) and applies authorization
+at that role's level.
+
+```
+  [alice@workstation]          [pgbouncer]               [postgres]
+        |                          |                          |
+        |  StartupMessage          |                          |
+        |  user=alice, db=mydb ───>                           |
+        |                          |                          |
+        |  AuthenticationGSS(7) <──|                          |
+        |                          |                          |
+        |  GSSResponse (AP-REQ) ───>                          |
+        |   (svc ticket for        |  gss_accept_sec_context()|
+        |    postgres/<pgb-host>)  |                          |
+        |                          |  gss_localname():        |
+        |  AuthenticationOk <──────|  alice@EXAMPLE.COM       |
+        |                          |   -> alice               |
+        |                          |  (matches claimed "alice")|
+        |                          |                          |
+        |  query ──────────────────>  StartupMessage          |
+        |                          |  user=pgbouncer, db=mydb─>
+        |                          |                          |
+        |                          |  AuthenticationGSS(7) <──|
+        |                          |                          |
+        |                          |  GSSResponse (AP-REQ) ───>
+        |                          |   (svc ticket for        |
+        |                          |    postgres/<pg-host>)   |
+        |                          |  gss_init_sec_context()  |
+        |                          |                          |
+        |                          |  AuthenticationOk <──────|
+        |                          |                          |
+        |  result <────────────────────────────────── query ──>
+```
+
+Postgres sees all connections as the `pgbouncer` role. The `pgbouncer` role is
+granted the necessary database permissions. This is not a limitation — it is
+the correct design for a connection pooler:
+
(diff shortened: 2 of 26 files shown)


AndrewJackson2020:
Author of #1209 here. I would love to hand this concept off to someone else as I no longer work in an environment where GSSAPI is relevant. 

That said I'm not sure if the direction that this PR takes is correct. Specifically:

> gss_accept_sec_context() is a local cryptographic operation with no network I/O, in the same class as verifying a SCRAM proof. PostgreSQL itself runs it inline. Threading adds locking, request queues, and thread lifecycle complexity for an operation that does not block.

If this was the case it would make the GSS implementation way more simple than what I did in #1209. Unfortunately according too the RFC linked in the postgres docs[0], specifically `2.2.2:  GSS_Accept_sec_context call`.

> This call may block pending network interactions for those mech_types in which a directory service or other network entity must be consulted on behalf of a context acceptor in order to validate a received input_token.

It seems like there is no guarantee that it is not network blocking and it is specifically called out in the RFC. I am not an expert on GSSAPI so I could be missing something, again happy to be wrong here.

[0] https://datatracker.ietf.org/doc/html/rfc2743

julesroussel3:
> there is no guarantee that it is not network blocking and it is specifically called out in the RFC

 Fair, that may have been too broad a claim and should probably be scoped to the krb5 mechanism. That RFC line is about mechanisms that need an online directory. For krb5 the acceptor decrypts the request with the server's own key from its keytab, so there's no KDC call. The receipt path is RFC 4120 §3.2.3 [1], and MIT documents it as a keytab lookup [2].

The initiator does hit the network though. gss_init_sec_context to postgres needs the backend service ticket, which means an inline TGS round trip when it isn't cached, so it blocks the loop until that returns or times out. It's cached after the first backend connection until expiry.

If you've seen krb5 accept touch the network let me know because that'd be a design change. Otherwise I think this ends up a doc-only update.

[1]: https://datatracker.ietf.org/doc/html/rfc4120#section-3.2.3
[2]: https://web.mit.edu/kerberos/krb5-devel/doc/appdev/gssapi.html#acceptor-names

AndrewJackson2020:
So MIT krb5 does offer other mechanisms than the default Kerberos. The 2 I played around with are SPNEGO and IAKERB. SPNEGO should just have the same end result as default kerberos (client negotiates with server what kerberos mechanism it supports). IAKERB does seem to talk to the KDC during `gss_accept_sec_context` [0] but I had to modify the postgres server source code in order to get it to accept IAKERB mechanism. Given this, assuming `gss_accept_sec_context` does not do any network blocking with MIT KRB5 is maybe okay.

Is there any impact with alternative kerberos implementations outside of MIT KRB5 like Heimdal? Also is there any possibility that a postgres installation can use a non kerberos auth method while still using the GSSAPI auth interface? Should we be concerned about the possibility that postgres/libpq adds non default mechanism selection which could contain network blocking calls?

If none of the above are a concern then I think we are probably fine. If they are a concern I think Its safer to stick with the threaded auth infra that pam and ldap currently use. We will see what the maintainers decide.

[0] https://github.com/krb5/krb5/blob/473b51b640bef2e97c490d6f2985658befa89c16/src/lib/gssapi/krb5/iakerb.c#L380

julesroussel3:
I went and traced this with `strace`. A normal krb5 accept sends no packets to the KDC. Validation is all local against the keytab. So the RFC 2743 caveat about blocking calls doesn't apply to krb5.

The default acceptor set MIT hands us is krb5 plus SPNEGO, and IAKERB isn't in it. That lines up with you having to patch the server to negotiate IAKERB, since a client can't select it on its own. SPNEGO accept was local in the trace too, it just negotiates down to krb5.

Heimdal is out of scope because this only builds against MIT. It depends on MIT's library and a couple of MIT-only functions.

The future-libpq case seemed worth checking properly, so I wired in a restriction and tested it. Locking our acceptor down to krb5 drops the cred to a single mechanism, normal clients still get accepted with no KDC traffic, and anything that isn't krb5 gets refused with GSS_S_BAD_MECH before it even dispatches, so nothing touches the network. I tested that against SPNEGO rather than IAKERB itself. Driving IAKERB would take deliberate setup (you had to patch the server just to get it negotiated), but it's the same rejection path either way. Honestly it may be unnecessary, since IAKERB isn't in the default set to begin with. I'm happy to add it if you or the maintainers want the belt and suspenders, otherwise I think we're fine as is and this stays a doc-only change.

.github/workflows/pgbouncer-ci.yml
```diff
@@ -112,6 +120,9 @@ jobs:
           [ -n "$CFLAGS_OVERRIDE" ] && export CFLAGS="$CFLAGS_OVERRIDE"
           [ "${USE_SCAN_BUILD:-}" = "yes" ] && export SCANBUILD="scan-build"
           sh ci/run.sh build "$BUILD_SYSTEM"
+      - name: Set up Kerberos KDC
+        if: ${{ env.SETUP_KDC == 'yes' }}
+        run: sudo bash test/setup_kdc.sh
```
AndrewJackson2020, line 125: I think we should move this away from root and into a pytest fixture, similar to how postgres and pgbouncer are set up. This will make it easier to run the tests locally. Take a look at the postgres tests for kerberos [0] for example.

[0] https://github.com/postgres/postgres/blob/master/src/test/perl/PostgreSQL/Test/Kerberos.pm

doc/gssapi-auth.md
```diff
@@ -0,0 +1,429 @@
...
+### No credential delegation
+
+pgbouncer requests **only** `GSS_C_MUTUAL_FLAG` when calling
+`gss_init_sec_context()` for backend connections. `GSS_C_DELEG_FLAG` and
+`GSS_C_DELEG_POLICY_FLAG` are never set.
+
+Credential delegation — forwarding a client's TGT so pgbouncer could
+authenticate to postgres *as the client* — is deliberately not used:
+
+- It destroys connection reuse: each user's forwarded credential is distinct,
```
AndrewJackson2020, line 85: Are the risks with credential delegation greater than with SCRAM pass through? Also any client can disable this during the kinit process, correct (ie its opt in)? 

The security issues here are definitely riskier in pgbouncer with a single process architecture (as opposed to postgres) but I was not expecting delegated credentials to be disallowed in a pgbouncer implementation entirely.

AndrewJackson2020:
> Multi-round GSS_S_CONTINUE_NEEDED support. The GSSAPI spec requires support for multi-round context establishment. Each round is a separate sbuf event; GSS context state persists on the PgSocket struct. PR https://github.com/pgbouncer/pgbouncer/pull/1209's author noted this as an unresolved concern: "I also am not sure how this loop type flow could be implemented in the pgbouncer async framework." The event-driven approach handles it naturally.

Do any of your unit tests trigger GSS_S_CONTINUE_NEEDED from `gss_accept_sec_context()`? This was my biggest concern with implementing it in my PR: implementing a feature that I was not sure how to test.

PR #1549, opened 2026-08-18
Fix non-canonical distribution archive paths

### LINKED ISSUE #1498: Release tarball for 1.25.1 stores files with non-canonical `/./`
The release tarball at https://www.pgbouncer.org/downloads/files/1.25.1/pgbouncer-1.25.1.tar.gz stores four entries with a /./ segment in the middle of their path:

```shell
$ tar -tzf pgbouncer-1.25.1.tar.gz | grep '/\./'
pgbouncer-1.25.1/./config.guess
pgbouncer-1.25.1/./config.sub
pgbouncer-1.25.1/./configure
pgbouncer-1.25.1/./install-sh
```

Every other entry uses canonical `pgbouncer-1.25.1/<name>` paths. Note the GitHub tag tarball at https://github.com/pgbouncer/pgbouncer/archive/refs/tags/pgbouncer_1_25_1.tar.gz doesn't include the above files,

Most extractors silently canonicalize `/./` away. But some CIs treat these malformed paths as fatal:

```shell
/usr/bin/gzip -dc pgbouncer-1.25.1.tar.gz | /usr/bin/tar -xof -
/usr/bin/tar: pgbouncer-1.25.1/./config.guess: Cannot open: No such file or directory
/usr/bin/tar: pgbouncer-1.25.1/./config.sub: Cannot open: No such file or directory
/usr/bin/tar: pgbouncer-1.25.1/./configure: Cannot open: No such file or directory
/usr/bin/tar: pgbouncer-1.25.1/./install-sh: Cannot open: No such file or directory
/usr/bin/tar: Exiting with failure status due to previous errors
```

As a workaround, extracting the tarball using `--strip-components=1` bypasses the issue

KarunakarKotha09:
## Summary

Fixes #1498 by removing the unintended `./` path component from archive prefixes used for root-level files added to the Autoconf distribution tarball.

The generated archive now uses canonical paths such as `pgbouncer-1.25.2/configure` instead of `pgbouncer-1.25.2/./configure`. Nested paths continue to retain their intended directory structure.

## Implementation

The Makefile recipe strips only a leading `./` from `$(dir $(file))` before passing the path to `git archive`. This keeps the fix scoped to root-level files and avoids changing archive layout for nested files.

## Testing

- Generated a distribution archive with the Makefile recipe.
- Verified no archive entry contains `/./`.
- Verified root-level and nested files extract at the expected paths.
- Confirmed the full CI suite passes, including Windows, macOS, Linux, Valgrind, and formatting/linting.

Fixes #1498

## Author and contact

Author: Karunakar Kotha

For questions about this change, please contact kkotha@microsft.com.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,7 @@ EXTRA_DIST = config.guess config.sub configure install-sh lib/usual/config.h.in
 dist: $(distdir).tar.gz
 
 $(PACKAGE_TARNAME)-$(PACKAGE_VERSION).tar.gz:
-	$(GIT) -C $(srcdir) -c core.autocrlf=false archive --format tar.gz -9 --prefix $(distdir)/ $(PG_GIT_REVISION) -o $(abs_top_builddir)/$@ $(foreach file,$(EXTRA_DIST),--prefix $(distdir)/$(dir $(file)) --add-file=$(file)) --prefix $(distdir)/
+	$(GIT) -C $(srcdir) -c core.autocrlf=false archive --format tar.gz -9 --prefix $(distdir)/ $(PG_GIT_REVISION) -o $(abs_top_builddir)/$@ $(foreach file,$(EXTRA_DIST),--prefix $(distdir)/$(patsubst ./,,$(dir $(file))) --add-file=$(file)) --prefix $(distdir)/
 
 #
 # test


cosgroveb:
Is this a genuine problem? Like @petere on #1498 I haven't been able to reproduce on MacOS or Linux. What platform, tar version, and invocation gives you difficulty?

Notably, the Debian Postgres Team [consumes](https://salsa.debian.org/postgresql/pgbouncer/-/blob/master/debian/watch) pgbouncer release tarballs to build new packages. Debian built `1.25.2-1` successfully on May 9th. I can't speak for other downstream packagers.

KarunakarKotha09:
Thanks for checking this. I agree that the issue is not reproducible with common GNU/FreeBSD tar setups, and Debian’s build of `1.25.2-1` confirms that normal extraction/build workflows are unaffected.

The issue is specifically that the `1.25.1` release archive contains non-canonical member names:
```

pgbouncer-1.25.1/./configure
pgbouncer-1.25.1/./config.guess
pgbouncer-1.25.1/./config.sub
pgbouncer-1.25.1/./install-sh
```
Most extractors normalize these paths, so the practical impact is limited. My proposed fix only canonicalizes the leading `./` for root-level files and leaves nested paths unchanged. PR #1549 implements this and its CI checks pass.

If canonical archive paths are not considered necessary, I understand if this is treated as harmless.

PR #1553, opened 2026-08-19
Support plan_cache_mode startup parameter

### LINKED ISSUE #1495: Handle plan_cache_mode in session mode (and perhap's transaction mode also)
Hi,

Using PgBouncer (version 1.25.1) with [CloudNative-Pg](https://cloudnative-pg.io/), we have a problem with prepared statements. My colleagues had identified the problem in [this blog article](https://mirakl.tech/how-a-postgresql-performance-optimisation-resulted-in-a-1000x-slowdown-37e7869e4a70).

In order to prevent the transaction preparation, yesterday, we tried to implement adding into PGOPTIONS `plan_cache_mode=force_custom_plan`, but we have never seen non-prepared queries on the PostgreSQL instance.

Finally, we used a workaround, setting `prepareThreshold=0`, that fixed our issue.

Today, we are making new tests, and find that the parameter is not configured in the connection after Pgbouncer.

Test script:

```sh
pooler_host="pooler-${myinstance}-rw.${myinstance}.svc.cluster.local"
direct_host="${myinstance}-pg17-rw.${myinstance}.svc.cluster.local"

psql "host=${pooler_host} port=5432 user=${PGUSER} dbname=${PGDATABASE}" -c "SHOW plan_cache_mode;"

PGOPTIONS="-c plan_cache_mode=force_custom_plan" psql "host=${pooler_host} port=5432 user=${PGUSER} dbname=${PGDATABASE}" -c "SHOW plan_cache_mode;"

PGOPTIONS="-c plan_cache_mode=force_custom_plan" psql "host=${direct_host} port=5432 user=${PGUSER} dbname=${PGDATABASE}" -c "SHOW plan_cache_mode;"
```

Notes:

* `pooler_host` uses pgbouncer
* `direct_host` does not use pgbouncer

Related output:

```console
 plan_cache_mode 
-----------------
 auto
(1 row)

 plan_cache_mode 
-----------------
 auto
(1 row)

  plan_cache_mode  
-------------------
 force_custom_plan
(1 row)

```

KarunakarKotha09:
## Summary

Fixes #1495 by tracking PostgreSQL's `plan_cache_mode` startup parameter.

When clients send `PGOPTIONS="-c plan_cache_mode=force_custom_plan"`, PgBouncer now preserves the value in its per-client variable cache and applies it to the assigned server connection. This works even when `options` is listed in `ignore_startup_parameters`, because recognized tracked parameters are processed before ignored parameters.

## Testing

- Added a regression test covering `PGOPTIONS`/`options` startup handling.
- `git diff --check` passed.
- CI will run the full project test matrix.

## Author

Karunakar Kotha

Questions: kkotha@microsft.com


JelteF:
Adding settings there safely by default requires postgres to report them back. It doesn't do that currently for `plan_cache_mode`: https://www.postgresql.org/docs/15/protocol-flow.html#PROTOCOL-ASYNC

PR #1554, opened 2026-08-19
Refresh cached timezone after database changes

### LINKED ISSUE #1255: Timezone persists after ALTER DATABASE
## Issue
After changing a database timezone, pgbouncer will incorrectly persist the previous timezone in future sessions, effectively hiding this change.

## Context
- pgbouncer `1.21.0`, `1.22.1`, `1.23.1`
- PostgreSQL `15.10`
- server in UTC

pgbouncer configuration is mostly unchanged from the default.
```
[pgbouncer]
logfile = /var/log/postgresql/pgbouncer.log
pidfile = /var/run/postgresql/pgbouncer.pid
listen_addr = ...
listen_port = 6432
unix_socket_dir = /var/run/postgresql
client_tls_sslmode = require
client_tls_key_file = ...
client_tls_cert_file = ...
auth_type = scram-sha-256
auth_query = SELECT usename, passwd FROM pg_shadow WHERE usename=$1
auth_user = pgbouncer
max_client_conn = 100000
default_pool_size = 100
server_login_retry = 0
```

## Reproduction step
Timezone setting is shown with `SELECT boot_val,reset_val,setting, source FROM pg_settings WHERE name = 'TimeZone';`
- create a fresh DB with no explicit timezone set
- connect from a user A:
```
boot_val | reset_val | setting |       source      
---------+-----------+---------+-------------------
GMT      | Etc/UTC   | Etc/UTC | configuration file
```
- change database default timezone & reconnect: **timezone did not change**. Connections skipping pgbouncer get the appropriate timezone.
```
boot_val | reset_val | setting |       source      
---------+-----------+---------+-------------------
GMT      | Etc/UTC   | Etc/UTC | configuration file
```
- kill connections with `pg_terminate_backend` & reconnect: database TZ is visible in `reset_val`, **but the session overrides this with the previous TZ** (despite not requested). Again, connections without pgbouncer only show the DB default.
```
boot_val |  reset_val   | setting | source 
---------+--------------+---------+--------
GMT      | Europe/Paris | Etc/UTC | session
```

## Additional info
- all future connections with the same user A & database are affected
- connecting with user A to db `postgres` shows no issue, until switching `\c affected_db`.
- Running `reset timezone` clears the unexpected session value, fixing the problem for the current session only. Reconnecting will again show the unwanted session-level override.
- Waiting for more than `server_idle_timeout` does not help
- Running `RECONNECT affected_db` on pgbouncer works

## Current workaround
Force the expected timezone for this DB in pgbouncer configuration
```
[databases]
affected_db = host=127.0.0.1 port=5432 timezone=Europe/Paris
* = host=127.0.0.1 port=5432
```

KarunakarKotha09:
## Summary

Fixes #1255 by keeping the pool's cached server parameters current after PostgreSQL reports a changed database setting.

When a database-level `TimeZone` setting changes, the existing welcome packet remains valid for already-connected clients, but the cached defaults used for new clients must be refreshed. PgBouncer now updates `orig_vars` whenever startup `ParameterStatus` messages arrive, even after the pool welcome message is ready. New sessions therefore receive the database's current timezone instead of a stale value from the first connection.

## Testing

- Added a regression test covering `ALTER DATABASE ... SET timezone`, PgBouncer reconnect, and restoration with `RESET timezone`.
- `git diff --check` passed.
- Full build and test validation will run in CI.

## Author

Karunakar Kotha

Questions: kkotha@microsft.com

### DIFF
--- a/src/proto.c
+++ b/src/proto.c
@@ -320,37 +320,37 @@ void log_server_error(const char *note, PktHdr *pkt)
  */
 
 /* add another server parameter packet to cache */
-bool add_welcome_parameter(PgPool *pool, const char *key, const char *val)
+bool add_welcome_parameter(PgSocket *server, const char *key, const char *val)
 {
+	PgPool *pool = server->pool;
 	PktBuf *msg = pool->welcome_msg;
 
-	/*
-	 * Keep the cached server defaults current when a database setting changes
-	 * while the pool remains alive. The welcome packet itself is immutable once
-	 * it has been sent, so only append parameters while it is being built.
-	 */
-	if (!varcache_set(&pool->orig_vars, key, val) && !pool->welcome_msg_ready) {
-		if (!msg) {
-			msg = pktbuf_dynamic(128);
-			if (!msg)
-				return false;
-			pool->welcome_msg = msg;
-		}
-
-		/* first packet must be AuthOk */
-		if (msg->write_pos == 0)
-			pktbuf_write_AuthenticationOk(msg);
-
-		pktbuf_write_ParameterStatus(msg, key, val);
+	if (!msg) {
+		msg = pktbuf_dynamic(128);
+		if (!msg)
+			return false;
+		pool->welcome_msg = msg;
 	}
 
-	return !msg || !msg->failed;
+	/* first packet must be AuthOk */
+	if (msg->write_pos == 0)
+		pktbuf_write_AuthenticationOk(msg);
+
+	/*
+	 * Do not publish values until this server reaches ReadyForQuery. A backend
+	 * can send ParameterStatus before rejecting the startup.
+	 */
+	if (!varcache_set(&server->startup_vars, key, val) && !pool->welcome_msg_ready)
+		pktbuf_write_ParameterStatus(msg, key, val);
+
+	return !msg->failed;
 }
 
 /* all parameters processed */
 void finish_welcome_msg(PgSocket *server)
 {
 	PgPool *pool = server->pool;
+	varcache_merge(&pool->orig_vars, &server->startup_vars);
 	if (pool->welcome_msg_ready)
 		return;
 	pool->welcome_msg_ready = true;
--- a/src/proto.c
+++ b/src/proto.c
@@ -324,25 +324,27 @@ bool add_welcome_parameter(PgPool *pool, const char *key, const char *val)
 {
 	PktBuf *msg = pool->welcome_msg;
 
-	if (pool->welcome_msg_ready)
-		return true;
+	/*
+	 * Keep the cached server defaults current when a database setting changes
+	 * while the pool remains alive. The welcome packet itself is immutable once
+	 * it has been sent, so only append parameters while it is being built.
+	 */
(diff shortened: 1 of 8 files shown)


cosgroveb (changes requested):
This is pretty close to my [target_session_attrs work](https://github.com/pgbouncer/pgbouncer/pull/1534) and [other adjacent](https://github.com/pgbouncer/pgbouncer/pull/1205) parameter [tracking work](https://github.com/cosgroveb/pgbouncer/pull/4) so I took a look.

I think that an edge case exists here where a server that isn't admitted to the back-end pool will send `ParameterStatus` messages like the timezone here and this PR would incorrectly populate them.

I have a test branch that demonstrates this:

https://github.com/KarunakarKotha09/pgbouncer/compare/fix-timezone-persistence...cosgroveb:pgbouncer:timezone-rejected-backend-test

The test:

- Configures good, bad hosts with a fake PostgreSQL backend
- Authenticate fake backend
- Occupy accepted backend
- Advance round-robin selection
- Send timezone, then error
- Observe poisoned timezone cache

This might look like a contrived test but I checked and GUCs [do precede login triggers](https://github.com/postgres/postgres/blob/083ac033419f690758508e08c1736089384bbee8/src/backend/tcop/postgres.c#L4701-L4705)

```c
/* Report any recently-changed GUC options */
ReportChangedGUCOptions();

ReadyForQuery(whereToSendOutput);
send_ready_for_query = false;
```

To do this more correctly, stage startup ParameterStatus values per backend, adding them them to orig_vars only after ReadyForQuery and discarding them on failure.

src/proto.c
```diff
@@ -324,25 +324,27 @@ bool add_welcome_parameter(PgPool *pool, const char *key, const char *val)
...
-		if (!msg)
-			return false;
-		pool->welcome_msg = msg;
-	}
+	/*
+	 * Keep the cached server defaults current when a database setting changes
+	 * while the pool remains alive. The welcome packet itself is immutable once
+	 * it has been sent, so only append parameters while it is being built.
+	 */
+	if (!varcache_set(&pool->orig_vars, key, val) && !pool->welcome_msg_ready) {
```
sambhusankar: The problem is that varcache_set() returns false when the variable was already cached (couldn't be set because it already exists), and true when it's successfully set or updated.

The test passes because it only checks if new clients eventually get the updated timezone (from orig_vars), but the implementation has a subtle bug where the welcome message packet isn't properly built with the latest parameter values in certain scenarios.

sambhusankar (changes requested):

KarunakarKotha09:
Thank you for the detailed review and for providing the reproducing test branch.

I understand the issue: `ParameterStatus` messages received from a backend that later fails can currently be stored in the shared cache, which may incorrectly affect future clients.


I will update the implementation to stage backend-specific startup parameters and commit them to `orig_vars` only after receiving `ReadyForQuery`. If backend authentication or startup fails before that point, the staged values will be discarded. I will also add regression coverage for the rejected-backend scenario and run the focused tests and formatting checks.


Thanks again for identifying this edge case.

PR #1555, opened 2026-08-19
Fix prepared-statement response ordering for batched Close

mble:
## Faults

PgBouncer can send prepared-statement responses in the wrong order when a batch mixes `Parse` and `Close` without an intervening `Sync`.

There are two related faults.

### Fake response after a skipped response

A cache eviction can produce this outstanding-request order:

`Parse / RA_FORWARD` → `Close / RA_SKIP` → `Close / RA_FAKE` → `Sync / RA_FORWARD`

When the server returns `CloseComplete` for the eviction, PgBouncer pops `RA_SKIP` and discards the packet. However, fake responses were flushed only after forwarded packets.

This leaves `RA_FAKE` at the queue head. The next real response cannot match or pop it, but is still forwarded to the client. The request queue then remains misaligned for the life of the server connection.

The client receives an unexpected response, and the poisoned server connection can later be assigned to another client.

### Direct `CloseComplete` overtakes buffered data

When PgBouncer suppresses a reused `Parse` and no earlier request exists, it queues a fake `ParseComplete` in `server->sbuf.extra_packets` without adding an outstanding-request entry.

`handle_close_statement_command()` previously treated an empty outstanding-request queue as sufficient evidence that it could send `CloseComplete` directly to the socket.

The direct `CloseComplete` could therefore overtake the buffered `ParseComplete`.

## Fix

This change:

- extracts fake-response handling into `send_pending_fake_responses()`;
- flushes eligible fake responses after both forwarded and skipped server
  packets;
- sends `CloseComplete` directly only when:
- the outstanding-request queue is empty;
- `sbuf.extra_packets` is empty; and
- the output `iobuf` has no pending bytes;
- otherwise records the client `Close` as `RA_FAKE`, preserving FIFO response
  order.

The existing allocation-failure behavior is unchanged: both links are
disconnected if a fake response cannot be queued.

## Tests

Two libpq pipeline regression tests cover the distinct failures:

1. `Parse` reuses a prepared query and queues a fake `ParseComplete` before
`Close`.
2. `Parse` evicts a cached statement, placing `RA_SKIP` directly before the fake
response for `Close`.

Both tests require this result order:

`COMMAND_OK` → `COMMAND_OK` → `PIPELINE_SYNC`

They also run another query on the same connection to confirm that it remains
usable.

Both tests fail against the parent commit and pass with this change.

Additional local validation:

- prepared-statement suite: 27 passed, 1 skipped;
- full differential suite: two additional passes and no additional failures;
- dual-PgBouncer reproduction, parent: 295 desynchronisations in 99,981 queries;
- dual-PgBouncer reproduction, patched: 0 desynchronisations in 112,566 queries.

Reproduction: https://github.com/ericallam/dual-pgbouncer-desync-reproduction


JelteF (approved):

JelteF:
Thank you! I'm pretty sure I saw these things randomly occuring in CI but I hadn't figured out what was causing them.

ronalders:
@JelteF @mble 
Thank you for fixing this. We appear to be encountering the same issue with pgJDBC prepared statements: transactions stall while PostgreSQL waits on ClientRead, and disabling server-side preparation with prepareThreshold=0 avoids the problem.

Since this fix is now merged into master, is there an expected release version or approximate release date that will include it?

PR #1558, opened 2026-08-20
Expose TCP keepalive settings in SHOW FDS

### LINKED ISSUE #1398: TCP Keepalive Settings In Admin Console
I've been having a hard time trying to view TCP keepalive settings from the linux side. Within postgres you can just use command like `SHOW tcp_keepalives_idle;` to get the current settings for that session but I do not believe there is any equivalent within the pgbouncer admin console. 

Maybe it would make sense to include these per socket settings in `SHOW FDS;` or something like that?

KarunakarKotha09:
## Summary

Fixes #1398 by exposing the TCP keepalive settings applied to each file descriptor in `SHOW FDS`.

The new columns report whether TCP keepalive is enabled and, when available, the configured idle, interval, and count values. Non-TCP and unsupported platforms return `NULL` for values that do not apply. The socket helpers centralize platform-specific option reads, and takeover handling preserves the metadata for inherited descriptors.

## Testing

- Added regression coverage for the new `SHOW FDS` fields.
- Ran whitespace/error validation with `git diff-tree --check`.
- Full PostgreSQL-backed test validation is delegated to upstream CI.

## Documentation

- Documented the new `SHOW FDS` columns in `doc/usage.md`.

## Author and contact

Author: Karunakar Kotha

Contact: kkotha@microsft.com

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -687,6 +687,19 @@ cancel
 link
 :   fd for corresponding server/client.  NULL if idle.
 
+keepalive
+:   Whether TCP keepalive is enabled for the socket, or **-1** when it does
+    not apply or is unavailable.
+
+keepidle
+:   TCP keepalive idle time in seconds, or **-1** when it is unavailable.
+
+keepintvl
+:   TCP keepalive interval in seconds, or **-1** when it is unavailable.
+
+keepcnt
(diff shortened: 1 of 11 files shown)


eulerto:
I'm closing this PR because commit e86ba07f51dd16790c2ce2364397826e8141f33a removed SHOW FDS command.

PR #1563, opened 2026-08-23
scram: use OpenSSL for PBKDF2 when available 

mble:
PgBouncer derives the SCRAM `SaltedPassword` on the event loop. The event
loop is single threaded, so each login delays all other connections.

The bundled PostgreSQL fallback uses a software SHA-256. It costs 4.0 ms
of CPU for each login on aarch64 and 2.5 ms on x86_64. PR #1356
replaced the `libusual` SHA-256 with this fallback and made each login 1.3
times slower.

PgBouncer already links OpenSSL for TLS. OpenSSL gives the same result
8.3 times faster on aarch64 and 5.7 times faster on x86_64. On a CPU
with no SHA instructions it is 2.9 times faster, because the fallback
does 4 SHA-256 compressions for each iteration and OpenSSL does 2.

PBKDF2 is fully specified, so the output does not change. The fallback
stays for builds without OpenSSL.

This might be a little controversial, happy to discuss. SCRAM remains
a bane due to clients not often doing what PgBouncer does internally - 
caching - making any challenge expensive.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -439,7 +439,7 @@ Default: 200
 The number of computational iterations to be performed when encrypting a
 password using SCRAM-SHA-256. A higher number of iterations provides additional
 protection against brute-force attacks on stored passwords, but makes
-authentication slower.
+authentication slower. This value must be at least 1.
 
 Default: 4096
 
--- a/lib/m4/usual.m4
+++ b/lib/m4/usual.m4
@@ -456,8 +456,8 @@ if test "$tls_support" = "auto" -o "$tls_support" = "libssl"; then
                     [[SSL_CTX *ctx = SSL_CTX_new(SSLv23_method());]])],
     [ tls_support=yes; AC_MSG_RESULT([found])],
     [ AC_MSG_ERROR([not found]) ])
-  dnl check LibreSSL-only APIs
-  AC_CHECK_FUNCS(SSL_CTX_use_certificate_chain_mem SSL_CTX_load_verify_mem asn1_time_parse)
+  dnl check optional OpenSSL and LibreSSL APIs
+  AC_CHECK_FUNCS(PKCS5_PBKDF2_HMAC SSL_CTX_use_certificate_chain_mem SSL_CTX_load_verify_mem asn1_time_parse)
   CPPFLAGS="$tmp_CPPFLAGS"
   LDFLAGS="$tmp_LDFLAGS"
   LIBS="$tmp_LIBS"
--- a/meson.build
+++ b/meson.build
@@ -326,6 +326,11 @@ if openssl.found()
   cdata.set('OPENSSL_API_COMPAT', '0x00908000L',
             description: 'OpenSSL API version in use, to silence deprecation warnings.')
 
+  cdata.set('HAVE_PKCS5_PBKDF2_HMAC',
+            cc.has_function('PKCS5_PBKDF2_HMAC',
+                            prefix: '#include <openssl/evp.h>',
+                            dependencies: openssl) ? 1 : false)
+
   # LibreSSL-only helpers, used when present.
   foreach func : ['SSL_CTX_use_certificate_chain_mem', 'SSL_CTX_load_verify_mem', 'asn1_time_parse']
     cdata.set('HAVE_' + func.to_upper(),
--- a/src/main.c
+++ b/src/main.c
@@ -217,6 +217,8 @@ int cf_scram_iterations;
 
 static bool set_defer_accept(struct CfValue *cv, const char *val);
 #define DEFER_OPS {set_defer_accept, cf_get_int}
+static bool set_scram_iterations(struct CfValue *cv, const char *val);
+#define SCRAM_ITERATIONS_OPS {set_scram_iterations, cf_get_int}
 
 static const struct CfLookup auth_type_map[] = {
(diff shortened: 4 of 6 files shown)


AndrewJackson2020:
> This might be a little controversial, happy to discuss.

Not controversial at all IMO. SCRAM is a known CPU hot bottleneck which has become more unfortunate as it has become the default auth method. Also the maintainers have mentioned this [0] as a potential path forward.

Might be helpful to provide output from pgbench or maybe some CPU flamegraphs to compare with the issue below.

Thanks for working on this.

[0] https://github.com/pgbouncer/pgbouncer/issues/1335#issuecomment-3001973120

JelteF:
Yeah definitely not controversial. But I vaguely remember that I didn't get anywhere close to the speedups you're getting when I tried out a similar implementation (which I then didn't pursue further). Could you share the script/approach you used to measure these speedups?

mble:
Yeah absolutely - this came up when I was looking at adding USDT probes.

Headline overview:

| Host | Architecture | master TPS | patch TPS | Speedup | master connection time | patch connection time |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| Rocky VM | aarch64 | 231.87 | 1,566.56 | 6.76x | 30.359 ms | 4.431 ms |
| Aerial | x86-64 | 406.11 | 2,383.32 | 5.87x | 18.744 ms | 3.260 ms |

- `Rocky`: Rocky Linux 10.2 under OrbStack, 8 aarch64 vCPUs, GCC 14.3.1,
  OpenSSL 3.5.5, perf 6.12.0.
- `Aerial`: Fedora 44, AMD Ryzen AI MAX+ 395 (16 cores/32 threads), GCC 16.2.1,
  OpenSSL 3.5.7, perf 7.1.9.

```sh
PGPASSWORD=bench pgbench \
  -h 127.0.0.1 -p 57322 -U bench -d bench \
  -n -C -c 16 -j 8 -T 30 -P 5 -f select.sql
```

(`select.sql` is just `SELECT 1;`).

I used `auth_type=plain` to exercise `scram_verify_plain_password()` and derivation
on each login.

`select.sql` contains `SELECT 1;`. `perf record` sampled only the PgBouncer
process at 997 Hz with DWARF call graphs. Aerial used `cycles:u`. The Rocky VM
does not expose a supported cycles event, so perf used `task-clock:u` there.
Both captures had zero lost samples. Folded stacks were produced with Brendan
Gregg's FlameGraph scripts; differential graphs were normalized because the two
runs can contain slightly different sample totals.

I've attached some flamegraphs, where the diff is represented by blue shading:

aerial-patch-vs-master-diff

rocky-patch-vs-master-diff

And flamegraphs for each:

rocky-master-flamegraph

rocky-patch-flamegraph

aerial-master-flamegraph

aerial-patch-flamegraph

Using just `Aerieal` as the example for `pgbench`:

Before:

```
pgbench (18.6)
progress: 5.0 s, 403.5 tps, lat 20.777 ms stddev 3.760, 0 failed
progress: 10.0 s, 405.6 tps, lat 20.779 ms stddev 3.981, 0 failed
progress: 15.0 s, 405.0 tps, lat 20.752 ms stddev 3.709, 0 failed
progress: 20.0 s, 407.1 tps, lat 20.563 ms stddev 3.407, 0 failed
progress: 25.0 s, 406.5 tps, lat 20.484 ms stddev 3.289, 0 failed
progress: 30.0 s, 407.5 tps, lat 20.499 ms stddev 3.404, 0 failed
transaction type: /home/mble/tmp/pr1563-flamegraphs-20260823/run-master/select.sql
scaling factor: 1
query mode: simple
number of clients: 16
number of threads: 8
maximum number of tries: 1
duration: 30 s
number of transactions actually processed: 12191
number of failed transactions: 0 (0.000%)
latency average = 20.635 ms
latency stddev = 3.616 ms
average connection time = 18.744 ms
tps = 406.105798 (including reconnection times)
```

After:

```
pgbench (18.6)
progress: 5.0 s, 2386.2 tps, lat 3.457 ms stddev 0.373, 0 failed
progress: 10.0 s, 2382.1 tps, lat 3.473 ms stddev 0.395, 0 failed
progress: 15.0 s, 2382.8 tps, lat 3.462 ms stddev 0.361, 0 failed
progress: 20.0 s, 2377.3 tps, lat 3.451 ms stddev 0.293, 0 failed
progress: 25.0 s, 2387.3 tps, lat 3.431 ms stddev 0.241, 0 failed
progress: 30.0 s, 2382.9 tps, lat 3.443 ms stddev 0.376, 0 failed
transaction type: /home/mble/tmp/pr1563-flamegraphs-20260823/run-patch/select.sql
scaling factor: 1
query mode: simple
number of clients: 16
number of threads: 8
maximum number of tries: 1
duration: 30 s
number of transactions actually processed: 71507
number of failed transactions: 0 (0.000%)
latency average = 3.453 ms
latency stddev = 0.345 ms
average connection time = 3.260 ms
tps = 2383.320787 (including reconnection times)
```

And for good measure, the raw perf captures:

[aerial-master-perf-report.txt](https://github.com/user-attachments/files/31354176/aerial-master-perf-report.txt)
[aerial-patch-pgbench.txt](https://github.com/user-attachments/files/31354177/aerial-patch-pgbench.txt)

PR #1571, opened 2026-08-31
Export pgbevent.dll entry points via a .def file

petere:
pgbevent_LINK passed -Wl,--export-all-symbols -Wl,--add-stdcall-alias. clang (or rather lld) has no --add-stdcall-alias, so linking with clang fails when using the make build system.

Neither option was needed.  --export-all-symbols only matters when there are explicit exports to override, and there were none. --add-stdcall-alias only did anything on 32-bit x86.

To improve this, list the two entry points in a .def file instead.  It states the intent more clearly, reduces the exports from "everything" to only what we need, and will also work with possible MSVC support in the future.

While here, build pgbevent as a shared_module in the meson build, which is the more correct choice.

---

I borrowed a bit from #1542, but I arrived at this from the other end of getting clang to work with the makefiles.  (Which may be partially obsolete, but as long as it's there, we might as well clean it up, especially if it also helps other possible efforts.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,9 +119,9 @@ bin_PROGRAMS += pgbevent
 endif
 
 pgbevent_SOURCES = win32/pgbevent.c win32/eventmsg.rc \
-		   win32/eventmsg.mc win32/MSG00001.bin
+		   win32/eventmsg.mc win32/MSG00001.bin win32/pgbevent.def
 pgbevent_EXT = .dll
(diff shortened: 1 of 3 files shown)


JelteF:
The windows build is failing. I restarted it incase it's a fluke, but it seemed very broken. So I guess this is incorrect somehow. Other than that the idea seems reasonable. Feel free to merge whenever you want once CI is green.

PR #1572, opened 2026-08-31
Add pam_start_confdir setting and add tests for PAM auth

AndrewJackson2020:
This commit adds test coverage over PAM authentication. In order to do this without root access to the system it makes use of pam_start_confdir [0] function that is avaiable in linux-pam from 1.4.0 [1]. Additionally it adds an auth_pam_confdir in order to allow the confdir parameter to pam_start_confdir to be exposed via pgbouncer configuration. Older non EOL'd linux distros (rocky8) for example, do not support pam_start_confdir. Besides being useful for pgbouncer testing switching to pam_start_confdir is potentially also helpful for end users who want to use PAM but also are not able to have root permissions over their systems.

Much of this work is based on my patch to add similar testing into postgres [2]. Much of the code and methodology is shared.

[0] https://man7.org/linux/man-pages/man3/pam_start.3.html
[1] https://github.com/linux-pam/linux-pam/blob/fb8a69f353b33ef14e36090bc77d0a2007de85a8/NEWS#L192
[2] https://www.postgresql.org/message-id/flat/CAKK5BkHsWCMLS4B8N=9juny1kCzkoHFXHkesEA=-ng4pRAHdUw@mail.gmail.com

### DIFF
--- a/.github/workflows/pgbouncer-ci.yml
+++ b/.github/workflows/pgbouncer-ci.yml
@@ -226,6 +226,7 @@ jobs:
           useradd -m user
           chown -R user .
           echo 'user ALL=(ALL) NOPASSWD: ALL' >> /etc/sudoers
+          echo -e 'auth required pam_permit.so\naccount required pam_permit.so' > /etc/pam.d/pgbouncer
       - name: Build
         run: su user -c "MESON_ARGS='$MESON_ARGS' CONFIGURE_ARGS='$CONFIGURE_ARGS' PATH=/venv/bin:\$PATH sh ci/run.sh build $BUILD_SYSTEM"
       - name: Test
--- a/meson.build
+++ b/meson.build
@@ -381,7 +381,12 @@ endif
 if pam.found()
   cdata.set('HAVE_PAM', 1, description: 'PAM support.')
   cdata.set('HAVE_SECURITY_PAM_APPL_H', 1)
+
+
 endif
+
+cdata.set('HAVE_' + 'pam_start_confdir'.to_upper(),
+          cc.has_function('pam_start_confdir', dependencies: pam) ? 1 : false)
 
 # ----------------------------------------------------------------------
 # LDAP (mirrors --with-ldap)
--- a/meson.build
+++ b/meson.build
@@ -381,8 +381,6 @@ endif
 if pam.found()
   cdata.set('HAVE_PAM', 1, description: 'PAM support.')
   cdata.set('HAVE_SECURITY_PAM_APPL_H', 1)
-
-
 endif
 
 cdata.set('HAVE_' + 'pam_start_confdir'.to_upper(),
--- a/src/main.c
+++ b/src/main.c
@@ -210,6 +210,9 @@ char *cf_server_tls13_ciphers;
 int cf_max_prepared_statements;
 
 int cf_scram_iterations;
+#ifdef HAVE_PAM_START_CONFDIR
+char *cf_auth_pam_confdir;
+#endif
 
 /*
  * config file description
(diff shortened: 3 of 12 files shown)


ardentperf:
I think this is actually two changes combined into one commit
1. adding support for auth_pam_confdir (with requisite autoconf and meson code)
2. adding tests, based on that

the detection of auth_pam_confdir actually seems more straightforward than i expected

i dont see any issues

meson.build
```diff
@@ -381,8 +381,13 @@ endif
 if pam.found()
   cdata.set('HAVE_PAM', 1, description: 'PAM support.')
   cdata.set('HAVE_SECURITY_PAM_APPL_H', 1)
+
+
```
JelteF: unnecessary change
```suggestion
```

AndrewJackson2020: Apologies for that. I'll clean this PR up by tommorow.

AndrewJackson2020:
> I think this is actually two changes combined into one commit
> 
> 1. adding support for auth_pam_confdir (with requisite autoconf and meson code)
> 2. adding tests, based on that
> 
> the detection of auth_pam_confdir actually seems more straightforward than i expected
> 
> i dont see any issues

Agree that it is 2 changes, one of which is an entirely new feature. Not ideal but I think this is the cleanest way of implementing tests for PAM.

meson.build
AndrewJackson2020: This has been fixed. Unfortunately the tests are all failing because of this [0] issue caused by the latest psycopg release. Will rebase and make sure everything is good here after the fix is merged.

[0] https://github.com/pgbouncer/pgbouncer/pull/1575

src/main.c
```diff
@@ -369,6 +372,9 @@ static const struct CfKey bouncer_params [] = {
 	CF_ABS("user", CF_STR, global_username, CF_NO_RELOAD, NULL),
 #endif
 	CF_ABS("verbose", CF_INT, cf_verbose, 0, NULL),
+#ifdef HAVE_PAM_START_CONFDIR
+	CF_ABS("auth_pam_confdir", CF_STR, cf_auth_pam_confdir, 0, NULL),
+#endif
```
cosgroveb: [This belongs](https://github.com/pgbouncer/pgbouncer/blob/51471315d026673a14b7219b7deebf33bf3eefd6/src/main.c#L265) around line 275.

` * Add new parameters in alphabetical order. This order is used by SHOW CONFIG.`

.github/workflows/pgbouncer-ci.yml
cosgroveb, line 215:
```suggestion
      CONFIGURE_ARGS: --enable-cassert --with-systemd --without-cares --with-pam
```

... maybe? not sure it's worth it.

AndrewJackson2020: I just put the pam pgbouncer file creation in the wrong place entirely. Should be with the pam pipeline.

Tbh I'm not even sure if this test should be included in the PR. It is possible to implement without the c code changes for pam_start_confdir that I did which make this PR not entirely about testing. I'm tempted to pull this out and move to a different PR, there's already a lot going on in this PR without that happy path test. Also it requires the exact file, character for character, be present on order to run the test which is a bit hacky imo.

src/main.c
AndrewJackson2020: Nice catch, ty for the review.

cosgroveb (approved):
lgtm except the one ordering thing in main.c

.github/workflows/pgbouncer-ci.yml
cosgroveb: If you're unsure about it don't take this suggested diff. This test is still exercised. All good..

AndrewJackson2020: I see what you mean. `linux-debian` builds pam/ldap for meson but not autoconf because meson build options default to "prefer if dependencies found". I wonder if this was intended when the meson build system was introduced.

AndrewJackson2020: I moved the standalone system pam test over to #1619. I figure it makes more sense to keep these 2 tests as separate as they only share maybe 15 lines of code across 3 files and it doesn't have the baggage of requiring changes to the C code in order to get some basic testing of PAM off of the ground.

src/main.c
AndrewJackson2020: This is fixed now. 

I also went ahead and raised #1618 in order provide an assertion that `SHOW CONFIG` is alphabetically sorted. Turned out one of the entries wasn't. It appears this has already been merged.

Thanks again for the review.

PR #1573, opened 2026-09-01
Track default_transaction_read_only in the varcache

hlinnaka:
Before this commit, if a client did `SET default_transaction_read_only = true` in transaction mode, the connection was returned to the pool in read only mode, causing subsequent write transaction to fail. This is a well-known problem, and the common work around is to add `default_transaction_read_only` to `track_extra_parameters`. However, we can easily fix it for everyone by adding it to the static list of variables that are always tracked.

`default_transaction_read_only` has been marked as `GUC_REPORT` since PostgreSQL 14. In older server versions, we still cannot track it, but that's OK.

### DIFF
--- a/test/test_misc.py
+++ b/test/test_misc.py
@@ -461,9 +461,6 @@ def test_track_extra_parameters(bouncer):
         "application_name": ["client1", "client2"],
     }
 
-    if not WINDOWS:
-        test_set["client_encoding"] = ["LATIN1", "LATIN5"]
-
     test_expected = {
         "intervalstyle": ["sql_standard", "postgres"],
         "standard_conforming_strings": ["on", "off"],
@@ -473,7 +470,13 @@ def test_track_extra_parameters(bouncer):
     }
 
     if not WINDOWS:
+        test_set["client_encoding"] = ["LATIN1", "LATIN5"]
         test_expected["client_encoding"] = ["LATIN1", "LATIN5"]
+
+    # default_transaction-read_only was not marked as GUC_REPORT until server version 14
+    if PG_MAJOR_VERSION >= 14:
+        test_set["default_transaction_read_only"] = ["true", "false"]
+        test_expected["default_transaction_read_only"] = ["on", "off"]
 
     with bouncer.cur(dbname="p1") as cur1, bouncer.cur(dbname="p1") as cur2:
         for key in test_set:
@@ -486,11 +489,17 @@ def test_track_extra_parameters(bouncer):
             cur1.execute(stmt)
             cur2.execute(stmt)
 
-            result1 = cur1.fetchone()
-            assert result1[0] == test_expected[key][0]
+            result1 = cur1.fetchone()[0]
+            expected1 = test_expected[key][0]
+            assert result1 == expected1, (
+                f'parameter {key}, expected "{expected1}", got "{result1}"'
+            )
 
-            result2 = cur2.fetchone()
-            assert result2[0] == test_expected[key][1]
+            result2 = cur2.fetchone()[0]
+            expected2 = test_expected[key][1]
+            assert result2 == expected2, (
+                f'parameter {key}, expected "{expected2}", got "{result2}"'
+            )
 
 
 async def test_wait_close(bouncer):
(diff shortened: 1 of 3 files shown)


hlinnaka:
I'm not sure what the policy is with tests and older PostgreSQL server versions. Do we need to cater for running the tests against server versions older than 14?

hlinnaka:
This PR adds only `default_transaction_read_only` to the list, but should we do the same for `search_path` and `IntervalStyle`?

`IntervalStyle` is in `track_extra_parameters` by default, but is there any reason someone would want to _remove_ it?

`search_path` is marked as GUC_REPORT since PostgreSQL version 18. Adding it to the list of variables to track won't have any effect on older server versions, but seems like we should track it when we can. Is there any harm in doing so?

test/test_misc.py
```diff
@@ -486,11 +489,17 @@ def test_track_extra_parameters(bouncer):
...
+                f'parameter {key}, expected "{expected1}", got "{result1}"'
+            )
 
-            result2 = cur2.fetchone()
-            assert result2[0] == test_expected[key][1]
+            result2 = cur2.fetchone()[0]
+            expected2 = test_expected[key][1]
+            assert result2 == expected2, (
+                f'parameter {key}, expected "{expected2}", got "{result2}"'
+            )
```
hlinnaka, line 502: These changes just add some context if the assertions fail. Before this, you'd just get an error like `on != off`. Now the failure tells you which parameter didn't match.

hlinnaka:
The CI is failing on the `test_discard_all` test. But I believe it's unrelated to the changes here, because I'm getting the same failure on a branch where I reverted all changes, making it equal to `master` branch: https://github.com/hlinnaka/pgbouncer/actions/runs/33501598030/job/99836070117. Weird failure..

hlinnaka:
> The CI is failing on the `test_discard_all` test. But I believe it's unrelated to the changes here, because I'm getting the same failure on a branch where I reverted all changes, making it equal to `master` branch: https://github.com/hlinnaka/pgbouncer/actions/runs/33501598030/job/99836070117. Weird failure..

Ok, that was https://github.com/pgbouncer/pgbouncer/issues/1574, unrelated to this PR.

JelteF:
At first glance this makes sense, but I need to think about how this interacts with this PR #1534: What makes `default_transaction_read_only` special when compared to other `GUC_REPORT` settings that are changeable by the user is that a server can change it on an existing connection if it turns into a standby. That probably means it could change mid-transaction for a connection. And I think that would cause our varcache logic to restore it even on read-write connections for that transaction with this change.

I'd love your thoughts on that problem, and also #1534 in general. 

> IntervalStyle is in track_extra_parameters by default, but is there any reason someone would want to remove it?

I cannot think of a reason. I don't remember why I didn't just add it to the list. It might have been to give the `track_extra_parameters` setting some test coverage.

> search_path is marked as GUC_REPORT since PostgreSQL version 18. Adding it to the list of variables to track won't have any effect on older server versions, but seems like we should track it when we can. Is there any harm in doing so?

Maybe, I seem to remember the varcache logic behaved strangely/buggy if there was no ParameterStatus at all for the setting. But I'm not exactly sure anymore in what way. And whatever it was it might be (easily) fixable too.

> I'm not sure what the policy is with tests and older PostgreSQL server versions. Do we need to cater for running the tests against server versions older than 14?

Whatever the lowest PG version is that we run in CI is the thing we officially support.

hlinnaka:
Thanks for looking!

> At first glance this makes sense, but I need to think about how this interacts with this PR #1534: What makes `default_transaction_read_only` special when compared to other `GUC_REPORT` settings that are changeable by the user is that a server can change it on an existing connection if it turns into a standby. That probably means it could change mid-transaction for a connection. And I think that would cause our varcache logic to restore it even on read-write connections for that transaction with this change.

Hmm, I don't think that's true; `default_transaction_read_only` is not automatically set on standbys. `transaction_read_only` is (see https://github.com/postgres/postgres/blob/348f4ab53a080af611c92ad638781500ac53fe09/src/backend/access/transam/xact.c#L2155), but that's different.

(Postgres doesn't support turning an existing server from primary to a standby, but standby -> primary is possible. Turning a primary back into a standby would be a nice feature though, so it's good to keep that in mind too..)

JelteF (approved):
Hmm, okay that does make it much more straight forward then.

JelteF:
> Maybe, I seem to remember the varcache logic behaved strangely/buggy if there was no ParameterStatus at all for the setting. But I'm not exactly sure anymore in what way. And whatever it was it might be (easily) fixable too.

I figured out what this was again and created a PR to fix it in #1576

I don't think this PR needs to be blocked on that though, since it would only affect PG13 and below which are all EOL. And for `default_transaction_read_only` it doesn't seem as huge of a problem as it is for `search_path`.

PR #1575, opened 2026-09-01
Fix 'test_discard_all' failure with psycopg 3.3.5

### LINKED ISSUE #1574: test_discard_all fails with psycopg 3.3.5
psycopg 3.3.5 includes a [change](https://github.com/psycopg/psycopg/pull/1307) to catch `DISCARD ALL` commands issued by the client, making it re-prepare any prepared statements. Pgbouncer's `test_discard_all` test depends on the old behavior, causing the test to fail with psycopg 3.3.5. See failure at https://github.com/pgbouncer/pgbouncer/actions/runs/33500134938

hlinnaka:
psycopg 3.3.5 includes a change to catch DISCARD ALL commands issued by the client, making it re-prepare any prepared statements. The test_discard_all test depended on the old behavior, causing the test to fail with psycopg 3.3.5. Silence the failure by removing the query that behaved differently on psycopg 3.3.5 and older versions.

Fixes #1574. See https://github.com/psycopg/psycopg/pull/1307 for the psycopg change.

### DIFF
--- a/test/test_prepared.py
+++ b/test/test_prepared.py
@@ -86,13 +86,13 @@ def test_discard_or_deallocate_all(bouncer, command):
 
         # Confirm that the prepared query is not available anymore on
         # client 1
-        result = conn1.pgconn.exec_prepared(b"mystmt", ())
+        with bouncer.log_contains("prepared statement did not exist"):
+            result = conn1.pgconn.exec_prepared(b"mystmt", ())
         assert result.status == pq.ExecStatus.FATAL_ERROR
         assert (
             b"prepared statement did not exist" in result.error_message
             or b"server closed the connection unexpectedly" in result.error_message
         )
-        assert bouncer.log_contains("prepared statement did not exist")
 
 
 def test_parse_larger_than_pkt_buf(bouncer):
--- a/test/test_prepared.py
+++ b/test/test_prepared.py
@@ -94,19 +94,17 @@ def test_discard_all(bouncer):
         # execute DISCARD ALL on client 1
         cur1.execute("DISCARD ALL")
 
-        # Run the prepared query again on server 2 and client 2
+        # Run the prepared query again on client 2
         cur2.execute(prepared_query)
 
-        # Confirm that the prepared query is not available anymore on
-        # client 1
-        with (
-            bouncer.log_contains("prepared statement did not exist"),
-            pytest.raises(
-                psycopg.OperationalError,
-                match="prepared statement did not exist|server closed the connection unexpectedly",
-            ),
-        ):
-            cur1.execute(prepared_query)
+        # Depending on the psycopg version, the statement may or may not work on
+        # client 1. Starting with psycopg 3.3.5, it recognizes the "DISCARD"
+        # command, resets its prepared statement cache, and prepares the
+        # statement again on the next call. In earlier versions, you get a
+        # "prepared statement did not exist" error. We're not trying to test
+        # psycopg here, so skip it.
+        #
+        # cur1.execute(prepared_query)
 
 
 def test_parse_larger_than_pkt_buf(bouncer):
--- a/test/test_prepared.py
+++ b/test/test_prepared.py
@@ -49,7 +49,14 @@ def test_prepared_statement_params(bouncer):
             cur2.execute(prepared_query, params=(1,), prepare=True)
 
 
-def test_deallocate_all(bouncer):
+@pytest.mark.parametrize("command", ["DISCARD ALL", "DEALLOCATE ALL"])
+def test_discard_or_deallocate_all(bouncer, command):
+    """
+    Client sends a "DEALLOCATE ALL" command (or "DISCARD ALL", which has the
+    same effect). pgbouncer recognizes it and resets the client's prepared
+    statement cache. This scenario is implemented with raw libpq calls to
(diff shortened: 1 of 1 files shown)


AndrewJackson2020:
I ran into this as well with a feature I'm working on. Seems not ideal to just comment out parts of tests. I tried to "trick" psyopg by defining a function that calls `DISCARD ALL` to avoid triggering the regex in psycopg but postgres doesn't allow that.

hlinnaka:
> I ran into this as well with a feature I'm working on. Seems not ideal to just comment out parts of tests.

That's fair. I thought it's not interesting to test psycopg's own prepared statement caching, but it would still be good to test the underlying pgbouncer behavior.

>  I tried to "trick" psyopg by defining a function that calls `DISCARD ALL` to avoid triggering the regex in psycopg but postgres doesn't allow that.

Heh, yeah I tried that too :-). 

I added a commit to this PR to rewrite the test using raw libpq calls, to bypass psycopg'ts prepared statement cache. What do you think?

While at it, I did the same to `test_discard_all`. It wasn't failing, but I think that was just an oversight in psycopg; it surely should recognize `DEALLOCATE ALL` as well as `DISCARD ALL`.

AndrewJackson2020:
> I added a commit to this PR to rewrite the test using raw libpq calls, to bypass psycopg'ts prepared statement cache. What do you think?

Did you push the commit? Not seeing the raw libpq calls.  Last commit just parameterizes the 2 test which are very duplicative functions (good catch). I think raw libpq calls is probably a lot better than any solution I have tried.

Also tried this
```sql
DO $$
BEGIN
    EXECUTE 'DISCAR' || 'D ALL';
END$$;
```

with the same error message `DISCARD ALL cannot be executed from a function`.

AndrewJackson2020:
This seems to work for me, granted accessing "private" functions/attributes is considered "not good" in python as they are prone to interface changes. At a minimum I think that there should be comment explaining why we need to use `_pgconn` instead of using the public interface.

```diff
diff --git a/test/test_prepared.py b/test/test_prepared.py
index 852d2de..4dbc5b7 100644
--- a/test/test_prepared.py
+++ b/test/test_prepared.py
@@ -92,7 +92,7 @@ def test_discard_all(bouncer):
         cur2.execute(prepared_query, prepare=True)

         # execute DISCARD ALL on client 1
-        cur1.execute("DISCARD ALL")
+        cur1._pgconn.exec_("DISCARD ALL".encode())

         # Run the prepared query again on server 2 and client 2
         cur2.execute(prepared_query)
```

hlinnaka:
Oh sorry, I indeed pushed wrong commit. Fixed

hlinnaka:
> This seems to work for me, granted accessing "private" functions/attributes is considered "not good" in python as they are prone to interface changes. 

Agreed, but there are many other tests in the file that are already using the same private functions, so it seems fine to do it here too.

> At a minimum I think that there should be comment explaining why we need to use `_pgconn` instead of using the public interface.
> 
> ```diff
> diff --git a/test/test_prepared.py b/test/test_prepared.py
> index 852d2de..4dbc5b7 100644
> --- a/test/test_prepared.py
> +++ b/test/test_prepared.py
> @@ -92,7 +92,7 @@ def test_discard_all(bouncer):
>          cur2.execute(prepared_query, prepare=True)
> 
>          # execute DISCARD ALL on client 1
> -        cur1.execute("DISCARD ALL")
> +        cur1._pgconn.exec_("DISCARD ALL".encode())
> 
>          # Run the prepared query again on server 2 and client 2
>          cur2.execute(prepared_query)
> ```

Yeah, we could do that. But I chose to replace all the prepare and execute steps in the test with raw libpq calls too. That way, it's not dependent on psycopg's prepared statement cache at all. Not wedded to that though.

test/test_prepared.py
```diff
@@ -49,64 +49,50 @@ def test_prepared_statement_params(bouncer):
...
-            ),
-        ):
-            cur1.execute(prepared_query)
+        result = conn1.pgconn.exec_prepared(b"mystmt", ())
+        assert result.status == pq.ExecStatus.FATAL_ERROR
+        assert (
+            b"prepared statement did not exist" in result.error_message
+            or b"server closed the connection unexpectedly" in result.error_message
+        )
+        assert bouncer.log_contains("prepared statement did not exist")
```
JelteF: This should be used as a `with` block:

```python
with bouncer.log_contains("prepared statement did not exist"):
    result = conn1.pgconn.exec_prepared(b"mystmt", ())
```

I think now the assertion is always true

JelteF:
Thanks for addressing this, approach looks good.

dvarrazzo:
Hello, I noticed this issue by following the comment @hlinnaka left in https://github.com/psycopg/psycopg/pull/1307.

Anything else we should take care of to coordinate?

Something else: we have [workflows to test 3rd party systems](https://github.com/psycopg/psycopg/blob/master/.github/workflows/3rd-party-tests.yml) with psycopg: maybe it would be useful to add pgbouncer (a selection of version and main operational modes) to our grid?

Thank you very much!

JelteF:
@dvarrazzo There's no actual bug here (neither in pgbouncer, nor psycopg, nor when combined), it was just that the feature you added interacted unexpectedly with one of our tests assertions and that made it fail.

PR #1576, opened 2026-09-01
Correctly handle tracked parameters that are not reported by the server

JelteF:
It's possible to configure PgBouncer such that a parameter is tracked in the varcache, while the backing postgres server doesn't actually report it using ParameterStatus. This would cause buggy behaviour when the client set such a parameter in its startup packet. This fixes that.

This becomes particularly relevant now that PG18 has made search_path a `GUC_REPORT` setting. This is very useful for PgBouncer and  we'd like to track `search_path` by default, but that shouldn't cause bugs on older versions.

In passing fixes a small out-of-bounds write of a null byte.

Related to #1573

### DIFF
--- a/src/varcache.c
+++ b/src/varcache.c
@@ -191,7 +191,8 @@ static int apply_var(PktBuf *pkt, const char *key,
 		/*
 		 * This happens when a parameter is in track_extra_parameters but
 		 * PostgreSQL does not report it in ParameterStatus (e.g. enable_seqscan,
-		 * or default_transaction_read_only on PG <= 13), so pool->orig_vars is NULL.
+		 * or search_path on PG <= 17), so the parameter is not set in
+		 * pool->orig_vars.
 		 * A previous client configured this setting in its StartupMessage / connection
 		 * options (so sval is non-NULL), and now a client without this startup
 		 * parameter connects (cval is NULL).
--- a/src/varcache.c
+++ b/src/varcache.c
@@ -173,17 +173,43 @@ static int apply_var(PktBuf *pkt, const char *key,
 	unsigned len;
 	const char *tmp;
 
-	/* if unset, skip */
-	if (!cval || !sval)
+	/* if both are unset, skip */
+	if (!cval && !sval)
 		return 0;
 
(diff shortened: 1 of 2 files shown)


src/varcache.c
```diff
@@ -165,17 +165,43 @@ static int apply_var(PktBuf *pkt, const char *key,
...
+		/* ignore case difference */
+		if (strcasecmp(cval->str, sval->str) == 0)
+			return 0;
+	}
+
+	if (!cval) {
+		/*
+		 * This happens when a parameter is in track_extra_parameters but
+		 * PostgreSQL does not report it in ParameterStatus (e.g. enable_seqscan,
+		 * or default_transaction_read_only on PG <= 13), so pool->orig_vars is NULL.
```
hlinnaka:
```suggestion
		 * or search_path on PG <= 17), so pool->orig_vars is NULL.
```
Few people care about `default_transaction_read_only` on old versions, but `search_path` is still relevant for a few years.

hlinnaka: IIUC `pool->orig_vars != NULL` in any case, it just won't contain the given parameter. Maybe rephrase that.

hlinnaka (approved):
Hmm, so the scenario this fixes is:
- client connects with `enable_seqscan=off` in startup parameters
- pgbouncer calls `set enable_seqscan=off` on the server connection
- new client connects, with no startup parameters
- the server connection with `enable_seqscan=off` is reused for it

You still have the same problem if the client does `SET enable_seqscan=off` in the connection, rather than specifying it in the startup packet. But there isn't much we can do about that; that's why `server_reset_query_always` exists.

Yeah, this is an improvement, I don't see any downside. Unless you want to argue that it'd be better to be consistent and have the same broken behavior with both `SET` and startup params.

JelteF:
> But there isn't much we can do about that; 

Yeah that's something that needs to happen at postgres. I'd love to be able to disable certain SQL features (`SET`, `PREPARE`, `LISTEN`) in Postgres using some GUC so that people don't shoot themselves in the foot. But that doesn't exist currently.

PR #1577, opened 2026-09-01
Prevent segfaults when a replication client disconnects

kevinwatson:
# Summary

Prevent a pgbouncer segfault when a `replication=database` client disconnects while it still has a replication-protocol simple query in flight.

## Fix

- After each sbuf_call_proto() callback (the packet-header callback around line 815, and the SBUF_EV_PKT_CALLBACK case around line 842), the local io pointer is now re-read from sbuf->io.
- If the callback closed the socket (which frees the IO buffer and NULLs sbuf->io), the function now returns false instead of continuing to dereference the stale io pointer.

## Doc 

Flowchart documenting the logical branch that leads to the error: https://github.com/kevinwatson/pgbouncer-client-disconnect-segfault/blob/main/CRASH.md#1-the-branch-that-kills-it

# Testing

Test app to reproduce the issue:

https://github.com/kevinwatson/pgbouncer-client-disconnect-segfault

### DIFF
--- a/src/sbuf.c
+++ b/src/sbuf.c
@@ -816,6 +816,11 @@ static bool sbuf_process_pending(SBuf *sbuf)
 				goto need_more_data;
 			}
 			Assert(sbuf->pkt_remain > 0);
+
+			/* callback may have closed sbuf and freed io */
+			io = sbuf->io;
+			if (!io)
+				return false;
 		}
 
 		if (sbuf->pkt_action == ACT_SKIP || sbuf->pkt_action == ACT_CALL) {
@@ -837,6 +842,10 @@ static bool sbuf_process_pending(SBuf *sbuf)
 			if (!sbuf_call_proto(sbuf, SBUF_EV_PKT_CALLBACK)) {
 				goto need_more_data;
 			}
+			/* callback may have closed sbuf and freed io */
+			io = sbuf->io;
+			if (!io)
+				return false;
 		/* fallthrough */
 		/* after callback, skip pkt */
 		case ACT_SKIP:


eulerto:
I reproduced it with the following test after building with ASAN. The oversight was introduced at commit e1c6c2f.

```
diff --git a/test/test_replication.py b/test/test_replication.py
index cc3ce70..4f12a55 100644
--- a/test/test_replication.py
+++ b/test/test_replication.py
@@ -1,5 +1,7 @@
 import asyncio
 import signal
+import socket
+import struct
 import subprocess
 import time
 
@@ -296,3 +298,63 @@ async def test_replication_pool_size_mixed_clients(bouncer):
     # eviction of one of the normal connections
     with bouncer.log_contains("closing because: evicted"):
         bouncer.test(**connect_args, replication="database")
+
+
+def _open_replication_client(bouncer, user, dbname):
+    """Log in over a raw socket with replication=database and return it."""
+    sock = socket.create_connection((bouncer.host, bouncer.port))
+    sock.settimeout(5)
+    params = {"user": user, "database": dbname, "replication": "database"}
+    body = b"".join(k.encode() + b"\0" + v.encode() + b"\0" for k, v in params.items())
+    body += b"\0"
+    sock.sendall(struct.pack("!II", len(body) + 8, 196608) + body)
+    received = b""
+    # ReadyForQuery: 'Z', int32 length 5, transaction status byte
+    while b"Z\x00\x00\x00\x05" not in received:
+        chunk = sock.recv(65536)
+        assert chunk, f"connection closed during login, got {received!r}"
+        received += chunk
+    return sock
+
+
+def _send_simple_query(sock, query):
+    body = query.encode() + b"\0"
+    sock.sendall(b"Q" + struct.pack("!I", len(body) + 4) + body)
+
+
+def _reset_connection(sock):
+    """Close the socket with an RST, like a process that was killed."""
+    sock.setsockopt(socket.SOL_SOCKET, socket.SO_LINGER, struct.pack("ii", 1, 0))
+    sock.close()
+
+
+def test_replication_clients_reset_with_query_in_flight(bouncer):
+    """
+    Regression test for PR #1577.
+
+    A replication client dedicated server connection finishes its login in
+    the SBUF_EV_READ callback of the server sbuf. That callback activates the
+    waiting client, which makes PgBouncer read from the client socket right
+    away. If the client already sent its first query and then died with a
+    reset, the read fails and the client is disconnected. Because the server
+    is ready but still has unflushed data, the disconnect also closes the
+    server, and frees the io buffer that sbuf_process_pending() is still
+    using. That function used to keep using its cached pointer to the freed
+    buffer and crashed with a segfault.
+
+    The window is small, so open several clients and reset them all at once,
+    repeatedly.
+    """
+    for _ in range(60):
+        socks = [
+            _open_replication_client(bouncer, "postgres", "user_passthrough")
+            for _ in range(4)
+        ]
+        time.sleep(0.05)
+        for sock in socks:
+            _send_simple_query(sock, "IDENTIFY_SYSTEM")
+        for sock in socks:
+            _reset_connection(sock)
+
+        # PgBouncer must still be alive
+        bouncer.admin("SHOW VERSION")
```

eulerto (approved):

PR #1578, opened 2026-09-02
Add `auth_query_client_addr` option to pass client IP to auth_query

tbe-odoo:
Adds a new `auth_query_client_addr` boolean setting.
When enabled, PGBouncer passes the client's connecting address as a second parameter
(`$2`) to `auth_query`, in addition to the user name (`$1`).

This lets a custom `auth_query` function (e.g. a `SECURITY DEFINER`
function) implement IP-based allow/deny lists for logins, which is
useful in shared/multi-tenant environments where PGBouncer sits in
front of many clients.

Default is `0` (off), so existing `auth_query` configurations are unaffected.

## Changes

- `auth_query_client_addr` config option (`include/bouncer.h`, `src/main.c`)
- `src/client.c`: conditionally binds the client address (via `pga_ntop`)
  as `$2` when the option is enabled
- Documentation in `doc/config.md` and an example in `etc/pgbouncer.ini`
- Tests in `test/test_auth.py` covering both the allow and deny cases

## Test plan

- [x] `pytest test/test_auth.py -k client_addr` - both new tests pass
- [x] Full `pytest test/` suite (waiting confirmation in CI, all green in local apart from the `test_discard_all` that needs https://github.com/pgbouncer/pgbouncer/pull/1575)

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -566,6 +566,20 @@ that no password that the client offers will be accepted.
 
 Default: `SELECT rolname, CASE WHEN rolvaliduntil < now() THEN NULL ELSE rolpassword END FROM pg_authid WHERE rolname=$1 AND rolcanlogin`
 
+### auth_query_client_addr
+
+Whether the client's connecting address is passed to `auth_query` as a
+second query parameter (`$2`), in addition to the user name (`$1`).
+
+This allows an `auth_query` function to implement IP-based allow/deny listing for logins, by inspecting the address and raising an error
+(or returning no rows) for addresses that should not be allowed to
+authenticate.
+
+The address is a numeric IPv4 or IPv6 address, or the literal string
+`unix` for Unix domain socket connections.
+
+Default: 0
+
 ### auth_dbname
 
 Database name in the `[database]` section to be used for authentication purposes. This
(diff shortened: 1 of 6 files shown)


AndrewJackson2020:
This seems very similar to functionality already provided in the hba auth method. What is the use case here that isn't already available with hba?

tbe-odoo:
> This seems very similar to functionality already provided in the hba auth method. What is the use case here that isn't already available with hba?

This is indeed similar to what can be expressed with the HBA authentication method. The difference is how the rules are managed and evaluated.

Our use case involves multiple tenant databases sharing the same PgBouncer instance. Each tenant must be able to manage its own IP allowlist dynamically. A network firewall cannot enforce this policy because PgBouncer uses the same address and port for every tenant, so the firewall cannot determine which PostgreSQL database the client is connecting to.

An HBA configuration could technically enforce these rules, but every tenant change would require regenerating the shared HBA file and reloading PgBouncer. In our case, the allowlists are stored in a database and should be evaluated during authentication.

We also considered using `application_name_add_host`, but `application_name` is client-controlled and is primarily intended for observability. Using a formatted value containing untrusted client data as an authentication input would require fragile parsing and would not provide a reliable security boundary.

By contrast, the proposed `$2` parameter is obtained directly from PgBouncer's `client->remote_addr` and is independent of client-provided PostgreSQL parameters. This allows the authentication mechanism to enforce database-backed IP allowlists in a multi-tenant environment.

PR #1580, opened 2026-09-02
Track all parameters that Postgres reports by default

### LINKED ISSUE #482: not support the startup parameter search_path and extra_float_digits

### LINKED ISSUE #1176: Search_path
How can I use **search_path** in pg_bouncer? I am using Django with pgbouncer and want to use search_path to get a specific schema.
Note: for pgbouncer,  We use a .env file to track the environment variables.

We getting errors like 
unsupported startup parameter: options=-c search_path=new_schema

### LINKED ISSUE #1313: pgbouncer pool_mode=transaction and postgres_fdw causes search_path not reset
The issue can be reproduced using minimal example at https://github.com/jackielii/pgbouncer-fdw-issue/tree/main?tab=readme-ov-file

Steps to reproduce:

1. set up 2 databases: `db1` and `db2`
2. setup pgbouncer that connects to both in `transaction` pooling mode. See [pgbouncer.ini](https://github.com/jackielii/pgbouncer-fdw-issue/blob/main/pgbouncer.ini)
3. connect to `db2` via pgbouncer and creates foreign schema connects to `db1` using postgres_fdw
4. Now open a connection to `db1` via pgbouncer and run select query without schema
5. observe an error occurs because wrong default schema `pg_catalog` is used: `relation "test1" does not exist`

The issue seems to be related to a connection sets the search_path gets reused and search_path is not reverted: [related code in postgres_fdw](https://github.com/postgres/postgres/blob/master/contrib/postgres_fdw/postgres_fdw.c#L3935-L3938)

My current work around is to create two db declarations in [pgbouncer.ini](https://github.com/jackielii/pgbouncer-fdw-issue/blob/separate-server/pgbouncer.ini#L4): which uses `pool_name=session` in the create foreign server statement

JelteF:
Now that #1576 is merged there's basically no downside to tracking parameters anymore (except for a tiny amount of extra memory usage per connection). So let's start tracking all writable GUCs that Postgres sets the GUC_REPORT flag for. Most notably this starts tracking search_path by default.

Fixes #1176
Partially fixes #482
Fixes #1313
Related to #1573

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -237,22 +237,38 @@ Default: 0
 
 ### track_extra_parameters
 
-By default, PgBouncer tracks `client_encoding`, `datestyle`, `timezone`,
-`standard_conforming_strings` and `application_name` parameters per client. To
-allow other parameters to be tracked, they can be specified here, so that
-PgBouncer knows that they should be maintained in the client variable cache and
-restored in the server whenever the client becomes active.
+By default, PgBouncer tracks all the parameters that Postgres reports to the
+client and that can be changed by the client: `application_name`,
+`client_encoding`, `DateStyle`, `default_transaction_read_only`,
+`IntervalStyle`, `scram_iterations`, `search_path`, `session_authorization`,
+`standard_conforming_strings` and `TimeZone`. To allow other parameters to be
+tracked, they can be specified here, so that PgBouncer knows that they should be
+maintained in the client variable cache and restored in the server whenever the
+client becomes active.
 
 If you need to specify multiple values, use a comma-separated list (e.g.
-`search_path, IntervalStyle`)
+`some_extension.setting, other_extension.setting`)
 
-Note: Most parameters cannot be tracked this way. The only parameters that can
-be tracked are ones that Postgres reports to the client. Postgres has
-[an official list of parameters that it reports to the client](https://www.postgresql.org/docs/15/protocol-flow.html#PROTOCOL-ASYNC).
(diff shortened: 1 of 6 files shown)


doc/config.md
```diff
@@ -237,22 +237,38 @@ Default: 0
...
-By default, PgBouncer tracks `client_encoding`, `datestyle`, `timezone`,
-`standard_conforming_strings` and `application_name` parameters per client. To
-allow other parameters to be tracked, they can be specified here, so that
-PgBouncer knows that they should be maintained in the client variable cache and
-restored in the server whenever the client becomes active.
+By default, PgBouncer tracks all the parameters that Postgres reports to the
+client and that can be changed by the client: `application_name`,
+`client_encoding`, `DateStyle`, `default_transaction_read_only`,
+`IntervalStyle`, `scram_iterations`, `search_path`, `session_authorization`,
+`standard_conforming_strings` and `TimeZone`. To allow other parameters to be
```
hlinnaka, line 244: This list is getting long.
```suggestion
client and that can be changed by the client:
- `application_name`
- `client_encoding`
- `DateStyle`
- `default_transaction_read_only`,
- `IntervalStyle`
- `scram_iterations` (since PostgreSQL version 16)
- `search_path` (since PostgreSQL version 18)
- `session_authorization`,
- `standard_conforming_strings`
- `TimeZone`.
 
To allow other parameters to be
```

Issue #882, opened 2023-07-06
Add option to not close server connections when DNS resolution changes


lukmajercak:
# What
Currently, when a host is resolved, pgbouncer checks all server connections for this host and tags connections using IPs not found in the new resolved list as dirty (close_needed). It would be helpful to be able to disable this behavior. This would help avoiding connection churn in scenarios where, for example, the host might return a list of different IP addresses every time (think weighted routing in Route53).

Another option to help with these scenarios would be to provide an admin command like `RELOAD_DNS` which would force host resolution regardless of the host's DNS TTL (`dns_ttl_max`). 

# How
[Draft PR](https://github.com/pgbouncer/pgbouncer/pull/881)

Issue #383, opened 2019-04-10
DNS changes not detected


joekohlsdorf:
I have the issue that pgBouncer doesn't see DNS changes, `SHOW DNS_HOSTS;` shows `ttl=0` and outdated IPs.

Version: `pgbouncer 1.9.0, libevent 2.1.8-stable (epoll), adns: c-ares 1.13.0, tls: LibreSSL 2.6.5`
Also tried the `udns` backend and get the same result.

I cranked up log level to maximum and the only time I see any activity resolving DNS is when the first client connects to pgBouncer.
`dns_max_ttl` is set to 5s. Activating SOA zone checks won't help as my domains are CNAMES to other CNAMES. TTL in DNS is 5s and I have no upstream caching.

In `SHOW LISTS;` the counter `dns_queries` shows a negative number.

The documentation on DNS is quite thin but from what I understand pgBouncer should update DNS every `dns_max_ttl` and invalidate open connections to old IPs. The logs tell me that DNS lookups aren't asynchronous because I don't see lookups every 5s.

Any hints on how I could debug this further? Also, what is the actual expected behavior?
I played with SOA zone checks and see that these get fired every `dns_zone_check_period`.

markokr:
The max_ttl just means cached answer is dropped.  Next request will happen when new connection is needed and that may still take some time.  And when different answer is given then that does not affect other connections as different answer is normal for round-robin DNS.

For fast change detection you can use dns_zone stuff, when SOA serial changes all names under it will be requeried, CNAMEs should work fine.  So new destination is taken into use as fast as possible.

I don't think there is any particular support for dropping old connections based on DNS changes, in case of takeover/failover you are supposed to do it yourself (as in, kill old machine).

PJMODOS:
Well if some IP disappears from the DSN record, we mark everything using that IP as dirty which will clean the connections from pool once they are done with whatever they are doing.

joekohlsdorf:
The problem with doing DNS lookups when needing a new connection is that needing a new connection is rare, after all that is the reason for having a connection pool.

What I am trying to achieve is to redirect traffic from one replica to another by making changes in DNS. I have to reload pgBouncer for it to pick up the updated IPs.
As pointed out there is code to reconnect when IPs change and it works well, I reload the config and this forces a DNS lookup which invalidates connections if IPs have changed.

[...]

Issue #332, opened 2018-10-04
Kubernetes DNS resolution fial


JindrichPilar:
I am having problems trying to use pgbouncer in (Azure) Kubernetes, it cannot resolve DNS.

It works on my local machine using docker-compose without problem.

I used custom alpine based image with [alpine pgbouncer package](https://pkgs.alpinelinux.org/package/edge/community/x86/pgbouncer), I also tried docker compiled package by [Edoburu](https://github.com/edoburu/docker-pgbouncer). Both of them use c-ares for domain name resolution (as mentioned in https://github.com/pgbouncer/pgbouncer/issues/122).

The error message I get is:
```
LOG S-0x5611d3b0df90: my-db/root@(bad-af):0 closing because: server dns lookup failed (age=0)
```

I checked domain resolution using ``host`` command from [bind-tools](https://pkgs.alpinelinux.org/package/edge/main/x86/bind-tools) and the same domain name inside same Kuberentes pod is resolved without an issue.

cpressland:
@JindrichPilar - we ran into a similar thing on Azure which was resolved by forcing the pod to use TCP for DNS instead of UDP. You're probably having similar issues in other pods but maybe haven't noticed yet.

Try adding this to the pod spec:

```
      dnsConfig:
        options:
        - name: use-vc
        - name: single-request-reopen
      dnsPolicy: ClusterFirst
```

We're still seeing a huge volume of issues with Postgres on Azure, but not specific to what you're seeing above.

JindrichPilar:
It started working without any changes on my part, no idea why.

@cpressland  Thank You for your response, today I got to try your changes, but Azure CLI forbids me using dnsConfig with ``` Forbidden: DNSConfig: custom pod DNS is disabled by feature gate```

petere:
So is it working now or not?

JindrichPilar:
It is working. Sorry for not closing the issue.

PierreF:
I've also hit this issue and spend some time finding the cause. In my finding this issue should not happen in production and was due to bad DNS server setup.

tl;dr; I was my minikube environment that had "bad" DNS behavior. It returned ServerFail instead of NXDomain from some DNS query. In my case a fix was to configure a real DNS in the /etc/resolv.conf on the physical server

Long version:

* pgbouncer is configured to connect to "pg-host" which is an IPv4 host.
[...]

PR #1582, opened 2026-09-02
Relaunch a DNS request whose resolver callback never fires (add dns_resolve_timeout)

### LINKED ISSUE #1583: In-process DNS resolver never retries a request whose callback never fires (permanent stall, all backends)
## Summary

If the in-process resolver ever fails to deliver a result for a request — the callback never fires — PgBouncer leaves that request pending **forever**. The database hostname becomes permanently unresolvable and only a process restart clears it, even though a fresh resolver on the same host resolves the name fine. There is no timeout on a pending resolution anywhere.

This is the pgbouncer-side root cause behind the recurring "DNS never recovers / restart fixes it" reports (#1119, #1353, #1125, #1126). It is **independent of the DNS backend** — the retry logic lives in the generic layer, not in the evdns2 / c-ares / libc code — which is why switching backends doesn't reliably help (#1353 reports c-ares never recovering after evdns2 didn't).

## Version

Seen with 1.25.2 (c-ares 1.34.6); the relevant code is unchanged on `master` and the defect is backend-independent.

## Symptom

Occasionally the in-process resolution for a database's hostname gets into a state where the resolver never returns a result, and from then on the hostname is **unresolvable indefinitely** — no restart, no recovery — while the same host resolves the name fine by every other means:

- `SHOW DNS_HOSTS` shows the hostname with `ttl=0` and an **empty** address list; `SHOW SERVERS` shows its servers at `(bad-af)`.
- Clients get `server DNS lookup failed` / `server_login_retry`.
- Meanwhile `getent`/`dig`/a fresh resolver on the same machine resolve the name, and a direct TCP connect to the database succeeds — so nothing outside PgBouncer is broken.
- The stuck process sends **no** DNS queries for that name — it is not even re-querying.
- `SHOW LISTS dns_pending` is no help: `adns_info()` hardcodes `*pending = 0`, so it never reflects an in-flight or stuck query.

Only a process restart (a fresh resolver state) clears it. This matches the long-running "DNS never recovers, restart fixes it" reports below.

## Root cause (`src/dnslookup.c`)

A request (`struct DNSRequest`) is re-issued **only once it is `->done`**:

- `adns_resolve()` relaunches only in its `if (req->done)` branch — a request that is still pending just attaches the new caller's token and returns, with no new query;
- `zone_requeue()` explicitly skips `!req->done` requests (and the SOA recheck that drives it is off by default, `dns_zone_check_period == 0`).

So if the resolver callback for a request never fires — a hung resolution — the request stays `->done == false` forever, its server sits at `(bad-af)`, and every later resolution of that name attaches to the stuck request without issuing a query. **Nothing times out a pending resolution.** Only a restart (a fresh channel/state) clears it — exactly the symptom above.

The upstream trigger for a never-firing callback varies (a libevent/evdns bug, a c-ares `getaddrinfo` hang, a resolver blip during a reboot/maintenance window), which is why this thread has spanned all three backends. But whatever produced the missing callback, **not retrying it is a PgBouncer-side defect** in the generic resolver layer.

## Reproduction

The precondition is a resolver callback that never fires. That is hard to trigger organically on demand — a normal DNS server always eventually answers or times out, and the timeout path *does* recover (it sets `->done`, so the TTL-expiry relaunch kicks in). It's the genuinely-never-returned case that gets stuck.

I observed it with the c-ares backend on long-running Kubernetes pods: the resolution stopped returning around a database maintenance window (a burst of connection failures / "the database system is shutting down"), after which the hostname stayed unresolvable until the pod was restarted — consistent with a c-ares `getaddrinfo` hang that leaves the callback unfired. I could not reproduce that organic hang locally (c-ares/evdns/libc all eventually time out and recover under packet loss, slow answers, CNAME-without-AAAA, and search-domain load).

The stuck *state*, however, reproduces deterministically. The accompanying PR does it in the standard pytest suite via a tiny env-gated test hook that drops one lookup so its callback never fires — the exact precondition. With the default config the hostname never recovers (the bug); with `dns_resolve_timeout` set it recovers without a restart (the fix). A second test confirms a late answer from the original query, arriving after recovery, is discarded rather than clobbering the fresh result. So the PR's tests demonstrate the failure and the fix on their own — no external DNS setup required.

## Proposed fix

Add a `dns_resolve_timeout` option (default `0`, disabled). When set, `adns_per_loop()` relaunches any request that has been outstanding longer than the timeout, so a hung resolution recovers on its own. A relaunched request can briefly have two queries in flight, so the first answer of a resolution episode wins and a late answer from the superseded query is discarded (it cannot clobber the fresh result).

PR: #1582. It builds warning-clean on all three DNS backends and adds an in-framework regression test.

Open question for maintainers: should the default be a protective non-zero value rather than disabled?

## Related

- #1119 — "Sometimes unable to recover from DNS server failure" (same symptom; concluded "probably a libevent bug, use c-ares").
- #1353 — "DNS reprobe issues" (c-ares **also** never recovers; closed pointing at #1119).
- #1125, #1126 — "DNS lookup failure" / "lookup failed … result=-4" (same never-recovers symptom on evdns2 and c-ares).
- #487 — negative `dns_queries` in `SHOW LISTS` (a related but distinct in-flight-counter accounting bug; the PR tightens the accounting on the relaunch path but does not claim to fix #487).
- #882 — `RESET_DNS` request (a manual/administrative recovery lever; `dns_resolve_timeout` is the automatic counterpart).

mdsakalu:
Fixes #1583

## Problem

PgBouncer re-issues a hostname lookup only once the request is `->done`: `adns_resolve()` relaunches in its `req->done` branch, and `zone_requeue()` skips requests that are not `->done`. So if the in-process resolver never delivers a result for a request — a hung resolution — the request stays pending forever, and every later resolution of that name attaches to the stuck request without issuing a new query. The database hostname then becomes permanently unresolvable (`SHOW DNS_HOSTS` shows it with no addresses, its servers stay at `(bad-af)`, clients get `server DNS lookup failed` / `server_login_retry`) even though a fresh resolver on the same host resolves the name — and only a restart clears it. There is no timeout on a pending resolution anywhere.

We hit this in production with the c-ares backend (PgBouncer 1.25.2, c-ares 1.34.6): reader pods stuck for 30+ days, cleared only by a pod restart, while `getent`, `dig`, a fresh c-ares channel, and a direct TCP connection to the DB all succeeded from the same pod. The upstream trigger is a c-ares `getaddrinfo` hang (the resolver returning no callback); this PR does not fix that c-ares race, but it removes PgBouncer's part of the failure — never retrying the stuck request — so the pooler recovers on its own.

Related upstream reports of "the in-process resolver never recovers": #383, #332, and the `RESET_DNS` request #882.

## Change

The first commit is a preparatory bug fix: `cleanup()` freed the DNS context (`adns_free_context()`, which frees every DNS token) *before* `objects_cleanup()` disconnected servers, so a server still resolving at exit had its DNS token freed and then cancelled (`adns_cancel()` → `list_del()` on freed memory) — a latent use-after-free reachable only in `--enable-cassert` builds (where `cleanup()` runs), which the new test surfaces under valgrind. It frees the DNS context after `objects_cleanup()`.

The main change adds a `dns_resolve_timeout` config option (default `0.0`, disabled). When set, `adns_per_loop()` relaunches any request that has been outstanding longer than the timeout, so a hung resolution recovers without a restart.

- `include/bouncer.h`, `src/main.c`: the new `CF_TIME_USEC` option.
- `src/dnslookup.c`:
  - `struct DNSRequest` gains `launch_time`, set by a new `launch_request()` helper (replacing the `ctx->active++; impl_launch_query()` pairs).
  - `adns_check_stuck()` walks the request tree once per second and relaunches stuck requests, collecting them into a dynamically-grown array first so the tree is not mutated mid-walk and no request is starved.
  - A relaunched request can briefly have several queries in flight, and a new resolution episode can start while an old query is still outstanding. `struct DNSRequest` gains an `epoch`/`answered_epoch` pair — a *new episode* (`start_resolution()`, from `adns_resolve()` and `zone_requeue()`) bumps `epoch`, a stuck relaunch does not — and each query is issued with a small per-query token (`struct DNSQuery`) stamped with the epoch it was launched for, threaded through all three backends as the callback argument. `got_result_gai()` accepts only an answer whose epoch is newer than `answered_epoch` and discards any superseded answer (a same-episode duplicate, or a straggler from an earlier episode), dropping its `ctx->active` count, so a late answer cannot clobber the fresh result or mark good addresses dirty.
- `doc/config.md`, `etc/pgbouncer.ini`: document the option.

The fix is always compiled and a no-op while `dns_resolve_timeout=0`, so it changes nothing by default. It builds warning-clean on all three DNS backends (c-ares, evdns, getaddrinfo_a).

## Test

`test/test_dns_stuck.py` (runs in the standard pytest suite, three cases pass):

- without `dns_resolve_timeout` (default) a hung lookup is never relaunched and the database stays unresolvable;
- with it set, the lookup is relaunched and the database resolves again without restarting PgBouncer;
- after recovery, a late answer from the original (superseded) query is discarded and does not clobber the fresh result — the database keeps resolving.

A normal DNS server can't produce a never-completing resolution (c-ares always eventually times out), so the tests reproduce the hang deterministically via minimal fault-injection hooks in `src/dnslookup.c`. They are compiled only in cassert builds (`#ifdef CASSERT`, like `main.c`'s `atexit` cleanup), so production binaries contain no test code, and are armed by the `PGB_TEST_DNS_FAULT` environment variable read once at startup in `main()` — following the standard debug-build fault-injection idiom (curl's `DEBUGBUILD` env vars, etcd's failpoints) rather than adding anything to the config surface. There is no `getenv()` on the resolve path. A cassert build advertises itself via `pgbouncer --version`, and the test suite skips when the hooks are absent.

## Notes / open questions for maintainers

- Should the default be a protective non-zero value rather than disabled?
- The per-query epoch token means a late answer is accepted only if newer than what was already delivered, so a straggler from any earlier episode is discarded rather than cancelled — no per-query cancel API is required (the legacy `ares_gethostbyname()` path doesn't expose one). The one remaining corner is a *synchronous* launch failure (only the `getaddrinfo_a` path has one) that delivers a failure without advancing `answered_epoch`; a real answer from a concurrent same-episode query is then still accepted, which is the desired recovery, and any straggler is corrected on the next refresh. Happy to adjust if a maintainer wants that path to advance `answered_epoch` too.

### DIFF
--- a/src/dnslookup.c
+++ b/src/dnslookup.c
@@ -899,6 +899,10 @@ static struct DNSQuery *new_query(struct DNSRequest *req)
 	return q;
 }
 
+/* test-only: the request (and its epoch) whose first lookup was dropped */
+static struct DNSRequest *test_hung_req;
+static unsigned test_hung_epoch;
+
 /*
  * Issue (or re-issue) the backend query for a request and record when, so that
  * adns_check_stuck() can relaunch a request whose resolver callback never fires.
@@ -908,6 +912,25 @@ static void launch_request(struct DNSRequest *req)
 	struct DNSQuery *q;
 
 	req->launch_time = get_cached_time();
+
+	/*
+	 * Test-only fault injection: drop the first lookup so its callback never
+	 * fires, reproducing the "hung in-process resolver" that leaves a request
+	 * pending forever (see adns_check_stuck / test/test_dns_stuck.py). Active
+	 * only when the PGB_TEST_HANG_ONCE environment variable is set; a no-op in
+	 * normal operation.  Dropped before active++/new_query, so nothing leaks.
+	 */
+	if (getenv("PGB_TEST_HANG_ONCE")) {
+		static bool hung_once = false;
+		if (!hung_once) {
+			hung_once = true;
+			test_hung_req = req;
+			test_hung_epoch = req->epoch;
+			log_warning("TEST: dropping lookup of '%s' to simulate a hung resolution",
+				    req->name);
+			return;
+		}
+	}
 
 	req->ctx->active++;
 	q = new_query(req);
@@ -1384,8 +1407,33 @@ static void adns_check_stuck(struct DNSContext *ctx)
 	free(w.stuck);
 }
 
+/*
+ * Test-only: once a dropped ("hung") request has recovered via a relaunch,
+ * simulate its original query finally answering late, carrying the epoch it was
+ * launched for.  The epoch guard in got_result_gai() must discard this stale
+ * answer so it cannot clobber the fresh result.  Pair it with an active++ so the
+ * discard's active-- balances (a real late query would have taken one when it
+ * was launched).  Active only under PGB_TEST_LATE_STALE.
+ */
+static void test_late_stale(void)
+{
+	struct DNSRequest *req = test_hung_req;
+	struct DNSQuery q;
+
+	if (!req || !req->done || !getenv("PGB_TEST_LATE_STALE"))
+		return;
+	test_hung_req = NULL;
+
+	q.req = req;
+	q.epoch = test_hung_epoch;
+	log_warning("TEST: delivering late stale callback for '%s'", req->name);
+	req->ctx->active++;
+	got_result_gai(0, NULL, &q);
+}
+
 void adns_per_loop(struct DNSContext *ctx)
 {
 	impl_per_loop(ctx);
 	adns_check_stuck(ctx);
+	test_late_stale();
 }
--- a/src/dnslookup.c
+++ b/src/dnslookup.c
@@ -85,6 +85,34 @@ struct DNSRequest {
 	struct addrinfo *oldres;
 
 	usec_t res_ttl;
+
+	/* when the currently-outstanding query was issued (see adns_check_stuck) */
+	usec_t launch_time;
+
+	/*
+	 * A request relaunched by adns_check_stuck() can briefly have more than one
+	 * query in flight, and a new resolution episode can start while an old query
+	 * is still outstanding.  epoch is bumped when a new episode starts (see
+	 * start_resolution), but not on a stuck relaunch; answered_epoch is the
+	 * highest epoch whose result has been delivered.  Each query carries the
+	 * epoch it was launched for (struct DNSQuery), and got_result_gai() accepts
+	 * only an answer newer than answered_epoch -- so a late answer from a
+	 * superseded query is discarded instead of clobbering the fresh result.
+	 */
+	unsigned epoch;
+	unsigned answered_epoch;
+};
+
+/*
+ * One per issued query.  Carries the epoch the query was launched for so a late
+ * answer from a superseded query (a stuck-relaunch duplicate, or a straggler
+ * from an earlier episode) can be recognised and discarded.  Passed to the
+ * backend as the callback argument and freed by the backend glue after
+ * got_result_gai() returns.
+ */
+struct DNSQuery {
+	struct DNSRequest *req;
+	unsigned epoch;
 };
 
 /* zone name serial */
@@ -229,7 +257,7 @@ const char *adns_get_backend(void)
 
 struct GaiRequest {
 	struct List node;
-	struct DNSRequest *req;
+	struct DNSQuery *q;
 	struct gaicb gairq;
 };
 
@@ -254,7 +282,8 @@ static void dns_signal(int f, short ev, void *arg)
 
 		/* got one */
 		list_del(&rq->node);
-		got_result_gai(e, rq->gairq.ar_result, rq->req);
+		got_result_gai(e, rq->gairq.ar_result, rq->q);
+		free(rq->q);
 		free(rq);
 	}
 }
@@ -286,7 +315,7 @@ static bool impl_init(struct DNSContext *ctx)
 	return true;
 }
 
-static void impl_launch_query(struct DNSRequest *req)
+static void impl_launch_query(struct DNSRequest *req, struct DNSQuery *q)
 {
 	static const struct addrinfo hints = { .ai_socktype = SOCK_STREAM };
 
@@ -300,7 +329,7 @@ static void impl_launch_query(struct DNSRequest *req)
 		goto failed2;
 
 	list_init(&grq->node);
-	grq->req = req;
+	grq->q = q;
 	grq->gairq.ar_name = req->name;
 	grq->gairq.ar_request = &hints;
 	list_append(&gctx->gairq_list, &grq->node);
@@ -321,6 +350,7 @@ failed:
 	list_del(&grq->node);
 	free(grq);
 failed2:
+	free(q);
 	req->done = true;
 	deliver_info(req);
 }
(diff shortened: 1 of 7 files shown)


src/dnslookup.c
```diff
@@ -847,6 +887,75 @@ void adns_free_context(struct DNSContext *ctx)
...
+	req->launch_time = get_cached_time();
+
+	/*
+	 * Test-only fault injection: drop the first lookup so its callback never
+	 * fires, reproducing the "hung in-process resolver" that leaves a request
+	 * pending forever (see adns_check_stuck / test/test_dns_stuck.py). Active
+	 * only when the PGB_TEST_HANG_ONCE environment variable is set; a no-op in
+	 * normal operation.  Dropped before active++/new_query, so nothing leaks.
+	 */
+	if (getenv("PGB_TEST_HANG_ONCE")) {
```
AndrewJackson2020: Are there other places in the codebase that use this pattern? This seems not ideal. 

I think this should be inserted via macro for test only builds or something like that, at a minimum.

mdsakalu: Good call, changed it to use CASSERT in c95a206. I don't think the env var pattern exists anywhere else in the codebase, but I couldn't think of a way to do fault-injection on per-test basis without it. Now CASSERT arms the fault and the env var fires it, but this whole section stays out of the release build.

Even though I regularly encounter the bug this fixes in the wild, I could not find a way to reliably reproduce it - hence this fault-injection strategy.

AndrewJackson2020: Understood, I just wanted to make sure that test code isn't going to make it into prod environments. I think CASSERT does that unless there is some use case for using CASSERT in prod that I'm not aware of. 

Since this is a new conception in the codebase (injecting code to force nondeterministic errors for testing) you may want to consider using a config entry instead of an env var. That way all of the relevant switches and whatnot are uniformly stored in structured lists like `bouncer_params` in `src/main.c` instead of "random envvar in the middle of the codebase".

mdsakalu: I did confirm that CASSERT is not used for releases. I tried a config entry as you suggested, but it gets flagged by `optscan.sh` during CI for being undocumented. It seemed like I'd have to change a lot more to make that work (build a separate test-only config, or make optscan test-aware). Instead, I moved the env var evaluation to main so the env var doesn't get loaded on every DNS resolve.

[curl](https://curl.se/libcurl/c/libcurl-env-dbg.html) and [etcd's gofail](https://github.com/etcd-io/gofail/blob/master/doc/design.md) both support env vars for fault injection, so it's at least not a crazy idea. I can pursue the config route though if you still think that's a better solution.

Postgres itself has something called [Injection Points](https://www.postgresql.org/docs/devel/xfunc-c.html#XFUNC-ADDIN-INJECTION-POINTS) for this same purpose, but I think something like that would need to be separate from this PR.

AndrewJackson2020: Yeah I just was reading about injection points today. Definitely not a crazy idea for testing conditions which are almost impossible to trigger deterministically.

I think probably envvar is the way to go here. Maybe in the future if we had a bunch more use cases for this functionality it might make sense to build out dedicated injection point infrastructure. For now this seems fine.

PR #1584, opened 2026-09-03
feat: add support for upstream TLSDirect

dpittner:
What: attempt in addressing the remaining part of https://github.com/pgbouncer/pgbouncer/issues/1203 as per https://github.com/pgbouncer/pgbouncer/issues/1203#issuecomment-3182650069 and adding support for upstream tlsdirect

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -1003,6 +1003,23 @@ verify-full
     according to `server_tls_ca_file`.  Server host name must match
     certificate information.
 
+### server_tls_direct
+
+If set to 1, PgBouncer initiates a TLS handshake immediately upon TCP connect,
+without sending the PostgreSQL `SSLRequest` message first.  Use this when
+connecting to PostgreSQL 17+ servers or proxies that expect a direct TLS
+connection rather than the standard SSL-negotiation protocol (e.g. a
+TLS-terminating load balancer or stunnel).
+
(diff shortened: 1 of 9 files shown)


dubee:
Thanks for working on this, direct TLS is exactly what I need for IBM Cloud VPE Postgres endpoints, which currently require an stunnel sidecar.

I hit one issue: the connection fails with tlsv1 alert no application protocol unless server_tls_sslmode is verify-full.

The cause is in sbuf_tls_connect(), src/sbuf.c:1461-1462:


	if (cf_server_tls_sslmode != SSLMODE_VERIFY_FULL)
		hostname = NULL;
The hostname passed from handle_connect() is correct, but this drops it before it reaches tls_connect_fds(), which is what sets SNI. That was harmless for STARTTLS, but with direct TLS the handshake is the first thing on the wire, so SNI is how the endpoint selects the backend. Without it the connection is rejected.

Removing those two lines fixes it, and doesn't weaken verification — setup_tls() already calls tls_config_insecure_noverifyname() for the modes below verify-full.

Confirmed with openssl s_client -alpn postgresql: succeeds with -servername, reproduces the alert with -noservername.

dpittner:
Thanks, I've added the fix you suggested and included a testcase to ensure we're covering.

victorrelkins-design:
Tracking Issue

PR #1597, opened 2026-09-12
Say what an eviction made room for

### LINKED ISSUE #1596: Replication client eviction closes every idle server in the pool instead of one, and over-full pools evict nothing
Summary:
When a replication client connects to a pool that's already at pool_size, PgBouncer is supposed to evict one idle server to make room. Instead, the eviction loop in launch_new_connection() compares the server count against the limit backwards, so it evicts far more (or far fewer) servers than it should.

Impact:
1. Over-eviction: A single replication client asking for one slot closes every idle server in the pool, not just one. In a two-server pool, both idle servers get evicted for one incoming client.
2. Under-eviction: If a pool ends up holding more servers than pool_size (e.g., after a config reload lowers pool_size, or via the cancel-request path that allows up to double pool_size), the same backwards comparison causes the loop to evict nothing at all, and the client opens yet another connection on top of the already-over-limit pool.

Steps to reproduce:
- Set up a pool with pool_size=2, open two idle server connections, then connect a replication client. Expect one eviction; observe both connections closed.
- With two idle servers, lower pool_size from 2 to 1 via reload. Expect the pool to shrink back to 1. Observe zero evictions and a third server connection opened instead.

benchub:
Fixes #1596.

Root cause:
In src/objects.c, the loop read as while (evict_if_needed && pool_pool_size(pool) >= max), where max was a one-time snapshot of the server count taken before the loop started and never updated. Because neither operand changed while the loop ran, once the pool was full the condition stayed true until evict_pool_connection() ran out of idle servers to close - it evicted everything rather than stopping once there was room. The same inverted comparison meant an over-full pool (count > limit) failed the loop's entry condition entirely, so it skipped eviction and let the client open a new connection instead.

Secondary issue (log clarity):
Related to diagnosing this, all three eviction paths (max_db_connections, pool_size, max_user_connections) logged the same generic "evicted" reason, so there was no way to tell from the log which limit actually triggered an eviction.

Fix (this branch):
1. Compare the server count against the limit in the correct direction, and re-read the count after each eviction so the loop stops as soon as there's room — instead of relying on a stale snapshot.
2. Give each of the three eviction paths its own log reason ("evicted for pool_size", "evicted for max_db_connections", "evicted for max_user_connections") so the log answers which limit was full.

### DIFF
--- a/src/objects.c
+++ b/src/objects.c
@@ -1930,11 +1930,12 @@ void launch_new_connection(PgPool *pool, bool evict_if_needed)
 			}
 
 			if (c && c->replication && !sending_auth_query(c)) {
-				while (evict_if_needed && pool_pool_size(pool) >= max) {
+				while (evict_if_needed && max >= pool_pool_size(pool)) {
 					if (!evict_pool_connection(pool))
 						break;
+					max = pool_server_count(pool);
 				}
-				if (pool_pool_size(pool) < max)
+				if (max < pool_pool_size(pool))
 					goto allow_new;
 			}
 			log_debug("launch_new_connection: pool full (%d >= %d)",
--- a/src/objects.c
+++ b/src/objects.c
@@ -1800,7 +1800,7 @@ bool evict_connection(PgDatabase *db)
 	}
 
 	if (oldest_connection) {
-		disconnect_server(oldest_connection, true, "evicted");
+		disconnect_server(oldest_connection, true, "evicted for max_db_connections");
 		return true;
 	}
 	return false;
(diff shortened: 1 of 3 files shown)


eulerto:
I checked this PR and it is a bug in #876 . I polished this fix a bit in the next update. I rebased it. I squashed both commits in a single one. Added test_replication_pool_size_over_full_pool. I fixed a wrong test command. I don't use max for initial part of `launch_new_connection()` -- it is a live count so I use a new variable server_count. I removed a redundant user-eviction test that doesn't cover this bug fix. I adjusted the commit message to reflect the current state.

PR #1599, opened 2026-09-12
Send NULL for a user pool_size that is not set

### LINKED ISSUE #1598: SHOW USERS sends '' instead of NULL for an unset pool_size/reserve_pool_size
SHOW USERS reports a user's pool_size and reserve_pool_size overrides as an empty string when they are not set, instead of SQL NULL. pool_mode in the same row already correctly returns NULL for "not set" — pool_size/reserve_pool_size should match.

benchub:
Fixes #1598.

pktbuf_write_DataRow writes a length of -1 for NULL and 0 for an empty string, so a client can tell the two apart -- but SHOW USERS sent the empty string for an unset pool_size or reserve_pool_size, leaving no way to distinguish "no override" from an override deliberately set to nothing.  pool_mode in the same row already sends NULL for precisely that meaning; send NULL for these two as well.  doc/usage.md has documented them as "or NULL if not set" for as long as they have existed.

Additionally, as long as we're in the area, drop the %9d fixed-width padding in favor of plain %d. This cosmetic change affects column alignment in tools like psql, not the data returned to any client.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -547,10 +547,10 @@ name
 :   The user name
 
 pool_size
-:   The user's override pool_size. or NULL if not set.
+:   The user's override pool_size, or NULL if not set.
 
 reserve_pool_size
-:   The user's override reserve_pool_size. or NULL if not set.
+:   The user's override reserve_pool_size, or NULL if not set.
 
 pool_mode
 :   The user's override pool_mode, or NULL if not set.
--- a/src/admin.c
+++ b/src/admin.c
@@ -398,12 +398,19 @@ static bool admin_show_users(PgSocket *admin, const char *arg)
 		"max_user_client_connections", "current_client_connections");
 	statlist_for_each(item, &user_list) {
 		PgGlobalUser *user = container_of(item, PgGlobalUser, head);
-		char pool_size_str[12] = "";
-		char res_pool_size_str[12] = "";
-		if (user->pool_size >= 0)
-			snprintf(pool_size_str, sizeof(pool_size_str), "%9d", user->pool_size);
-		if (user->res_pool_size >= 0)
-			snprintf(res_pool_size_str, sizeof(res_pool_size_str), "%9d", user->res_pool_size);
+		char pool_size_buf[12], res_pool_size_buf[12];
+		/* pool_size and reserve_pool_size are NULL when unset, as pool_mode is */
+		const char *pool_size_str = NULL;
+		const char *res_pool_size_str = NULL;
+
+		if (user->pool_size >= 0) {
+			snprintf(pool_size_buf, sizeof(pool_size_buf), "%d", user->pool_size);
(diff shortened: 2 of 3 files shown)


eulerto:
I reviewed this PR. I'm mainly concerned about the compatibility risks:

* any client that checks `pool_size = ''` to means "unset" stops matching. A Python .strip() on a NULL now raises AttributeError.
* padding removal: anything comparing raw padded strings or parsing by column width would break.

The column type stays text so there is no wire-level change; only the empty vs NULL distinction.

The commit e18f600 that introduced the per-user `pool_size` (1.23.0) introduced the empty string. The commit 83eb532 that introduced the per-user `reserve_pool_size` (1.24.0) followed the same pattern. IMO they were an oversight.

I'm wondering if we agree with such change, it would be a good opportunity to change the type from text to integer. Every adjacent column in SHOW USERS (`max_user_connections`, `current_connections`, ...) is int. If you compare to SHOW DATABASES, columns like `pool_size`, `min_pool_size` and `reserve_pool_size` are also int. Once the current idea is accepted (NULL rather than ''), there is no reason to keep it as text. What about the compatibility? The psycopg3 turns an int column into a Python int, so user["pool_size"] == "1" starts failing. psql output is unaffected. In summary, doing NULL rather than '' now and text -> int later means clients adapt twice.

benchub:
That sounds like a reasonable enhancement, @eulerto. I am also leery of the compatibility problems, but I'm even more leery about inconsistency of some places using NULL and others using '' when both are supposed to represent "unknown." As DBAs, we should be comfortable with the correct masochism of triple valued boolean logic. :)

AndrewJackson2020:
Fwiw I raised a similar change a while ago [0] but I withdrew it after receiving feedback and thinking about "what is the benefit to this change and how many scripts, collectors, etc are deployed in the wild that could break if it was merged". 

The benefit is consistency and accuracy and new scripts/collectors not needing to convert strings to ints, be able to differentiate 0 from null, etc. the detriment is that existing scripts may need to update how they handle these values: have an if statement to check if it's a string before attempting type conversion, convert previously 0 but now null values to 0 to match current data to historical. Similarly when I saw there was a PR to remove the "show fds" functionality I checked pg_exporter and pgwatch to confirm they did not use those views (they did not).

After thinking through this I closed the PR as I figured the value was questionable relative to the risks and would prefer maintainers look at other PRs that did not have this risk/value trade off. Maybe that was an incorrect calculation, just throwing my perspective out there.

[0] https://github.com/pgbouncer/pgbouncer/pull/1441

benchub:
You're not wrong, @AndrewJackson2020, but of course the longer this inconsistency persists the harder it gets baked in. I don't care terribly much either way, but I guess I do lean more towards the side of "let's be consistent". I just found this while working on a new feature and thought I'd file the bug.

PR #1603, opened 2026-09-13
Feature: Add LDAP user name mapping

kmsarabu:
Some setups namespace their PostgreSQL roles, like `postgres://prod/alice`, while the directory only knows the person as `alice`. PgBouncer currently sends the role name to LDAP exactly as it received it, and the search+bind path rejects any name containing a `/`, so these roles can't use LDAP at all. 

Note that this is already supported in PAM, ldap lacks this.

This feature adds two options that rewrite the name for the bind only:

    `ldapusernameregex="^postgres://[a-z0-9_]+/(.+)$" ldapusernamereplacement="\1"`

The PostgreSQL name is untouched, still the role we connect to the server as, still what HBA rules match on. With no regex set, nothing changes.


AndrewJackson2020:
As you point out, postgres does natively support name mapping for external auth systems via `pg_ident.conf`[0]. Awkwardly this does not apply to LDAP auth, though there was an attempt to add it in the past [1]. I feel like any user name mapping in pgbouncer should stick as close to the postgres implementation as possible, or possibly just a subset of it.

That said, why not implement this feature as a new `pg_ident.conf` file in pgbouncer instead of new fields in pgbouncers `pg_hba.conf` that dont existing in the postgres `pg_hba.conf`?

[0] https://www.postgresql.org/docs/current/auth-username-maps.html
[1] https://www.postgresql.org/message-id/flat/1a61806047c536e7528b943d0cfe12608118ca31.camel@vmware.com

kmsarabu:
Thanks @AndrewJackson2020. I did look at `pg_ident`, but the main issue is the mapping direction.

`pg_ident` maps an external identity to a PostgreSQL role, where the external identity is already known (cert CN, peer user, Kerberos principal, etc.). With LDAP, we need to derive the LDAP username before we can perform the search/bind.

In this case the client connects as `postgres://dev/alice`, which is also the PostgreSQL role, while LDAP only knows `alice`. So we need:

`postgres://dev/alice` → `alice` → LDAP auth

while keeping `postgres://dev/alice` unchanged for HBA matching and the backend connection.

My understanding is that the existing `pg_ident` regex syntax doesn't seem to support that reverse transformation. One option that would work is reading exact-match entries in reverse, e.g. `ldapmap alice postgres://dev/alice`, but that would require one entry per user and is not scalable.

If you'd prefer that approach over adding LDAP mapping options, I'm happy to rework it that way.

Thanks.

Issue #1605, opened 2026-09-15
sbuf_wait_for_data_forced() can insert sbuf->ev twice and livelock the event loop


knizhnik:
# Event loop livelock: `sbuf.ev` I/O list becomes a self-cycle after `sbuf_wait_for_data_forced()`

## Summary

PgBouncer can stop accepting connections while spinning at 100% CPU inside libevent. We caught a live process (PgBouncer **1.25.1**, libevent **2.1**, aarch64) whose `struct event` for a backend socket had `ev_io_next == itself`. `evmap_io_active_()` never returns, so `accept()` on the listen socket never runs.

This is not fixed in **1.25.2** or current `master`. The same `sbuf_wait_for_data_forced()` / `sbuf_recv_forced_cb` code is still present.

## Environment

- PgBouncer 1.25.1 (`pgbouncer_1_25_1`)
- libevent 2.1 (`libevent-2.1.so.7`)
- Linux aarch64
- Sidecar in front of local Postgres (`127.0.0.1:5432`), listen `0.0.0.0:6432`

## Symptoms

- One core at 100% CPU
- New clients hang (kernel accept queue grows; we saw `LISTEN Recv-Q = 7` on `:6432`)
- Existing backend sockets are not read (one Postgres connection had ~3.6 MiB `Recv-Q`)
- Admin / metrics scrape of that process also stalls

## Evidence

`perf record -F 99 -g -p $PID -- sleep 15` spent **~100%** in:

```
event_base_loop → epoll_dispatch → evmap_io_active_ → event_active_nolock_
```

Zero time in PgBouncer `sbuf_*` callbacks, almost zero kernel / `epoll_wait`. The thread never left the I/O-map walk.

gdb (offsets for this libevent 2.1 `struct event` on aarch64: `ev_fd` at +56, `le_next` at +72, `ev_events` at +104, callback at +24):

```
(gdb) p/d *(int *)($x19 + 56)
$2 = 13                          # backend fd to Postgres
(gdb) p/x $w20
$3 = 0x2                         # EV_READ
(gdb) p/x *(short *)($x19 + 104)
$4 = 0x2                         # EV_READ (not EV_PERSIST)
(gdb) p/x *(void **)($x19 + 72)
$5 = 0x55844bb938                # same as $x19
(gdb) info symbol *(void **)($x19 + 24)
sbuf_recv_forced_cb in section .text of pgbouncer
```

Walking the I/O list:

```
[...]

PR #1606, opened 2026-09-16
prevent libevent event-loop livelock

knizhnik:
See https://github.com/pgbouncer/pgbouncer/issues/1605

Live incident on 1.25.1 + libevent 2.1: `sbuf->ev` had `ev_io_next == self`, callback `sbuf_recv_forced_cb`. `evmap_io_active_()` never returned (100% CPU, listen Recv-Q grew, accept starved).

`event_assign()` does not unlink. A second `event_add()` on an event that is still the I/O-map head is `LIST_INSERT_HEAD` of an already-inserted element → self-cycle.

The old `sbuf_wait_for_data_forced()` only `event_del()`d when `wait_type != W_NONE`, ignored del failure, then always assigned. `sbuf_wait_for_data()` assigned with no del at all. The loopcnt path always called `sbuf_wait_for_data_forced()` after ignoring `sbuf_process_pending()`.

## What this patch does (vs a pending-only del)

1. **`sbuf_event_remove()`** — skip `event_del` until `event_initialized()` (never-assigned `ev_base == NULL` → del returns -1). After that, always `event_del`. Do **not** use `event_pending()` as the only guard: it is not the same as “in the I/O map”, and it warnx-es on a never-assigned event.

2. Call it before every `event_assign(&sbuf->ev, …)` on the recv/forced/TLS-once paths.

3. **Loopcnt** — `sbuf_process_pending()` returning false is not “ignore me”. It can mean pause, close, destination backpressure (`W_SEND`), or “need another turn”. Forced rearm only if the socket is still open **and** `wait_type == W_RECV`. That keeps an existing EV_WRITE watcher.

4. **Failed del / failed forced add** — libevent 2.1 unlinks from the inserted queue *before* `evmap_io_del_()`, so a del error does **not** mean the event is still on the I/O list. Returning false and continuing as W_RECV would leave the socket with no watcher. Loopcnt now `sbuf_call_proto(…, SBUF_EV_RECV_FAILED)`.

## Tests

`test/test_sbuf_loopcnt.py` is a regression for “loopcnt=1 still accepts” and “client backpressure still accepts”. It does **not** deterministically reproduce the I/O-map self-cycle (that still needs the assign-without-del sequence).


cosgroveb:
`It does not deterministically reproduce the I/O-map self-cycle` can confirm. This looks convincing but haven't found a great way to demonstrate it in test yet.

cosgroveb (changes requested):

test/test_sbuf_loopcnt.py
```diff
@@ -0,0 +1,53 @@
...
+        "fails with EINVAL instead of reporting a test timeout. "
+        "Win32 also ignores test.ini tcp_socket_buffer=4096."
+    ),
+)
+def test_sbuf_loopcnt_client_backpressure_still_accepts(bouncer):
+    bouncer.write_ini("sbuf_loopcnt = 1")
+    bouncer.admin("RELOAD")
+
+    # A few 8KiB rows is enough to fill test.ini's tcp_socket_buffer=4096
+    # and push the client sbuf into W_SEND.  Do not leave tens of MB unread:
```
cosgroveb, line 42:
```suggestion
    # and push the client sbuf into W_SEND waiting for client sock readability.
    # leave tens of MB unread:
```

Adds a little more "why". Optional. Take-it or leave-it.

test/test_sbuf_loopcnt.py
```diff
@@ -0,0 +1,53 @@
...
+from .utils import WINDOWS
+
+
+def test_sbuf_loopcnt_still_accepts(bouncer):
+    bouncer.write_ini("sbuf_loopcnt = 1")
+    bouncer.admin("RELOAD")
+
+    rows = bouncer.sql("SELECT repeat('x', 200) FROM generate_series(1, 5000)")
+    assert len(rows) == 5000
+    assert bouncer.sql("SELECT 1", connect_timeout=3)[0][0] == 1
```
cosgroveb: We have a helper you can use here.

```suggestion
-    assert bouncer.sql("SELECT 1", connect_timeout=3)[0][0] == 1
+    assert bouncer.sql_value("SELECT 1", connect_timeout=3) == 1
```

test/test_sbuf_loopcnt.py
```diff
@@ -0,0 +1,53 @@
...
+    # and push the client sbuf into W_SEND.  Do not leave tens of MB unread:
+    # Connection.close() would then block in libpq draining the socket.
+    stalled = bouncer.conn()
+    try:
+        stalled.pgconn.send_query(
+            b"SELECT repeat('x', 8192) FROM generate_series(1, 32)"
+        )
+        stalled.pgconn.flush()
+        time.sleep(0.5)
+        assert bouncer.sql("SELECT 1", connect_timeout=3)[0][0] == 1
```
cosgroveb, line 51: Likewise: 

```suggestion
        assert bouncer.sql_value("SELECT 1", connect_timeout=3) == 1
```

test/test_sbuf_loopcnt.py
cosgroveb, line 24:
```suggestion
    assert bouncer.sql_value("SELECT 1", connect_timeout=3) == 1
```

We have a helper you can use here.

cosgroveb (changes requested):
I spent some time with this and have a deterministic repro of the write-wait registration loss. My branch also shows this change fixes it:

https://github.com/pgbouncer/pgbouncer/compare/master...cosgroveb:pgbouncer:pr-1606-write-wait-regression-test

How it works:

1. The fake backend [builds 4096-byte segments ending with an 11-byte `DataRow`](https://github.com/cosgroveb/pgbouncer/blob/4ba48b710b75151b56d031c3c1eabf802ba33748/test/sbuf_fake_backend.py#L139-L156). With the buffer full, `sbuf_process_pending()` [defers `SBUF_SMALL_PKT` bytes or fewer](https://github.com/cosgroveb/pgbouncer/blob/4ba48b710b75151b56d031c3c1eabf802ba33748/src/sbuf.c#L773-L783), leaving that row for the next pass.
2. The test sets `sbuf_loopcnt = 1` and stops reading from the client. At the loop limit, PgBouncer [processes pending bytes again](https://github.com/cosgroveb/pgbouncer/blob/4ba48b710b75151b56d031c3c1eabf802ba33748/src/sbuf.c#L977-L990); the blocked client send returns `EAGAIN`.
3. `sbuf_queue_send()` [replaces the backend read event with `EV_WRITE` on the client socket and changes the backend `SBuf` state to `W_SEND`](https://github.com/cosgroveb/pgbouncer/blob/4ba48b710b75151b56d031c3c1eabf802ba33748/src/sbuf.c#L571-L597).
4. Vulnerable code then calls [`sbuf_wait_for_data_forced()`](https://github.com/cosgroveb/pgbouncer/blob/4ba48b710b75151b56d031c3c1eabf802ba33748/src/sbuf.c#L521-L540), which deletes the client-write watcher. The PR [returns when the state changes to `W_SEND`](https://github.com/pgbouncer/pgbouncer/blob/5c071365c97111648739e8d76604b8508d7e7c7d/src/sbuf.c#L1018-L1023), preserving it.
5. A later retry can restore the write wait, so the transfer can still finish. The test [verifies the result and fails if the watcher disappeared](https://github.com/cosgroveb/pgbouncer/blob/4ba48b710b75151b56d031c3c1eabf802ba33748/test/test_sbuf_write_wait.py#L118-L138).


I don't know that it would be worth the complexity to include this test in the PR. I think we should use it instead of the current PR test, though, because the PR's test passes without the fix. I'll leave that decision to @JelteF and am requesting changes for that reason. The C fix looks correct to me. 👍


(Edit: simplifications)

test/test_sbuf_loopcnt.py
```diff
@@ -0,0 +1,54 @@
...
+        "Win32 also ignores test.ini tcp_socket_buffer=4096."
+    ),
+)
+def test_sbuf_loopcnt_client_backpressure_still_accepts(bouncer):
+    bouncer.write_ini("sbuf_loopcnt = 1")
+    bouncer.admin("RELOAD")
+
+    # A few 8KiB rows is enough to fill test.ini's tcp_socket_buffer=4096
+    # and push the client sbuf into W_SEND waiting for client sock readability.
+    # leave tens of MB unread:
```
cosgroveb: Apologies I futzed the wording on my previous suggestion. This is a corrected comment:

```suggestion
    # and push the server sbuf into W_SEND, where it waits for the client socket
    # to become writable. Do not leave tens of MB unread:
```

cosgroveb:
FWIW I tried but couldn't repro the livelock. I wrote that new test above and also:

- Traced what PgBouncer waited on while repeatedly forcing the blocked-write path.
- Saw it lose and restore the write wait, but never enter the CPU spin.

Doesn't mean it doesn't happen. I just couldn't...

eulerto:
I couldn't make PgBouncer hit the livelock either. However, the libevent misuse is real. The changes do what they claim:
(a) the sbuf_event_remove() always deletes an initialized event before every re-assign, (b) the loopcnt block re-arms only when the socket is open and wait_type == W_RECV and (c) a failed re-arm goes to SBUF_EV_RECV_FAILED.
I traced this issue until commit 09e7ab4. It added sbuf_wait_for_data_forced(), with the if (wait_type != W_NONE) event_del() guard and the ignored return value.
I'm inclined to merge this PR (without the test because it cannot reproduce reliably) despite of not confirming the incident.

knizhnik:
I have now yet another repro at prod.
If you want to me to check something in debugger - I can do it.
But unfortunately it is release build, so debugging is not trivial: it is not possible to inspect values of all variables.

eulerto:
@knizhnik Did you check if your fix still works for it? As I said, I'm fine with your patch but won't include the test.

cosgroveb:
My test here red/greens sorry if I was unclear https://github.com/pgbouncer/pgbouncer/compare/master...cosgroveb:pgbouncer:pr-1606-write-wait-regression-test

might not be worth adding to the test suite though unsure? 

> I'm inclined to merge this PR (without the test because it cannot reproduce reliably) despite of not confirming the incident.

+1 to deleting the PRs test though

knizhnik:
> @knizhnik Did you check if your fix still works for it? As I said, I'm fine with your patch but won't include the test.

Certainly this is instance without this patch. I was not able to reproduce the problem with this patch.

cosgroveb (approved):

cosgroveb:
I'm finally able to consistently reproduce 100% CPU spin with a self-linked event-list but it requires a protocol-violation. If `ReadyForQuery` is sent twice this happens. I also confirmed the code in this PR prevents that spinlock-ish behavior https://github.com/pgbouncer/pgbouncer/compare/master...cosgroveb:pgbouncer:pr-1606-auth-query-double-ready-test#diff-0223f4eea1bd234ea9c842a5476e062930740912fa9fec56479cee1525d9199eR103-R104

knizhnik:
Is it ok to merge this PR with current test or it is better to replace it with
https://github.com/pgbouncer/pgbouncer/commit/78f7f4c3573ce50e2fca8b747de737bab33f1054  ?

eulerto:
@knizhnik I would like to commit it without a test. I already have too much flaky tests that I don't want to spend time on a new one. Could you prepare a commitable patch?

knizhnik:
> @knizhnik I would like to commit it without a test. I already have too much flaky tests that I don't want to spend time on a new one. Could you prepare a commitable patch?

Done: test removed.

PR #1609, opened 2026-09-17
Add a logo after 19 years

### LINKED ISSUE #1425: No logo
hi folks, i wanted to drop a possible logo suggestion

Image

thanks and i hope you have a lovely day

JelteF:
While preparing the n-th talk where I cannot reference PgBouncer in the slides using a logo. I decided it's time to bite the bullet and try to agree on one.

Created with a lot of back and forth with Codex. I think this end result looks nice. Modern and slightly polished, but not too professionally designed either. The color is "Postgres blue" but since it's monochrome you can use any color. And it's an svg.

Closes #1425


petere:
Looks like the elephant is ... er ... pooping?

Without that part it actually looks quite fun.

But I'm not sure a logo is needed.

JelteF:
I had meant it as a "bounce animation", but I definitely also understand your interpretation. So I removed the "poop" and agree that it looks good without it.

For me the main reason to have a logo is so I can use it in slidedecks and diagrams. e.g. on a slide with all connection poolers I could find, PgBouncer was the only one I didn't have a logo for. I obviously can just use this one even if it's not the official logo, but it seems nicer to have something official that people can recognize.

Also, I think it would be nice to replace the pgbouncer github org logo with it. And use it as the pgbouncer docs favicon (currently nothing) so I can find it more easily in my open tabs.

cosgroveb (approved):

eulerto:
It is strange to have a logo after almost 2 decades. The project was always identified by its name. It will take some time to associate the name (PgBouncer) with the new logo. BTW I'm not opposed to the logo idea. As you said in the previous reply, diagrams don't sound good with just logos because PgBouncer doesn't have one.

You use the same color (#336791) as the Postgres logo. It also looks nice in all-black and all-white. I also tried the favicon and it looks good to me.

Regarding the image, the elephant only have three legs. Is the hidden leg aligned with the other one? I also noticed that the shape of the foot isn't visible on one of the legs (in the middle).

eulerto (approved):

JelteF:
> Regarding the image, the elephant only have three legs. Is the hidden leg aligned with the other one? I also noticed that the shape of the foot isn't visible on one of the legs (in the middle).

Addressed those things now

JelteF:
I added it as the github org avatar, and as the website favicon and in the website navbar. (and this PR already added it to the readme).

